PR #4, opened 2022-08-12
Add a CI workflow using neo4j.

mihaimaruseac:
This has dual purpose:
- we are adding some CI to the repository, which will evolve shortly as we keep adding glue/starter code and actual tests, etc.
- we are testing using neo4j from docker and creating a reproducible build/test environment that developers can use for testing on their  machine too

The local go files are temporary, for testing purposes only. Still, please review them as I'm learning go while doing this :)

---

### To test locally 

In one terminal run the neo4j docker service

```console
$ docker run --name aff -it -d --env NEO4J_AUTH=none -p=7474:7474 -p=7687:7687 neo4j
```

Then in another terminal, you can either connect to this docker instance and run Cipher queries manually (via `docker exec --interactive --tty aff cypher-shell`) or run the test suite/code:

```console
$ go run main.go 
Connected to neo4j://localhost:7687
Found the following nodes: 
	&{[{0 [Msg] map[text:Hello world!]}] [a]}
	&{[{1 [Msg] map[text:Welcome to AFF.]}] [a]}
Found the following links: 
	&{[{0 [Msg] map[text:Hello world!]} {1 [Msg] map[text:Welcome to AFF.]}] [a b]}
```

When done, terminate the docker instance

```console
$ docker stop aff; docker rm aff
```

---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,47 @@
+#
+# Copyright 2022 The AFF Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


mihaimaruseac:
I put it here for now as it was only for testing the infra. The files in the main repo would surely look different, but let me think for a little bit, maybe it's actually worth moving.

lumjjb:
Will merge first and we can change it later, want to close out the PRs and move the code over

PR #8, opened 2022-08-20
Implement initial DSSE verification

mlieberman85:
This adds dsse envelop processing and tests. This also adds an
    inmemory keystore that can be used for now until we figure out
    what we are using long term for fetching and storing public keys
    for verification. This also creates a verifier interface along
    with a Sigstore verifier for the purposes of verifying DSSE
    envelopes through Sigstore


mlieberman85:
Is DCO working right?

mlieberman85:
oh looks like commits were rewritten so I just cherry-picked. Closing this one.

PR #18, opened 2022-08-23
add new collector interface

pxp928:

Collector interface based on discussion with the team

### DIFF
--- a/pkg/ingestor/collector/collector.go
+++ b/pkg/ingestor/collector/collector.go
@@ -15,4 +15,45 @@
 
 package collector
 
-type Collector interface{}
+import (
+	"context"
+
+	"github.com/guacsec/guac/pkg/ingestor/processor"
+)
+
+const (
+	BufferChannelSize int = 1000
+)
+
+type Collector interface {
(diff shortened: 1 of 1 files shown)


pkg/ingestor/collector/collector.go
```diff
@@ -15,4 +15,40 @@
...
+
+	"github.com/guacsec/guac/pkg/ingestor/processor"
+)
+
+const (
+	BufferChannelSize int = 1000
+)
+
+type Collector interface {
+	// Retrieve the artifacts from the collector
```
lumjjb, line 29: Add a bit more detail about how this should block until everything is done and return nil or error in either case. And also talk about how documents should be emitted to the channel.

Also golang style comments `// RetrieveArtifacts ...` so it gets picked up by godoc!

lumjjb (approved):
Minor nit on adding more details to comments if not, LGTM

pkg/ingestor/collector/collector.go
pxp928: added comment.

mihaimaruseac (approved):

PR #22, opened 2022-08-24
Add DSSE processor

mlieberman85:
This implements a processor that unpacks DSSE documents

### DIFF
--- a/pkg/handler/processor/dsse/dsse.go
+++ b/pkg/handler/processor/dsse/dsse.go
@@ -1,0 +1,94 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dsse
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
+)
+
+type dssePayloadType string
+
+const (
+	dsseITE6 dssePayloadType = "https://in-toto.io/Statement/v0.1"
+)
+
+type DSSEProcessor struct {
+}
+
+func (d *DSSEProcessor) ValidateSchema(i *processor.Document) error {
+	if i.Type != processor.DocumentDSSE {
+		return fmt.Errorf("expected document type: %v, actual document type: %v", processor.DocumentDSSE, i.Type)
+	}
+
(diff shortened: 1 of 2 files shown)


mlieberman85:
A few things that were unclear is whether or not the document tree stuff is intended to handle all the traversal of the tree or if unpack should call unpack on all children.

Another thing, which may be premature optimization is if we don't verify the signature at unpacking time, it means we would need to go back and verify the signature later and cascade the info down the tree

lumjjb:
>A few things that were unclear is whether or not the document tree stuff is intended to handle all the traversal of the tree or if unpack should call unpack on all children.

Only 1 round of unpacking. Since the document types that result from that may not a format they can handle or unknown. The tree construction is handled by the processor code.

>Another thing, which may be premature optimization is if we don't verify the signature at unpacking time, it means we would need to go back and verify the signature later and cascade the info down the tree

Yea - i think that would be a validation step as the ingestor is working through the documents. If it fails a validation step on one of the nodes then it doesn't walk the rest of that branch.

pkg/handler/processor/dsse/dsse_test.go
```diff
@@ -0,0 +1,138 @@
...
+		name:      "DSSE Envelope with Unknown Payload",
+		doc:       unknownDSSEDoc,
+		expected:  []*processor.Document{&unpackedUnknownDSSEDoc},
+		expectErr: false,
+	}, {
+		name:      "DSSE Envelope with ITE6",
+		doc:       ite6DSSEDoc,
+		expected:  []*processor.Document{&ite6SLSADoc},
+		expectErr: false,
+	}}
```
lumjjb, line 125: is there a scenario for DSSE where unpack fails , and unpacks a list of len==0.

pkg/handler/processor/dsse/dsse_test.go
```diff
@@ -0,0 +1,138 @@
...
+		Blob:   []byte(ite6SLSA),
+		Type:   processor.DocumentITE6,
+		Format: processor.FormatJSON,
+		SourceInformation: processor.SourceInformation{
+			Collector: "TestCollector",
+			Source:    "TestSource",
+		},
+	}
+)
+
```
lumjjb, line 108: can we add a simple ValidateSchema test with good schema and bad schema (which should fulfill one case in the unpack tests as well)

lumjjb:
minor requests, else LGTM

pkg/handler/processor/dsse/dsse.go
```diff
@@ -0,0 +1,94 @@
...
+
+	var doc *processor.Document
+	decodedPayload, err := base64.StdEncoding.DecodeString(envelope.Payload)
+	if err != nil {
+		return nil, err
+	}
+	switch pt := envelope.PayloadType; pt {
+	case string(dsseITE6):
+		doc = &processor.Document{
+			Blob:              decodedPayload,
```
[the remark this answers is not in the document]
pxp928: I see that its stored via the doc tree above. Nevermind.

pkg/handler/processor/dsse/dsse_test.go
mlieberman85: Will add

pkg/handler/processor/dsse/dsse_test.go
mlieberman85: Yep, will add

mihaimaruseac (approved):

lumjjb (approved):
LGTM

PR #30, opened 2022-08-24
Add initial graphdb package

### LINKED ISSUE #10: task: [assembler] create graphDB package for neo4j
Create a graph DB package that will be to create an instance of the neo4j/cypher driver to talk to the graph. No need for plug-ability of graph DB for now, since we currently do not foresee supporting additional graph DBs in the near future.

mihaimaruseac:
For now, this handles mostly testing interface. Given the way neo4j works, we'll need to add multiple functions as we keep adding more node/edge types and more queries. But for now this is ok.

Fixes #10


pkg/assembler/graphdb/graphdb_test.go
```diff
@@ -0,0 +1,124 @@
...
+		client,
+		"MATCH (a:Msg) RETURN a.text",
+		nil)
+	if err != nil {
+		t.Fatalf("Could not retrieve nodes: %v", err)
+	}
+	returned_nodes := make([]string, len(nodes))
+	for _, node := range result {
+		// each node is a list of 1 string, collect the strings
+		returned_nodes = append(returned_nodes, node[0].(string))
```
lumjjb, line 93: not sure if this is what you were intending, `make([]string, n)` makes a slice of size `n` of zero values, and appending to it appends after the slice to create an `n+1` size slice. `make([]string,0,n)` may do what you're thinking about, or in this case, can use `i` in `range result`.

pkg/assembler/graphdb/graphdb_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	result, err := ReadQueryForTesting(
+		client,
+		"MATCH (a:Msg) RETURN a.text",
+		nil)
+	if err != nil {
+		t.Fatalf("Could not retrieve nodes: %v", err)
+	}
+	returned_nodes := make([]string, len(nodes))
```
lumjjb, line 90: should this be `len(result)`?

pkg/assembler/graphdb/graphdb_test.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, node := range result {
+		// each node is a list of 1 string, collect the strings
+		returned_nodes = append(returned_nodes, node[0].(string))
+	}
+	for _, node := range nodes {
+		found := false
+		for _, returned_node := range returned_nodes {
```
lumjjb, line 97: nit: golang style uses camlCase

lumjjb (approved):
minor nits, else LGTM

pkg/assembler/graphdb/graphdb.go
```diff
@@ -0,0 +1,129 @@
...
+		driver.Close()
+		return nil, err
+	}
+	return driver, nil
+}
+
+// A transaction for the database
+type Transaction = neo4j.Transaction
+// Work done inside a transaction for the database
+type TransactionWork = neo4j.TransactionWork
```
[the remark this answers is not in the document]
mihaimaruseac: I just wanted to define aliases to allow for changing to the next golang neo4j version where `context.Context` will be used.

pkg/assembler/graphdb/graphdb.go
```diff
@@ -0,0 +1,129 @@
...
+
+// TODO(mihaimaruseac): Define queries needed for GUAC production
+
+// Runs a simple write query against the graph database.
+// Use only for testing!
```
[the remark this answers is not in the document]
mihaimaruseac: Sure, moved

pkg/assembler/graphdb/graphdb_test.go
mihaimaruseac: Right, typo from previous version

pkg/assembler/graphdb/graphdb_test.go
mihaimaruseac: Oh, that's better. I forgot why I was appending first, and probably there's a better way to write the test (we got 2 nodes and they are exactly the nodes we inserted).

pkg/assembler/graphdb/graphdb_test.go
mihaimaruseac: Thanks. Fixed everywhere.

Issue #41, opened 2022-08-29
Logging library not being developed anymore


rgreinho:
For logging, the [logrus](https://github.com/Sirupsen/logrus) library is being used. However is it not actively developed anymore:

> Logrus is in maintenance-mode. We will not be introducing new features. It's simply too hard to do in a way that won't break many people's projects, which is the last thing you want from your Logging library (again...).

They recommend using other libraries:

> Check out, for example, [Zerolog], [Zap], and [Apex].

[zerolog]: https://github.com/rs/zerolog
[zap]: https://github.com/uber-go/zap
[apex]: https://github.com/apex/log

lumjjb:
i like zap!

PR #42, opened 2022-08-29
Replace logging library

rgreinho:
Replaces the Sirupsen logging library with Zap.

Drive-by:
- Adds a .gitignore file

Fixes guacsec/guac#41

### DIFF
--- a/cmd/collector/cmd/example.go
+++ b/cmd/collector/cmd/example.go
@@ -18,12 +18,11 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/guacsec/guac/cmd/collector/cmd/mockcollector"
 	"github.com/guacsec/guac/pkg/handler/collector"
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/sirupsen/logrus"
+	"github.com/guacsec/guac/pkg/logging"
 	"github.com/spf13/cobra"
 )
 
@@ -34,18 +33,20 @@ var exampleCmd = &cobra.Command{
 		// Do Stuff Here
 		fmt.Println("GUAC example mock collector, it will emit 5 documents within 5 second intervals")
 
-		ctx := context.Background()
+		ctx := logging.WithLogger(context.Background())
+		logger := logging.FromContext(ctx)
 
 		// Register collector
 		// We create our own MockCollector and register it
 		mc := mockcollector.NewMockCollector()
-		collector.RegisterDocumentCollector(mc, mc.Type())
+		if err := collector.RegisterDocumentCollector(mc, mc.Type()); err != nil {
+			logger.Fatal(err)
+		}
 
 		// Collect
 		docChan, errChan, numCollectors, err := collector.Collect(ctx)
 		if err != nil {
-			logrus.Error(err)
-			os.Exit(1)
+			logger.Fatal(err)
 		}
 
 		collectorsDone := 0
@@ -53,12 +54,12 @@ var exampleCmd = &cobra.Command{
 			select {
 			case d := <-docChan:
 				emit(d)
-				logrus.Infof("emitted document: %+v", d)
+				logger.Infof("emitted document: %+v", d)
 			case err = <-errChan:
 				if err != nil {
-					logrus.Errorf("collector ended with error: %v", err)
+					logger.Errorf("collector ended with error: %v", err)
 				} else {
-					logrus.Info("collector ended gracefully")
+					logger.Info("collector ended gracefully")
 				}
 				collectorsDone += 1
 			}
@@ -68,7 +69,7 @@ var exampleCmd = &cobra.Command{
 		for len(docChan) > 0 {
 			d := <-docChan
 			emit(d)
-			logrus.Infof("emitted document: %+v", d)
+			logger.Infof("emitted document: %+v", d)
 		}
 	},
 }
--- a/pkg/handler/collector/collector.go
+++ b/pkg/handler/collector/collector.go
@@ -17,9 +17,9 @@ package collector
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/sirupsen/logrus"
 )
 
 const (
@@ -42,11 +42,13 @@ var (
 	documentCollectors = map[string]Collector{}
 )
 
-func RegisterDocumentCollector(c Collector, collectorType string) {
+func RegisterDocumentCollector(c Collector, collectorType string) error {
 	if _, ok := documentCollectors[collectorType]; ok {
-		logrus.Warnf("the document collector is being overwritten: %s", collectorType)
+		return fmt.Errorf("the document collector is being overwritten: %s", collectorType)
 	}
 	documentCollectors[collectorType] = c
+
+	return nil
 }
 
 // Collect takes all the collectors and starts collecting artifacts
--- a/pkg/handler/processor/guesser/format_guesser.go
+++ b/pkg/handler/processor/guesser/format_guesser.go
@@ -16,12 +16,13 @@
 package guesser
 
 import (
+	"fmt"
+
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/sirupsen/logrus"
 )
 
 func init() {
-	RegisterDocumentFormatGuesser(&jsonFormatGuesser{}, "json")
+	_ = RegisterDocumentFormatGuesser(&jsonFormatGuesser{}, "json")
 }
 
 // DocumentFormatGuesser guesses the format of the document given a blob
@@ -35,9 +36,11 @@ var (
 	documentFormatGuessers = map[string]DocumentFormatGuesser{}
 )
 
-func RegisterDocumentFormatGuesser(g DocumentFormatGuesser, name string) {
+func RegisterDocumentFormatGuesser(g DocumentFormatGuesser, name string) error {
 	if _, ok := documentFormatGuessers[name]; ok {
-		logrus.Warnf("the document type guesser is being overwritten: %s", name)
+		return fmt.Errorf("the document type guesser is being overwritten: %s", name)
 	}
 	documentFormatGuessers[name] = g
+
+	return nil
 }
--- a/pkg/handler/processor/guesser/type_guesser.go
+++ b/pkg/handler/processor/guesser/type_guesser.go
@@ -16,13 +16,14 @@
 package guesser
 
 import (
+	"fmt"
+
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/sirupsen/logrus"
 )
 
 func init() {
-	RegisterDocumentTypeGuesser(&ite6TypeGuesser{}, "ite6")
-	RegisterDocumentTypeGuesser(&dsseTypeGuesser{}, "dsse")
+	_ = RegisterDocumentTypeGuesser(&ite6TypeGuesser{}, "ite6")
+	_ = RegisterDocumentTypeGuesser(&dsseTypeGuesser{}, "dsse")
 }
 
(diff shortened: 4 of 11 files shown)


cmd/collector/cmd/example.go
```diff
@@ -36,15 +36,20 @@ var exampleCmd = &cobra.Command{
 
 		ctx := context.Background()
 
+		// Prepapre the logger.
+		zapLogger, _ := zap.NewProduction()
```
pxp928, line 40: I wonder if we should create a separate package for the logger that does the initialization. Similar to how Knative does it https://github.com/knative/pkg/blob/main/logging/logger.go

pxp928: Or just utilize it from an upstream project.

rgreinho: Yeah, that looks nice. The Zap documentation is definitely not the best in the world and they do not mention anything about the best practices to initialize the logger. But if we find a good way to do so, I am happy to implement it.

mihaimaruseac (commented):
In general it looks good, but I'm worried that now all APIs will need to take a `logger` parameter. Wondering if we can make logging an aspect, a singleton class that would be initialized when we import the logging package for the first time (not sure if this would be possible in Go though, still learning Go)

jchestershopify:
@mihaimaruseac a typical pattern is that functions that work together pass a `context.Context` as their first argument, and cross-cutting objects like loggers get attached to it.

mihaimaruseac:
Thanks, that makes sense (and is similar to C's pattern of opaque pointers). Thanks for explaining!

rgreinho:
> @mihaimaruseac a typical pattern is that functions that work together pass a `context.Context` as their first argument, and cross-cutting objects like loggers get attached to it.

So something like my latest update? That looks very ugly, so I am pretty sure I missed something... Would you be able to clarify or provide an example please?

jchestershopify:
Somewhat, yes. One difference is about what to do if the context didn't carry a logger. If you look at the Knative example posted upthread, they provide the "fallback" logger at the time the process launches, rather than creating one at the time the logger is first loaded within a package. Then you can look for a "fallback" string in your logs to determine that somewhere in your code, a context object is missing a logger that it should have.

That said, I am a fairly so-so Go programmer, I'd want a second opinion.

rgreinho:
I think I got it right.

pkg/logging/logger.go
```diff
@@ -0,0 +1,36 @@
...
+package logging
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type loggerKey struct{}
+
```
pxp928, line 25: Should add in the initialization here via `func init()` so that you can remove it downstream:
```
	// Prepapre the logger.
	zapLogger, _ := zap.NewProduction()
	defer zapLogger.Sync()
	logger := zapLogger.Sugar()
	ctx := logging.WithLogger(context.Background(), logger)
```

pkg/logging/logger.go
```diff
@@ -0,0 +1,36 @@
...
+package logging
+
+import (
+	"context"
+
+	"go.uber.org/zap"
+)
+
+type loggerKey struct{}
```
pxp928, line 24: Initialize the logger here:
```
var logger *zap.SugaredLogger

type loggerKey struct{}

func init() {
	zapLogger, _ := zap.NewProduction()
	defer zapLogger.Sync()
	logger = zapLogger.Sugar()
}

func WithLogger(ctx context.Context) context.Context {
	return context.WithValue(ctx, loggerKey{}, logger)
}
```
and upstream you can change to:
```
```

rgreinho: Addressed ✅

lumjjb:
I'm a little unfortable seeing a lot of the smaller functions that do logging require passing of context... 

I think either we would remove the logging messages in them or, do a global logger like: https://github.com/anchore/syft/blob/main/internal/log/log.go.

rgreinho:
Alright, so what are we deciding for this issue?

cpendery:
> I'm a little uncomfortable seeing a lot of the smaller functions that do logging require passing of context... Mainly looking at RegisterXXX, i think for those we can remove the logging, since it's weird to initialize the logger on init with a totally different context
> 
> I think either we would remove the logging messages in them or, do a global logger like: https://github.com/anchore/syft/blob/main/internal/log/log.go.

Even Syft still uses logrus as its logging library. While I understand wanting to use an actively developed library, I feel like we won't need any logging functionality that logrus doesn't already provide

pkg/handler/processor/guesser/format_guesser.go
```diff
@@ -16,12 +16,15 @@
...
+
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/sirupsen/logrus"
+	"github.com/guacsec/guac/pkg/logging"
 )
 
 func init() {
-	RegisterDocumentFormatGuesser(&jsonFormatGuesser{}, "json")
+	ctx := logging.WithLogger(context.Background())
+	RegisterDocumentFormatGuesser(ctx, &jsonFormatGuesser{}, "json")
```
lumjjb, line 27: I'm concerned of the use of `ctx` here because it depends on the order of `init()` which is not ideal. 

i think let's remove the `ctx` here and make `RegisterDocumentFormatGuesser` return an error with a comparable error enum, and not log errors during init? Then should be good to go.

lumjjb: yea it looks like this behavior is happening and is caught by the CI https://github.com/guacsec/guac/runs/8139988072?check_suite_focus=true

rgreinho: Yeah, it is because I removed the `init()` functions, so we end up with a null pointer exception...

rgreinho: I put it back, now the CI is happy 😀

pkg/handler/processor/process/process_test.go
```diff
@@ -498,9 +498,17 @@ func Test_SimpleDocProcessTest(t *testing.T) {
...
 
-	logrus.SetLevel(logrus.DebugLevel)
+	zapLogger, _ := zap.NewDevelopmentConfig().Build()
+
+	// flushes buffer, if any
+	defer func() {
+		// intentionally ignoring error here, see https://github.com/uber-go/zap/issues/328
+		_ = zapLogger.Sync()
+	}()
+	logger := zapLogger.Sugar()
```
lumjjb, line 508: nit: can we have a helper function in  to do this in logger library?

pkg/handler/processor/process/process.go
```diff
@@ -92,7 +94,7 @@ func processDocument(i *processor.Document) ([]*processor.Document, error) {
 }
 
 func preProcessDocument(i *processor.Document) error {
-	docType, format, err := guesser.GuessDocument(i)
+	docType, format, err := guesser.GuessDocument(logging.WithLogger(context.Background()), i)
```
lumjjb, line 97: this should probably propagate context all the way to the top of what is calling this since the logger will be initialized by that.. using context like this doesn't quite match context usage patterns.

pkg/handler/processor/process/process_test.go
```diff
@@ -498,9 +498,17 @@ func Test_SimpleDocProcessTest(t *testing.T) {
...
+
 	// Register
-	RegisterDocumentProcessor(&simpledoc.SimpleDocProc{}, simpledoc.SimpleDocType)
+	RegisterDocumentProcessor(&simpledoc.SimpleDocProc{}, simpledoc.SimpleDocType, logger)
```
lumjjb, line 511: have this not take in a logger and return error instead? With an err that can be compared to check if it is a warning for a overwrite of a registration.

pkg/handler/processor/guesser/type_guesser.go
```diff
@@ -38,7 +39,7 @@ var (
 
 func RegisterDocumentTypeGuesser(g DocumentTypeGuesser, name string) {
 	if _, ok := documentTypeGuessers[name]; ok {
-		logrus.Warnf("the document type guesser is being overwritten: %s", name)
+		fmt.Printf("the document type guesser is being overwritten: %s", name)
```
lumjjb, line 42: all `Register*` functions should probably have the same pattern, we should stick to one - i don't this we should have warning messages print to stdout using `fmt`.. I think returning an err may be a good option here (see comment below).

pkg/handler/processor/process/process_test.go
rgreinho: Addresed ✅

pkg/handler/processor/guesser/type_guesser.go
rgreinho: Addresed ✅

pkg/handler/processor/process/process.go
rgreinho: you mean from `Process()` down to this function?

lumjjb: yup!

rgreinho: Done ✅

pkg/handler/collector/collector.go
```diff
@@ -42,9 +42,10 @@ var (
 	documentCollectors = map[string]Collector{}
 )
 
-func RegisterDocumentCollector(c Collector, collectorType string) {
+func RegisterDocumentCollector(ctx context.Context, c Collector, collectorType string) {
```
pxp928, line 45: Just one update. All the other `Register` functions are returning an error. So lets make this one also return an error to be consistent.

pkg/handler/processor/guesser/format_guesser.go
```diff
@@ -37,7 +38,7 @@ var (
 
 func RegisterDocumentFormatGuesser(g DocumentFormatGuesser, name string) {
```
pxp928, line 39: same with this one. Should just return an error.

pxp928 (commented):
fix quick changes but seems like this will work.

pkg/handler/processor/guesser/format_guesser.go
rgreinho: Addressed ✅

pkg/handler/processor/guesser/format_guesser.go
```diff
@@ -35,9 +36,11 @@ var (
 	documentFormatGuessers = map[string]DocumentFormatGuesser{}
 )
 
-func RegisterDocumentFormatGuesser(g DocumentFormatGuesser, name string) {
+func RegisterDocumentFormatGuesser(g DocumentFormatGuesser, name string) error {
 	if _, ok := documentFormatGuessers[name]; ok {
-		logrus.Warnf("the document type guesser is being overwritten: %s", name)
+		fmt.Printf("the document type guesser is being overwritten: %s", name)
```
pxp928, line 41: same with this one. Just return an error and upstream we can check if it is getting "overwritten" and log there.

pxp928 (approved):
LGTM

pkg/handler/processor/process/process.go
```diff
@@ -91,8 +93,8 @@ func processDocument(i *processor.Document) ([]*processor.Document, error) {
 	return ds, nil
 }
 
-func preProcessDocument(i *processor.Document) error {
-	docType, format, err := guesser.GuessDocument(i)
+func preProcessDocument(ctx context.Context, i *processor.Document) error {
+	docType, format, err := guesser.GuessDocument(logging.WithLogger(ctx), i)
```
lumjjb, line 97: WithLogger shouldn’t need to be done here, it should already be initialized from the context passed in.

lumjjb (approved):
minor change requested, else lgtm

lumjjb:
fyi https://github.com/golang/go/discussions/54763

pkg/handler/processor/process/process.go
rgreinho: Addressed ✅

Issue #33, opened 2022-08-25
task: [processor] create DSSE DocumentTypeGuesser


lumjjb:
As part of #26, we want to be able to create guessers to identify what format type a document is (using the guesser interface defined in https://github.com/guacsec/guac/tree/main/pkg/handler/processor/guesser). Based on the foundations laid in #27 .

This issue is to create DSSE document guesser.

PR #44, opened 2022-08-29
feat: implement dsse type guesser

cpendery:
Closes https://github.com/guacsec/guac/issues/33

### DIFF
--- a/pkg/handler/processor/guesser/type_dsse.go
+++ b/pkg/handler/processor/guesser/type_dsse.go
@@ -24,16 +24,10 @@ import (
 
 type dsseTypeGuesser struct{}
 
-type envelope struct {
-	PayloadType *string           `json:"payloadType"`
-	Payload     *string           `json:"payload"`
-	Signatures  *[]dsse.Signature `json:"signatures"`
-}
-
 func (_ *dsseTypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
-	var envelope envelope
+	var envelope dsse.Envelope
 	if json.Unmarshal(blob, &envelope) == nil && format == processor.FormatJSON {
-		if envelope.Payload != nil && envelope.PayloadType != nil && envelope.Signatures != nil {
+		if envelope.Payload != "" && envelope.PayloadType != "" {
 			return processor.DocumentDSSE
 		}
 	}
--- a/pkg/handler/processor/guesser/type_dsse.go
+++ b/pkg/handler/processor/guesser/type_dsse.go
@@ -1,0 +1,41 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package guesser
+
+import (
+	"encoding/json"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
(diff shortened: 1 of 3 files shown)


pkg/handler/processor/guesser/type_dsse.go
```diff
@@ -0,0 +1,41 @@
...
+type dsseTypeGuesser struct{}
+
+type envelope struct {
+	PayloadType *string           `json:"payloadType"`
+	Payload     *string           `json:"payload"`
+	Signatures  *[]dsse.Signature `json:"signatures"`
+}
+
+func (_ *dsseTypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
+	var envelope envelope
```
pxp928, line 34: Replace with `dsse.Envelope{}` see https://github.com/guacsec/guac/blob/main/pkg/handler/processor/dsse/dsse.go#L88.

pxp928 (changes requested):
Thanks @cpendery! Quick update

pkg/handler/processor/guesser/type_dsse.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"encoding/json"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
+)
+
+type dsseTypeGuesser struct{}
+
+type envelope struct {
```
lumjjb, line 27: can we use https://github.com/secure-systems-lab/go-securesystemslib/blob/main/dsse/sign.go#L28?

lumjjb (changes requested):

pkg/handler/processor/guesser/type_dsse.go
cpendery: @lumjjb @pxp928 if we use the `dsse.Envelope` struct, we would have to assert that the fields in the struct are not empty strings. Would we want the envelope below to be marked as `DocumentUnknown`? If so, I'll make the switch
```json
{
  "payload": "",
  "payloadType": "",
  "signatures": [...]
}
```

pxp928: Yes that is true. Otherwise it would be a malformed DSSE envelope.

lumjjb (approved):
lgtm

pkg/handler/processor/guesser/type_dsse_test.go
```diff
@@ -0,0 +1,44 @@
+package guesser
```
lumjjb, line 1: add copyright header

lumjjb (changes requested):

lumjjb (approved):
LGTM

lumjjb:
@pxp928 for your re-review!

pkg/handler/processor/guesser/type_dsse.go
```diff
@@ -0,0 +1,35 @@
...
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
+)
+
+type dsseTypeGuesser struct{}
+
+func (_ *dsseTypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
+	var envelope dsse.Envelope
+	if json.Unmarshal(blob, &envelope) == nil && format == processor.FormatJSON {
+		if envelope.Payload != "" && envelope.PayloadType != "" {
```
pxp928, line 30: We should also check that there is atleast one signature for a valid payload

pxp928 (changes requested):

pkg/handler/processor/guesser/type_dsse.go
lumjjb: yea.. i don't like how there isn't really a good way to guess that it's a DSSE based on the fields (they are all so generic). Yea I think its fair to just check that there's at least 1 signature. @SantiagoTorres i'm guessing this is for space reasons? but may be weird without versioning if there's a version 2.0?

(a minor nit here is that it doesn't actually have to be a valid envelope, but a side effect of it could be - since schema validation is handled again by the processor).

pxp928 (approved):
LGTM

Issue #11, opened 2022-08-22
task: [assembler] define minimal set of GuacNodes and GuacEdges


lumjjb:
Define an initial set of GuacNodes and GuacEdges to satisfy the basic example of SLSA attestations based on @mihaimaruseac 's script and @mlieberman85 's data.

PR #46, opened 2022-08-29
Add nodes for attestations, artifacts and builders

mihaimaruseac:
Implements most of #11.

Still working on the testing interface for this as I also want to be able to write and read subgraphs from DB (#10/#12/#14).

### DIFF
--- a/pkg/assembler/assembler.go
+++ b/pkg/assembler/assembler.go
@@ -17,6 +17,52 @@ package assembler
 
 type assembler struct{} //nolint: unused
 
+// NOTE: `GuacNode` and `GuacEdge` interfaces are very experimental and might
+// change in the future as we discover issues with reading/writing from the
+// graph database.
+//
+// For now, the design of the interface follows these guideliness:
+//
+//   1. We want to serialize `GuacNode`s and `GuacEdge`s to graph database
+//      (e.g. Neo4j) without creating duplicate nodes. To do this, we need
+//      ability to uniquely identify a node. Since a node could be created from
+//      different document types, it can be uniquely identified by different
+//      subsets of attributes/properties. For example, we could have a node
+//      that is identified by an `"id"` field from one document and by the pair
+//      `"name"`, `"digest"` from another one.
+//   2. Nodes can also have attributes that are not unique and are generated
+//      from various documents.
+//   3. In order to write the serialization/deserialization code, we need to
+//      get the name of the attributes separate from the pairing between the
+//      attribute and the value.
+//
+// In broad lines, the serialization process for a node would look like:
+//
+//   1. For each identifiable set in `IdentifiablePropertyNames()` check if the
+//      node has values for all of the specified properties. If one is missing,
+//      try the next set. If no set is left, panic.
+//   2. If a set of identifiable properties is found and we have values for all
+//      of these, write a query that would match on nodes which have these
+//      property:value attributes. The graph database engine will allow us to
+//      run separate code if a node already exists or one is newly created. In
+//      our case, in both instances we will just need to set the other
+//      attributes that have a value. To do this, the `Properties()` returned
+//      map will be passed directly to the prepared statement (which uses
+//      `Type()` to select the graph database node type and `PropertyNames()`
+//      to build the rest of the query).
+//
+// The serialization process for an edge would be similar, with the caveat that
+// an edge is always created between two existing nodes.
+//
+// Deserialization is left for later, with the only caveat that we might
+// envision a case where we'd like to match on edges without first matching on
+// their endpoints (e.g., "retrieve all attestations from this time period and
+// for each of them return the artifact nodes"). Hence, we need ways to
+// uniquely identify edges without having endpoint nodes.
+//
+// TODO(mihaimaruseac): Look into using tags of fields to automate
+// serialization/deserialization, similar to how json is done.
+
 // GuacNode represents a node in the GUAC graph
 type GuacNode interface {
 	// Type returns the type of node
--- a/pkg/assembler/assembler.go
+++ b/pkg/assembler/assembler.go
@@ -21,7 +21,7 @@ type assembler struct{} //nolint: unused
 // change in the future as we discover issues with reading/writing from the
 // graph database.
 //
-// For now, the design of the interface follows these guideliness:
+// For now, the design of the interface follows these guidelines:
 //
 //   1. We want to serialize `GuacNode`s and `GuacEdge`s to graph database
 //      (e.g. Neo4j) without creating duplicate nodes. To do this, we need
@@ -64,6 +64,8 @@ type assembler struct{} //nolint: unused
 // serialization/deserialization, similar to how json is done.
 
 // GuacNode represents a node in the GUAC graph
+// Note: this is experimental and might change. Please refer to source code for
+// more details about usage.
 type GuacNode interface {
 	// Type returns the type of node
 	Type() string
@@ -86,6 +88,8 @@ type GuacNode interface {
 }
 
 // GuacEdge represents an edge in the GUAC graph
+// Note: this is experimental and might change. Please refer to source code for
+// more details about usage.
 type GuacEdge interface {
 	// Type returns the type of edge
 	Type() string
@@ -117,6 +121,8 @@ type GuacEdge interface {
 }
 
 // Subgraph represents a subgraph read from the database or written to it.
+// Note: this is experimental and might change. Please refer to source code for
+// more details about usage.
 type Subgraph struct {
 	V []GuacNode
 	E []GuacEdge
--- a/pkg/assembler/assembler.go
+++ b/pkg/assembler/assembler.go
@@ -17,43 +17,66 @@ package assembler
 
 type assembler struct{} //nolint: unused
 
-// Identifiable implements the ability to retrieve a set of
-// attributes such that a graph query is able to identify a
-// GuacNode or GuacEdge uniquely (or as a GuacHyperNode).
-type Identifiable interface {
-	// Identifiers returns a map of fields and values which
-	// can be used to identify an object in the graph.
-	Identifiers() map[string]interface{}
-}
-
 // GuacNode represents a node in the GUAC graph
 type GuacNode interface {
-	Identifiable
-
 	// Type returns the type of node
 	Type() string
 
 	// Properties returns the list of properties of the node
 	Properties() map[string]interface{}
+
+	// Attributes returns the names of the properties of the node.
+	//
+	// If a string `s` is in the list returned by `Attributes` then it
+	// should also be a key in the map returned by `Properties`.
+	Attributes() []string
+
+	// IdentifiableAttributes returns a list of tuples of property names
+	// that can uniquely specify a GuacNode.
+	//
+	// Any string found in a tuple returned by `IdentifiableAttributes`
+	// must also be returned by `Attributes`.
+	IdentifiableAttributes() [][]string
 }
 
 // GuacEdge represents an edge in the GUAC graph
 type GuacEdge interface {
-	Identifiable
-
-	// Nodes returns the (v,u) nodes of the edge
-	// where v--edge-->u for directional edges.
-	Nodes() (v, u GuacNode)
-
 	// Type returns the type of edge
 	Type() string
 
-	// Properties returns the list of properties of the node
+	// Nodes returns the (v,u) nodes of the edge
+	//
+	// For directional edges: v-[edge]->u.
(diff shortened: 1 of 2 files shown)


pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+
+func (an ArtifactNode) Properties() map[string]interface{} {
+	properties := make(map[string]interface{})
+	properties["name"] = an.Name
+	properties["checksum"] = an.Checksum
+	return properties
+}
+
+func (an ArtifactNode) Attributes() []string {
+	return []string{"name", "checksum"}
```
jchestershopify, line 36: I wonder if there is a way to look this up automatically? It seems like a future maintenance burden to need to know that you need to set `"name"` and `"checksum"` in two different places.

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+	// An artifact can be uniquely identified by checksum
+	return [][]string{{"checksum"}}
+}
 
 // AttestationNode is a node that represents an attestation
-// TODO: implement GuacNode
-type AttestationNode struct{}
+type AttestationNode struct {
+	// TODO(mihaimaruseac): Unsure what fields to store here
+	File     string
```
jchestershopify, line 47: Is this meant to be the contents of the file? A path? URI?

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+}
+
+func (bn BuilderNode) IdentifiableAttributes() [][]string {
+	// A builder needs both type and id to be identified
+	return [][]string{{"type", "id"}}
+}
+
+// AttestationEdge is an edge that represents the fact that an
+// `AttestationNode` is an attestation for an `ArtifactNode`.
+type AttestationEdge struct {
```
jchestershopify, line 99: Suggest `AttestationForEdge`, so that it can be read as "`AttestationNode AttestationForEdge ArtifactNode`". This would make it more consistent with `BuiltByEdge`, as in "`ArtifactNode BuiltByEdge BuilderNode`".

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+func (e DependsOnEdge) Properties() map[string]interface{} {
+	return map[string]interface{}{}
+}
+
+func (e DependsOnEdge) Attributes() []string {
+	return []string{}
+}
+
+func (e DependsOnEdge) IdentifiableAttributes() [][]string {
+	return [][]string{}
```
jchestershopify, line 175: Wouldn't the combination of `ArtifactNode` and `Dependency` be uniquely identifying? I'm not convinced but it's a pity to missing a primary key (even a synthetic one if necessary).

jchestershopify (commented):
Some thoughts; feel free to disregard if I am missing a larger context or design decision.

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,66 @@ package assembler
...
+	// If a string `s` is in the list returned by `Attributes` then it
+	// should also be a key in the map returned by `Properties`.
+	Attributes() []string
+
+	// IdentifiableAttributes returns a list of tuples of property names
+	// that can uniquely specify a GuacNode.
+	//
+	// Any string found in a tuple returned by `IdentifiableAttributes`
+	// must also be returned by `Attributes`.
+	IdentifiableAttributes() [][]string
```
pxp928, line 39: What is the thought process around `IdentifiableAttributes() [][]string`?  Wouldn't this also be just a []string?

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,66 @@ package assembler
...
+// Subgraph represents a subgraph read from the database or written to it.
+type Subgraph struct {
 	V []GuacNode
 	E []GuacEdge
 }
+
+// TODO(mihaimaruseac): Write queries to write/read subgraphs from DB?
+
+// AssemblerInput represents the inputs to add to the graph
+type AssemblerInput = Subgraph
```
pxp928, line 82: Why would these be equal?

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,66 @@ package assembler
...
 type GuacNode interface {
-	Identifiable
-
 	// Type returns the type of node
 	Type() string
 
 	// Properties returns the list of properties of the node
 	Properties() map[string]interface{}
+
+	// Attributes returns the names of the properties of the node.
```
lumjjb, line 28: These interfaces are likely going to have two types of uses, one is as a return to a query, where the returned object would act as an instance of a GuacNode. I think it's not super clear to me which of these functions relate to a schema vs relate to the actual instance of the object. 

i.e. should these calls return fields that may not be populated and Properties() return a map where they are set to the zero value? or should they return only if they are present. For the return value scenario i think it is pretty clear to me, but in the creation of an identifiable node for adding an edge for example, i still need to think how that would play out.

EDIT: now that i've seen the code below it seems like the former. That leaves the conversion to graph queries a little difficult now i think because its not certain which unique attributes exists (can't differentiate between missing or zero-value. e.g. int zero value is 0). Maybe `Get (name) error` instead of `Properties()`?.

pkg/assembler/assembler.go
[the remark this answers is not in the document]
lumjjb: i think the Identifiable interface does exist here, that would be passed to something that makes the cypher queries. (doesn't have to be a element of this interface, but seems like it could be)

lumjjb: I think the thought here is that you could have a hash or a purl, or something else.. And a document may only refer to one of them, so it does need to have an alternative to do the linking.

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
 package assembler
 
 // ArtifactNode is a node that represents an artifact
-// TODO: implement GuacNode
-type ArtifactNode struct{}
+type ArtifactNode struct {
+	Name     string
+	Checksum string
```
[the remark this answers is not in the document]
lumjjb: Digest should also possibly come with an alg, unless we want to force a certain string convention, also some of them may have multiple entries here.

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+}
+
+func (an ArtifactNode) Type() string {
+	return "Artifact"
```
lumjjb, line 25: nit: i have a slight preference for camel case or all lowercase-with-hypens.

lumjjb (commented):
Shaping up! Couple of comments around the defining of incomplete nodes (e.g. when creating edges between existing nodes where the ingestor doesn't have full information)

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+func (an AttestationNode) Attributes() []string {
+	return []string{"file", "checksum"}
+}
+
+func (an AttestationNode) IdentifiableAttributes() [][]string {
+	// An attestation can be uniquely identified by filename?
+	return [][]string{{"file"}}
+}
+
+// BuilderNode is a node that represents an attestation
```
mihaimaruseac, line 71:
```suggestion
// BuilderNode is a node that represents a builder for the artifact
```

pkg/assembler/assembler.go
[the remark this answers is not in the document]
mihaimaruseac: So I'm using this one to determine which attributes (in neo4j parlance) / property names (in the way the interface is written) are uniquely identifying a node to avoid duplicated nodes/edges. I'll add a comment describing this.

mihaimaruseac: We could have scenarios where a node can be uniquely identified by multiple sets of attributes. Each set can be a singleton or contain multiple attributes. E.g., I could identify a node by `id` or by the pair `name`, `hash`, so for that node the function should return `{{"id"},{"name","hash"}}`. I'll add a comment describing this.

pkg/assembler/assembler.go
mihaimaruseac: Type synonyms for now to not require changes in other files that are also under PR review. Will clean later

pkg/assembler/nodes.go
mihaimaruseac: Agree, this should be possible the same the json serialization/deserialization is done via tags on struct fields, but I didn't get a chance to play with those yet, punting for another PR.

Added a TODO

pkg/assembler/nodes.go
mihaimaruseac: It would be, but for edges this function allows for other ways to uniquely identify them.

For example, if we support a query of "for all attestations from this time window return all artifacts" if `AttestationForEdge` has the date as an attribute (instead of being stored in the attestation node) we could use that date to select the edge and then from there select the artifacts.

It is very likely that by the time we have an end-to-end example we realize that we don't need other ways to identify an edge except by its endpoints, at which point I'd remove this.

pkg/assembler/nodes.go
mihaimaruseac: Should be a path. Changed to `FilePath`.

In my demo for neo4j I used the path of the attestation as the only attribute for this node, but not completely sure that this is what we need.

pkg/assembler/nodes.go
```diff
@@ -16,9 +16,161 @@
...
+}
+
+func (an ArtifactNode) Type() string {
+	return "Artifact"
+}
+
+func (an ArtifactNode) Properties() map[string]interface{} {
+	properties := make(map[string]interface{})
+	properties["name"] = an.Name
+	properties["checksum"] = an.Checksum
```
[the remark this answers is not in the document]
mihaimaruseac: Yes, this is not ergonomic, but I hope to get rid of it when we do the serialization via tags on struct fields (don't know the technical name of it)

pkg/assembler/nodes.go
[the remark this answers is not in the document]
mihaimaruseac: Changed to Digest for now and will probably add the algorithm at a later stage.

pkg/assembler/nodes.go
mihaimaruseac: Neo4j uses all capitals for node/edge types, but that would be too much. Let's leave like this for now as it will probably get replaced with the serialization via field tags later.

pkg/assembler/assembler.go
mihaimaruseac: Oh, I'll need to think about missing fields. I picked `Properties` with this return type because this would have matched 1:1 to the type expected by the graphdb transactions.

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,112 @@ package assembler
...
-type Identifiable interface {
-	// Identifiers returns a map of fields and values which
-	// can be used to identify an object in the graph.
-	Identifiers() map[string]interface{}
-}
+// NOTE: `GuacNode` and `GuacEdge` interfaces are very experimental and might
+// change in the future as we discover issues with reading/writing from the
+// graph database.
+//
+// For now, the design of the interface follows these guideliness:
```
lumjjb, line 24: nit: typo guidelines

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,112 @@ package assembler
...
+// Deserialization is left for later, with the only caveat that we might
+// envision a case where we'd like to match on edges without first matching on
+// their endpoints (e.g., "retrieve all attestations from this time period and
+// for each of them return the artifact nodes"). Hence, we need ways to
+// uniquely identify edges without having endpoint nodes.
+//
+// TODO(mihaimaruseac): Look into using tags of fields to automate
+// serialization/deserialization, similar to how json is done.
 
 // GuacNode represents a node in the GUAC graph
```
lumjjb, line 66: (add a note here for experimental and please refer to source file for more details) - since the big comment block won't appear in the golang doc

pkg/assembler/assembler.go
```diff
@@ -17,43 +17,112 @@ package assembler
...
+	// IdentifiablePropertyNames returns a list of tuples of property names
+	// that can uniquely specify a GuacNode.
+	//
+	// Any string found in a tuple returned by `IdentifiablePropertyNames`
+	// must also be returned by `PropertyNames`.
+	IdentifiablePropertyNames() [][]string
 }
 
 // GuacEdge represents an edge in the GUAC graph
 type GuacEdge interface {
```
lumjjb, line 89: (add a note here for experimental and please refer to source file for more details) - since the big comment block won't appear in the golang doc

lumjjb (approved):
minor nits else lgtm

Issue #12, opened 2022-08-22
task: [ingestor] create SLSA document ingestor


lumjjb:
Create SLSA document ingestor off @mihaimaruseac 's script to take in SLSA documents as a list of `Documents` and outputs `GuacNodes` and `GuacEdges`, depends on https://github.com/guacsec/guac/issues/11

pxp928:
Depends on #36 where SLSA docs are processed.

PR #57, opened 2022-08-30
Serialize subgraph to graph database

### LINKED ISSUE #11: task: [assembler] define minimal set of GuacNodes and GuacEdges

mihaimaruseac:
Fixes #11, opens a way to ingest SLSA attestations all the way to the graph database (#12, though the tests for ingesting SLSA nodes is WIP for a future PR).

Changes the interfaces for `GuacNode` and `GuacEdges`, makes them even simpler than before.

Left to do for future PRs:

- more testing
- read from the database (for now we only see that the nodes have been created by running an external query, see below)
- convert the serialization to use struct fields tags, further simplifying `GuacNode` and `GuacEdges`

Testing code generates the following structure (running `match (a)-[e]->(b) return a,e,b`):

```
╒═══════════════════════════════════════════════════════╤════════╤═════════════════════════════════════════════════════════════════╕
│"a"                                                    │"e"     │"b"                                                              │
╞═══════════════════════════════════════════════════════╪════════╪═════════════════════════════════════════════════════════════════╡
│{"address":"addr1","name":"name1","id":"id1","age":42} │{"id":0}│{"score":0,"address":"addr1","name":"name2","id":"id2"}          │
├───────────────────────────────────────────────────────┼────────┼─────────────────────────────────────────────────────────────────┤
│{"score":0,"address":"addr1","name":"name2","id":"id2"}│{}      │{"score":42,"address":"addr2","name":"name3","id":"id3","age":42}│
└───────────────────────────────────────────────────────┴────────┴─────────────────────────────────────────────────────────────────┘
```

Or, as an image:
9xp33CstArXvJux

### DIFF
--- a/pkg/assembler/graphdb.go
+++ b/pkg/assembler/graphdb.go
@@ -19,8 +19,8 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
 	"github.com/guacsec/guac/pkg/assembler/graphdb"
+	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
 )
 
 // Note: This module is experimental and might change often!
@@ -42,7 +42,7 @@ func StoreGraph(g Graph, client graphdb.Client) error {
 		node_queries[i] = sb.String()
 		node_dicts[i] = map[string]interface{}{}
 		for k, v := range n.Properties() {
-			node_dicts[i]["n_" + k] = v
+			node_dicts[i]["n_"+k] = v
 		}
 	}
 
@@ -61,20 +61,20 @@ func StoreGraph(g Graph, client graphdb.Client) error {
 		edge_queries[i] = sb.String()
 		edge_dicts[i] = map[string]interface{}{}
 		for k, v := range a.Properties() {
-			edge_dicts[i]["a_" + k] = v
+			edge_dicts[i]["a_"+k] = v
 		}
 		for k, v := range b.Properties() {
-			edge_dicts[i]["b_" + k] = v
+			edge_dicts[i]["b_"+k] = v
 		}
 		for k, v := range e.Properties() {
-			edge_dicts[i]["e_" + k] = v
+			edge_dicts[i]["e_"+k] = v
 		}
 	}
 
 	queries := append(node_queries, edge_queries...)
 	params := append(node_dicts, edge_dicts...)
 	_, err := session.WriteTransaction(
-		func (tx graphdb.Transaction) (interface{}, error) {
+		func(tx graphdb.Transaction) (interface{}, error) {
 			for i, query := range queries {
 				fmt.Printf("%v(where: %v)\n\n", query, params[i])
 				if _, err := tx.Run(query, params[i]); err != nil {
@@ -91,7 +91,7 @@ func StoreGraph(g Graph, client graphdb.Client) error {
 func queryPartForMergeNode(sb *strings.Builder, n GuacNode, label string) error {
 	node_data := n.Properties()
 	sb.WriteString("MERGE (")
-	sb.WriteString(label)    // not user controlled
+	sb.WriteString(label) // not user controlled
 	sb.WriteString(":")
 	sb.WriteString(n.Type()) // not user controlled
 	sb.WriteString(" {")
--- a/pkg/assembler/graphdb.go
+++ b/pkg/assembler/graphdb.go
@@ -1,0 +1,153 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


pkg/assembler/graphdb.go
```diff
@@ -0,0 +1,148 @@
...
+	} else {
+		sb.WriteString(":")
+	}
+	switch val.(type) {
+	case string:
+		sb.WriteString("\"")
+		sb.WriteString(val.(string))
+		sb.WriteString("\"")
+	default:
+		sb.WriteString(fmt.Sprint(val))
```
mihaimaruseac, line 146: Go experts: I need a way to mark this accepted by the linter. If I do

```
switch v := val.(type); v {
```

I get an AST error.

If I do

```
switch v := val.(type) {
```

I get that `v` is not used, and the only place I could use it is to replace `(string)` with `(v)` but then `v` is a string, so of course this doesn't work.

lumjjb: i think you can do `switch val.(type) {`

lumjjb: ahh i see what you're doing...

Replacing `val` with `v` in the case statements should fix it

pkg/assembler/graphdb.go
```diff
@@ -0,0 +1,148 @@
...
+	"strings"
+
+	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
+	"github.com/guacsec/guac/pkg/assembler/graphdb"
+)
+
+// Note: This module is experimental and might change often!
+
+// StoreSubgraph stores a Graph to the graph database given by Client
+func StoreGraph(g Graph, client graphdb.Client) error {
```
lumjjb, line 29: hmm, thinking if this would be a good candidate for a function receiver for `graphdb.Client`, so defining an interface for `GraphDB` and doing it as function receivers in  `github.com/guacsec/guac/pkg/assembler/graphdb/neo4j`?

pkg/assembler/graphdb.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return nil
+}
+
+// Creates the "ON CREATE SET ${ATTR}=${VALUE}, ..." part of the query
+// Creates the "ON MATCH SET ${ATTR}=${VALUE}, ..." part of the query
+func queryPartForNodeAttributes(sb *strings.Builder, when string, n GuacNode, label string) {
+	node_data := n.Properties()
+	sb.WriteString("ON ")
+	sb.WriteString(when)
```
lumjjb, line 98: this seems fairly dangerous to me.. my injection alarms are going off, but maybe the sanitization is done somewhere else

pkg/assembler/graphdb.go
lumjjb, line 146: could potentially inject as well?

pkg/assembler/graphdb.go
```diff
@@ -0,0 +1,148 @@
...
+	sb.WriteString(key)
+	if set {
+		sb.WriteString("=")
+	} else {
+		sb.WriteString(":")
+	}
+	switch val.(type) {
+	case string:
+		sb.WriteString("\"")
+		sb.WriteString(val.(string))
```
lumjjb, line 143: could potentially inject as well?

pkg/assembler/graphdb.go
lumjjb: if there isn't a simple way to do this, we can defer as well and put a large disclaimer somewhere.

pkg/assembler/graphdb.go
mihaimaruseac: This is a possibility but has a potential issue that it would cause a dependency cycle between `graphdb/*.go` and `nodes.go` (or similar). I'd say, let's try this in a separate PR?

pkg/assembler/graphdb.go
mihaimaruseac: Yes, these 2 are issues. Thinking of a fix right now.

pkg/assembler/nodes_test.go
```diff
@@ -0,0 +1,146 @@
...
+
+type MockNode struct {
+	Id      string
+	Address string
+	Name    string
+	Age     *int
+	Score   *int
+}
+
+func (n MockNode) Type() string {
```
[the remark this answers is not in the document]
mihaimaruseac: Agree. The problem here is that if we make the receiver be a pointer then we won't be able to collect the nodes in a list for the `Graph` interface (`[]GuacNode`)

See line34 at https://go.dev/play/p/1SmYgbe0TyL, it gives an error

```
./prog.go:34:26: cannot use p (variable of type Ptr) as type I in array or slice literal:
	Ptr does not implement I (Foo method has pointer receiver)
```

pkg/assembler/graphdb.go
mihaimaruseac: This specifc case is ok as `when` is either `CREATE` or `MATCH`. Changed to a boolean flag though, to make it more readable.

pkg/assembler/graphdb.go
mihaimaruseac: I removed it completely :D

pkg/assembler/graphdb.go
mihaimaruseac: This is fixed now.

lumjjb (approved):
LGTM

PR #58, opened 2022-08-31
fix: fmt cmd & add copywrite check to fmt cmd

cpendery:
Fix the `fmt` command as it doesn't have the shell directive and add copywrite checks as part of the formatting check

### DIFF
--- a/.github/scripts/copywrite.sh
+++ b/.github/scripts/copywrite.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+#
+# Checks if a file has the correct copywrite
+#
+
+year=`date +%Y`
+header="//
+// Copyright ${year} The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the \"License\");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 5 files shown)


mihaimaruseac (changes requested):
Can you also add it to `.github/workflows/ci.yaml` so it runs in CI too?

cpendery:
> Can you also add it to `.github/workflows/ci.yaml` so it runs in CI too?

`make fmt` runs in the ci already, so this can be seen in https://github.com/guacsec/guac/runs/8104971397?check_suite_focus=true under `Check Format`

.github/scripts/copywrite.sh
```diff
@@ -0,0 +1,25 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an \"AS IS\" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License."
+content=`cat $1`
+if  [[ $content != $header* ]]; then
+    echo $1
+fi
```
rgreinho, line 25: Just a new line at the end of file missing. We should probably add '.editorconfig'to automatically solve this.

rgreinho (changes requested):
LGTM 👍

mihaimaruseac (approved):

PR #60, opened 2022-08-31
Add identity verifier interface

mlieberman85:
This will be useful if we verify identities through mechanisms
like GPG in addition to Sigstore

### DIFF
--- a/pkg/ingestor/verifier/verifier.go
+++ b/pkg/ingestor/verifier/verifier.go
@@ -1,0 +1,47 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verifier
+
+import (
+	"crypto"
+)
+
+// Verifier allows for multiple signature or identity verifiers that will
+// check envelopes, signatures, and other payloads for linking a payload
+// back to an identity. In most cases this will be done through verifying
+// one or more signatures against known public keys.
+type Verifier interface {
+	// Verify takes in bytes and returns a list of identities. Those
+	// identities are marked as verified if the payload can be tied back to
+	// an identity, usually through signature validation. In cases where
+	// signatures can't be verified, it is still valuable to return data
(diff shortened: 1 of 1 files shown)


pkg/ingestor/verifier/verifier.go
```diff
@@ -0,0 +1,42 @@
...
+	// about the unverified identity.
+	Verify(payloadBytes []byte) ([]Identity, error)
+	Type() string
+}
+
+// Elements of this struct might be nil/empty if the key is invalid or the
+// Id of the identity can't be determined.
+type Identity struct {
+	Id       string
+	Key      crypto.PublicKey
```
lumjjb, line 40: maybe add a comment here that this should only be used if `Verified` is true?

pkg/ingestor/verifier/verifier.go
```diff
@@ -0,0 +1,42 @@
...
+// verifiers that will check envelopes, signatures, and other payloads
+// for linking a payload back to an identity. In most cases this will be
+// done through verifying one or more signatures against known public keys.
+type Verifier interface {
+	// Verify takes in bytes and returns a list of identities. Those
+	// identities are marked as verified if the payload can be tied back to
+	// an identity, usually through signature validation. In cases where
+	// signatures can't be verified, it is still valuable to return data
+	// about the unverified identity.
+	Verify(payloadBytes []byte) ([]Identity, error)
```
lumjjb, line 32: i think it looks good - i'm guessing all signature types would have a header to describe how to parse them right? or do we need to provide hints?

pkg/ingestor/verifier/verifier.go
```diff
@@ -0,0 +1,42 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verifier
+
+import (
+	"crypto"
+)
+
+// The purpose of Verifier is to allow for multiple signature or identity
```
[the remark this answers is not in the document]
mlieberman85: Yep, just learned about that lint. Will fix.

pkg/ingestor/verifier/verifier.go
mlieberman85: What do you mean?

pkg/ingestor/verifier/verifier.go
mlieberman85: Yeah, the way I see this getting used is, the ingestor would recognize that it's a DSSE, COSE, or another signature/envelope type and know it needs to verify it.

I also could see a future where there might be other ways of verifying identities as well that aren't purely through signature verification. e.g. verify that a specific key came from a specific root.

lumjjb: ah ok - so would there be one Verifier for each type?

mlieberman85: I would think so.

lumjjb: ok yea thanks for clarifying! lgtm

lumjjb (approved):
lgtm

PR #63, opened 2022-08-31
fix: dsse processor document type

cpendery:
Fix the dsse processor to set the format type to be JSON. At this point, the input has been successfully parsed as json


lumjjb:
hmm, i don't think there's a requirement that the DSSE envelope contain a JSON? or i may be wrong @SantiagoTorres @mlieberman85 @pxp928

mlieberman85:
Dsse could contain any blob. In most cases it's an ite6 in json but it doesn't have to be.

lumjjb:
Closing this PR since it is not needed.

PR #65, opened 2022-09-01
Ingestor Parser 

### LINKED ISSUE #12: task: [ingestor] create SLSA document ingestor

pxp928:
Ingestor that takes in DocumentTree and outputs GuacNodes and GuacEdges (AssemblerInput)

Closes #12


pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,261 @@
...
+
+	assemblerinput := assembler.AssemblerInput{
+		Nodes: builder.createNodes(),
+		Edges: builder.createEdges(),
+	}
+
+	return assemblerinput, nil
+}
+
+func (b *graphBuilder) createEdges() []assembler.GuacEdge {
```
lumjjb, line 71: hmm, it looks like this is creating a complete subgraph between types of nodes. Is this intended or just a step towards what we want. I think if this is done locally per node, that would work in creating edges recursively, but it seems here that it's parsing the entire tree and creating edges between each node found within the entire tree?

Not sure if i'm missing something, maybe you can help clarify

pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,261 @@
...
+	}
+	b.foundBuilders = append(b.foundBuilders, build...)
+
+	return nil
+}
+
+func getSubject(doc *processor.Document) ([]assembler.ArtifactNode, error) {
+	foundSubject := []assembler.ArtifactNode{}
+	switch doc.Type {
+	case processor.DocumentITE6SLSA:
```
lumjjb, line 169: in the future probably this can be a plugin model as well, for now, would it be possible to put the processing in a function in a `parser_slsa.go` maybe?

pkg/ingestor/parser/parser.go
pxp928: fixed to create assembler input per document node

pkg/ingestor/parser/parser.go
pxp928: I will push out another PR for this change. Create an interface and breakout the parser.

pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// ParseDocumentTree takes the DocumentTree and create graph inputs (nodes and edges) per document node
+func ParseDocumentTree(docTree processor.DocumentTree) ([]assembler.AssemblerInput, error) {
+
+	assemblerinputs := []assembler.AssemblerInput{}
+	docTreeBuilder := newDocTreeBuilder()
+	err := docTreeBuilder.parse(docTree)
+	if err != nil {
+		return []assembler.AssemblerInput{}, err
```
lumjjb, line 74: nit: return nil, err

pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,286 @@
...
+
+func (b *graphBuilder) createEdges(foundIdentities []assembler.IdentityNode) []assembler.GuacEdge {
+	edges := []assembler.GuacEdge{}
+	for _, i := range foundIdentities {
+		for _, a := range b.foundAttestations {
+			edges = append(edges, assembler.IdentityForEdge{IdentityNode: i, AttestationNode: a})
+		}
+	}
+	for _, s := range b.foundSubjectArtifacts {
+		for _, b := range b.foundBuilders {
```
lumjjb, line 100: nit: use different variable so that it's not confusing to have two versions of `b`

pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,286 @@
...
+		for _, d := range b.foundDependencies {
+			edges = append(edges, assembler.DependsOnEdge{ArtifactNode: s, Dependency: d})
+		}
+	}
+	return edges
+}
+
+func (b *graphBuilder) createNodes() []assembler.GuacNode {
+	nodes := []assembler.GuacNode{}
+	for _, i := range b.foundIdentities {
```
lumjjb, line 115: can do `nodes = append(nodes, b.foundIdentities...)`, etc.

pkg/ingestor/parser/parser.go
```diff
@@ -0,0 +1,286 @@
...
+	for _, c := range root.Children {
+		err := t.parse(c)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (b *graphBuilder) parserHelper(doc *processor.Document) error {
```
lumjjb, line 156: i think it may make sense to do the document switch statement here and then call the get functions within that ? trying to imagine what it would like as a pluggable interface. And move the SLSA specific stuff to `parser_slsa.go`

lumjjb (commented):
mostly LGTM, just minor nit/refactor comments

pkg/ingestor/parser/parser.go
pxp928: this throws an error hence why I did it this way.
```
cannot use b.foundIdentities (variable of type []assembler.IdentityNode) as []assembler.GuacNode value in argument to append
```
unless I am stupid and there is an easy fix to this? Tried to convert to GuacNode Type before append and it does not like that.

pkg/ingestor/parser/parser.go
pxp928: broke the parser out. Might need to discuss on the predicate dictionary implementation.

pkg/ingestor/parser/parser.go
lumjjb: can you try `append(nodes, []assembler.GuacNode(b.foundIdentities...))`? does that work?

lumjjb: yea no, doesn't work nvm...

lumjjb (approved):
lgtm

mlieberman85 (approved):
LGTM

mihaimaruseac (approved):

PR #67, opened 2022-09-01
add key wrapper for Identity struct

pxp928:

created a key wrapper to capture specific pub key information that can later be used downstream.


pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,33 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package key
+
+type KeyType string
+
+type KeyScheme string
+
+/* const (
```
mlieberman85, line 22: Is this commented section supposed to be here?

pxp928: I put that as a reference but will be used by the verifier implementation to specify `KeyType` and `KeyScheme` into the `Identity` struct

mlieberman85: Should we push in-toto to make this public? @SantiagoTorres ?

pxp928: I will remove this comment for now.

pxp928:
closing this as it will be captured in other PRs

PR #69, opened 2022-09-02
Ingestor key interface

### LINKED ISSUE #70: task: [ingestor] create Key interface for ingestor
Create an Key interface that will be used to implement various key providers

pxp928:

Adds key interface to retrieve, store and delete keys from the various key providers.

Closes #70


pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,115 @@
+//
```
mlieberman85, line 1: Do we want to store any hints about the key types at all, or would the expectation be that it is always just a collection of bytes and we have to figure out what key type it is?

pxp928:
Moved key wrapper into the Key package such that it does not need to be done at the verifier package.

internal/testing/ingestor/keyutil/keyutil.go
```diff
@@ -0,0 +1,60 @@
...
+	"crypto/ed25519"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/rsa"
+	"testing"
+
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+)
+
+func GetECDSAPubKey(t *testing.T) crypto.PublicKey {
```
lumjjb, line 30: change this to not take in testing, and return err, and have test that calls this output err?

pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,213 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/sigstore/rekor/pkg/pki/x509"
+	"github.com/sigstore/sigstore/pkg/cryptoutils"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	rsaKeyType            string = "rsa"
```
lumjjb, line 35: create alias type, `type KeyType string`

pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,213 @@
...
+	StoreKey(id string, pk crypto.PublicKey) error
+	// DeleteKey takes in the ID and will remove the associated key from the provider
+	DeleteKey(id string) error
+	// Type returns the key provider type
+	Type() KeyProviderType
+}
+
+type Key struct {
+	// KeyHash sha256 hash of the canonical representation of the key
+	KeyHash string
```
lumjjb, line 61: nit: remove Key prefix for attributes.

pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,213 @@
...
+func RegisterKeyProvider(k KeyProvider, providerType KeyProviderType) {
+	if _, ok := keyProviders[providerType]; ok {
+		logrus.Warnf("the key provider is being overwritten: %s", providerType)
+	}
+	keyProviders[providerType] = k
+}
+
+// Find goes through each of the registered key providers and retrieves the key
+// TODO: Should this handle if multiple keys are returned
+func Find(id string) (*Key, error) {
```
lumjjb, line 86: Find and Retrieve have quite a bit of re-used code, can we refactor for find to use retrieve?

pkg/ingestor/key/key_test.go
```diff
@@ -0,0 +1,406 @@
...
+			err := Store(tt.args.id, tt.args.pk, tt.args.providerType)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("Store() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if err != nil {
+				if !strings.Contains(err.Error(), tt.wantErrMessage) {
+					t.Errorf("Retrieve() error = %s, wantErrMessage %s", err, tt.wantErrMessage)
+				}
+			}
+			if err == nil {
```
lumjjb, line 222: nit: else {

lumjjb (approved):

pkg/ingestor/key/key.go
pxp928: ah yes good catch.

pkg/ingestor/key/key.go
```diff
@@ -0,0 +1,200 @@
...
+
+}
+
+func getKeyInfo(pub crypto.PublicKey) (KeyType, KeyScheme, error) {
+	switch pub.(type) {
+	case *rsa.PublicKey:
+		return rsaKeyType, rsassapsssha256Scheme, nil
+	case *ecdsa.PublicKey:
+		return ecdsaKeyType, ecdsaSha2nistp256, nil
+	case ed25519.PublicKey:
```
mlieberman85, line 195: The others are pointers to the key, this is just the key. Is that right?

mlieberman85 (approved):
LGTM

Issue #73, opened 2022-09-06
task: [ingestor] Create sigstore verifier 


pxp928:
Added a sigstore verifier to validate the signatures based on the public keys

PR #74, opened 2022-09-06
sigstore verifier

pxp928:

Implementation of the verifier interface for sigstore. Currently validates the DSSE envelope. Supports multiple signatures verification.

Closes: #73

### DIFF
--- a/pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
+++ b/pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
@@ -31,11 +31,11 @@ import (
 type SigstoreVerifier struct {
 }
 
+// NewSigstoreVerifier initializes the sigstore verifier
 func NewSigstoreVerifier() *SigstoreVerifier {
 	return &SigstoreVerifier{}
 }
 
-// TODO: Add more than DSSE support here.
 // Verify validates that the signature is valid for the payload
 func (d *SigstoreVerifier) Verify(payloadBytes []byte) ([]verifier.Identity, error) {
 	identities := []verifier.Identity{}
@@ -53,7 +53,7 @@ func (d *SigstoreVerifier) Verify(payloadBytes []byte) ([]verifier.Identity, err
 			ID:  signature.KeyID,
 			Key: *key,
 		}
-		err = verifySignature(key.KeyVal, payloadBytes)
+		err = verifySignature(key.Val, payloadBytes)
 		if err != nil {
 			foundIdentity.Verified = false
 			identities = append(identities, foundIdentity)
@@ -66,6 +66,7 @@ func (d *SigstoreVerifier) Verify(payloadBytes []byte) ([]verifier.Identity, err
 	return identities, nil
 }
 
+// Type returns the type of the verifier
 func (d *SigstoreVerifier) Type() verifier.VerifierType {
 	return "sigstore"
 }
--- a/pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
+++ b/pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
@@ -28,16 +28,17 @@ import (
 	sig_dsse "github.com/sigstore/sigstore/pkg/signature/dsse"
 )
 
-type SigstoreVerifier struct {
+type sigstoreVerifier struct {
 }
 
 // NewSigstoreVerifier initializes the sigstore verifier
-func NewSigstoreVerifier() *SigstoreVerifier {
-	return &SigstoreVerifier{}
+func NewSigstoreVerifier() *sigstoreVerifier {
+	return &sigstoreVerifier{}
 }
 
 // Verify validates that the signature is valid for the payload
-func (d *SigstoreVerifier) Verify(payloadBytes []byte) ([]verifier.Identity, error) {
+// TODO: this currently only supports SHA256 hash function when validating signatures
+func (d *sigstoreVerifier) Verify(payloadBytes []byte) ([]verifier.Identity, error) {
 	identities := []verifier.Identity{}
 	envelope, err := parseDSSE(payloadBytes)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
```diff
@@ -0,0 +1,95 @@
...
+	"fmt"
+
+	"github.com/guacsec/guac/pkg/ingestor/key"
+	"github.com/guacsec/guac/pkg/ingestor/verifier"
+	"github.com/secure-systems-lab/go-securesystemslib/dsse"
+	"github.com/sigstore/sigstore/pkg/signature"
+	sig_dsse "github.com/sigstore/sigstore/pkg/signature/dsse"
+)
+
+type SigstoreVerifier struct {
```
lumjjb, line 31: should be private if exposing `NewSigstoreVerifier()`

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
```diff
@@ -0,0 +1,95 @@
...
+			Key: *key,
+		}
+		err = verifySignature(key.Val, payloadBytes)
+		if err != nil {
+			foundIdentity.Verified = false
+			identities = append(identities, foundIdentity)
+		} else {
+			foundIdentity.Verified = true
+			identities = append(identities, foundIdentity)
+		}
```
lumjjb, line 63: nit:
```
foundIdentity.Verified = (err == nil)
identities = append(identities, foundIdentity)
```

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
lumjjb: add TODO note that this only supports SHA256? (or is that the only one supported?)

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
```diff
@@ -0,0 +1,393 @@
...
+	return &mockKeyProvider{
+		collector: map[string]key.Key{},
+	}
+}
+
+func (m *mockKeyProvider) RetrieveKey(id string) (*key.Key, error) {
+	if key, ok := m.collector[id]; ok {
+		return &key, nil
+	}
+	return nil, nil
```
lumjjb, line 55: hmm , on 2nd thought, maybe the interface should be err with a custom NotFound error to check against... but it's fine for now... just crossed my mind, not an issue since its documented in the interface.

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
```diff
@@ -0,0 +1,393 @@
...
+		t.Fatal("failed to store into mock key provider")
+	}
+	err = key.Store("SHA256:843yiXZzbDfB0gA1snxYG5SISWMnDimw8/8Aew0nVNg", []byte(rsapub), "mock")
+	if err != nil {
+		t.Fatal("failed to store into mock key provider")
+	}
+}
+
+func randomData(t *testing.T, n int) []byte {
+	t.Helper()
```
lumjjb, line 87: this is cool, didn't realize this was a thing.

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
```diff
@@ -0,0 +1,393 @@
...
+	data := make([]byte, n)
+	if _, err := rand.Read(data[:]); err != nil {
+		t.Fatal(err)
+	}
+	return data
+}
+
+func TestSigstoreVerifier_Verify(t *testing.T) {
+	setupOneProvider(t)
+	foundECDSAKey, err := key.Find("SHA256:s9b/UAMASq9HN7RPBm5cIHQGoBOQA120kFdWLW/lT88")
```
lumjjb, line 98: it looks like there's an implicit (non-encapsulating data format) contract between the key provider and the KeyID.... in this case, store should return a key ID so that it can be looked up.

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
```diff
@@ -0,0 +1,393 @@
...
+	}}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sigVerifier := NewSigstoreVerifier()
+			got, err := sigVerifier.Verify(tt.doc.Blob)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("SigstoreVerifier.Verify() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
lumjjb, line 331: err != nil should skip this step

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
pxp928: so that is the default one that supported by in-toto and rekor. I will add in the note.

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier.go
pxp928: simple!

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
pxp928: so the thought process here was that if it is not found by the `RetrieveKey` it would move on to the next key provider when using the `Find`. When `Retrieve` is called, it would return an error: `errors.New("failed to find key from key provider")`

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
pxp928: got some ideas from rekor :)

pxp928:
Update: Changed the key hash function to use upstream `dsse.SHA256KeyID(ecdsaPub)` function from https://github.com/secure-systems-lab/go-securesystemslib/blob/main/dsse/verify.go#L134

pkg/ingestor/verifier/sigstore_verifier/sigstore_verifier_test.go
pxp928: When storing the key, you pass in the keyID. In most cases the KeyID will be the hash of the key but it is not enforced.

PR #78, opened 2022-09-07
added inmemory key provider

### LINKED ISSUE #77: task: [ingestor] create inmemory key provider for ingestor
inmemory key provider is an implementation of the key interface. This will store keys in-memeory

pxp928:

in-memory key provider that implements the key interface.

Closes #77

### DIFF
--- a/internal/testing/ingestor/keyutil/keyutil.go
+++ b/internal/testing/ingestor/keyutil/keyutil.go
@@ -27,28 +27,40 @@ import (
 	"github.com/sigstore/sigstore/pkg/cryptoutils"
 )
 
-func GetECDSAPubKey() (crypto.PublicKey, error) {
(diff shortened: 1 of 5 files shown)


mlieberman85:
I can't remember if we had the conversation about whether or not the keyprovider should hold onto the wrapped key. Did we have a decision on that?

pxp928:
> I can't remember if we had the conversation about whether or not the keyprovider should hold onto the wrapped key. Did we have a decision on that?

Currently we create the wrapped key during the retrieval from the provider. What would be the use case to store them?

pxp928:
Changed to store wrapped Key directly into the key provider.

lumjjb (approved):
lgtm

mlieberman85 (approved):
LGTM

PR #61, opened 2022-08-31
fix: update collect interface

cpendery:
Closes https://github.com/guacsec/guac/issues/32 & fixes a concurrency bug since there was no synchronization on the document collectors before trying to process the results

PR #79, opened 2022-09-11
Add file collector

mlieberman85:


pkg/handler/collector/file/file.go
```diff
@@ -0,0 +1,90 @@
...
+			return err
+		}
+
+		blob, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		doc := &processor.Document{
+			Blob: blob,
```
lumjjb, line 67: need to initialize the DocumentType and FormatType

mlieberman85: Should GCS collector be doing the same thing: https://github.com/guacsec/guac/blob/main/pkg/handler/collector/gcs/gcs.go#L172-L177

pkg/handler/collector/file/file.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+)
+
+const CollectorFile = "File"
+
+type file struct {
```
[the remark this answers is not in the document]
mlieberman85: Yep, I'm not done with this yet. Pushed it out as a draft to help walk @shafeeshafee through this.

pkg/handler/collector/file/file.go
lumjjb: ah yes it should @pxp928 , should be set to the unknown enum, since the unknown enum is not the empty string.

pkg/handler/collector/file/file.go
```diff
@@ -0,0 +1,90 @@
...
+	"io/fs"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+)
+
+const CollectorFile = "File"
```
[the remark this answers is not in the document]
mlieberman85: I will probably make this one configurable

pkg/handler/collector/file/file_test.go
```diff
@@ -0,0 +1,148 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	type fields struct {
+		path        string
+		lastChecked time.Time
+		poll        bool
+		interval    time.Duration
+	}
+	type args struct {
```
pxp928, line 37: remove as not used.

pkg/handler/collector/file/file.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+
+		blob, err := ioutil.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		doc := &processor.Document{
+			Blob:   blob,
+			Type:   processor.DocumentITE6Unknown,
```
pxp928, line 86: this should be `processor.DocumentUnknown`

mlieberman85: thanks

pkg/handler/collector/file/file_test.go
```diff
@@ -0,0 +1,148 @@
...
+					if !reflect.DeepEqual(d, tt.want) {
+						t.Errorf("fileCollector.RetrieveArtifacts() = %v, want %v", d, tt.want)
+					}
+				case err := <-errChan:
+					if (err != nil) != tt.wantErr {
+						t.Errorf("fileCollector.RetrieveArtifacts() = %v, want %v", err, tt.wantErr)
+						return
+					}
+				}
+				collectorsDone += 1
```
lumjjb, line 137: should be in errChan case

pkg/handler/collector/file/file_test.go
```diff
@@ -0,0 +1,148 @@
...
+			}
+			// Drain anything left in document channel
+			for len(docChan) > 0 {
+				<-docChan
```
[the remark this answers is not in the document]
pxp928: hmm I will need to make this fix in the GCS one also.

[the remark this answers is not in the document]
lumjjb: https://github.com/guacsec/guac/pull/61 should help make that easier as well

pkg/handler/collector/file/file_test.go
mlieberman85: Wouldn't this cause timeouts in the docChan case?

mihaimaruseac (approved):

pkg/handler/collector/file/file.go
```diff
@@ -0,0 +1,126 @@
...
+
+		docChannel <- doc
+
+		return nil
+	}
+
+	if f.poll {
+		for {
+			err := filepath.WalkDir(f.path, readFunc)
+			if errors.Is(err, ctx.Err()) {
```
lumjjb, line 103: i think this has to be within the err != nil check. not sure what the behavior is for if err == nil for `errors.Is`, since `ctx.Err()` may be nil, nil may == nil? https://go.dev/play/p/RGIItjVt5V0

I think this is a bug? if the test are passing, may want to add a test that checks for this. Check for poll ending prematurely

pkg/handler/collector/file/file_test.go
```diff
@@ -0,0 +1,133 @@
...
+				errChan <- f.RetrieveArtifacts(ctx, docChan)
+			}()
+
+			numCollectors := 1
+			collectorsDone := 0
+
+			s := []*processor.Document{}
+
+			for collectorsDone < numCollectors {
+				err := <-errChan
```
lumjjb, line 113: you still need to select on the doc channel here if not its possible that the go routine running the collector will block... for example, if you change the buffer in the channel to 0 instead of 1, it will stall indefinitely

lumjjb (approved):
LGTM!

PR #82, opened 2022-09-12
changed to use upstream sslib hash function

pxp928:

Update: Changed the key hash function to use upstream `dsse.SHA256KeyID(ecdsaPub)` function from https://github.com/secure-systems-lab/go-securesystemslib/blob/main/dsse/verify.go#L134


pkg/ingestor/verifier/verifier_test.go
```diff
@@ -101,12 +99,12 @@ func newMockSigstoreVerifier() *mockSigstoreVerifier {
 }
 
 func (m *mockSigstoreVerifier) Verify(payloadBytes []byte) ([]Identity, error) {
-	h := sha256.Sum256(pemBytes)
+	keyHash, _ := dsse.SHA256KeyID(ecdsaPubKey)
```
lumjjb, line 102: probably should check error

lumjjb (approved):
minor nit, else lgtm

pkg/ingestor/verifier/verifier_test.go
pxp928: done.

Issue #55, opened 2022-08-30
Set up bots to benefit from more automation


rgreinho:
Bots can help with various tasks, and it would be useful to setup a few for this project/organization:

- Automerge: Either using the [GitHub built-in feature](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-auto-merge-for-pull-requests-in-your-repository) of [Kodiak](https://kodiakhq.com/)
- [Pull Request size](https://github.com/marketplace/pull-request-size) 
- [Stale Bot](https://github.com/marketplace/stale)

I think we should consider to add at least these 3 bots.

rgreinho:
Regarding the automerge feature, Kodiak has the ability to automatically rebase PRs, which is very nice.

PR #85, opened 2022-09-14
Add bot configurations

rgreinho:
Adds the configurations for the following bots:
- dependabot
- kodiak
- stale

The bots will have to be enabled by an admin in order to get activated.

Fixes guacsec/guac#55

Signed-off-by: Rémy Greinhofer <remy.greinhofer@gmail.com>

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,32 @@
+# Number of days of inactivity before a PR becomes stale.
+daysUntilStale: 60
+# Number of days of inactivity before a stale PR is closed.
+daysUntilClose: 30
+
+# Limit to only `pulls`.
+only: pulls
+
+# Issues or Pull Requests with these labels will never be considered stale.
+# Set to `[]` to disable.
+exemptLabels:
+  - no-mergify
+  - do-not-merge
+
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):
LGTM

rgreinho:
The [PR size]( https://github.com/marketplace/pull-request-size) bot will also need to be activated, but does not have any configuration.

rgreinho:
The [PR size](https://github.com/marketplace/pull-request-size) bot will also need to be activated, but does not have any configuration.

.github/stale.yml
```diff
@@ -0,0 +1,32 @@
+# Number of days of inactivity before a PR becomes stale.
+daysUntilStale: 60
```
lumjjb, line 2: i think we agreed on 90 here. other than that, lgtm

lumjjb:
bots enabled!

.github/stale.yml
rgreinho: And we did :) 

If you look at the line bellow, it says "daysUntilClose: 30", so 60 days for a warning + 30 days before closing, and we're at 90 as discussed.

pxp928 (approved):

PR #86, opened 2022-09-14
Ingest/assembler example command

lumjjb:
Add sample example to create a graph in a neo4j database from SLSA documents.


mihaimaruseac (approved):
LGTM.

Also, nice on `AppendGraph` function

PR #103, opened 2022-09-21
refactor ingestor in preparation for pluggability

lumjjb:

### DIFF
--- a/pkg/ingestor/parser/common.go
+++ b/pkg/ingestor/parser/common.go
@@ -1,0 +1,93 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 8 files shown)


mlieberman85:
Can you elaborate in the PR/commit notes about the pluggability? Based on previous discussions it seems like we want to take a different approach for pluggability here compared to the collectors.

The predicate dictionary sounded like it would be a separate component that would live elsewhere with some schema to help us map from predicate -> nodes/edges?

Do we plan to have something like a `PredicateDictionaryParser` that will call out to an external predicate dictionary for most/all ite6 predicates?

pxp928 (commented):
This looks good. My only comment is should the file be called `common.go`? or would it be similar to like the processor and collector where the SPDX `graphbuilder` will be different compared to the SLSA one.

pxp928:
We should also break them out into different packages to be consistent with the others

lumjjb:
updated!

pxp928 (approved):
LGTM

mihaimaruseac (approved):
LGTM, but I think in general we'd like to have predicate dictionaries build a subset of the graph and as we traverse the document(s) (tree) we keep extending the graph with only what the predicate says at the time (as I'm thinking we might get too many spurious edges if we'll be going with this approach in the long term). But maybe this would be for a later time.

PR #106, opened 2022-09-22
Add test attestations based on TensorFlow

### LINKED ISSUE #98: Extracting some information from bazel and we can make slsa attestations out of it

mihaimaruseac:
These are generated via a script, not real SLSA provenance generators.

Putting them here as I still plan to test the graph serialization with real SLSA nodes.

Should close #98 


lumjjb:
Actually... we now have a repo for this!! https://github.com/guacsec/guac-data

PR #107, opened 2022-09-23
add package node and edges

lumjjb:

### DIFF
--- a/pkg/assembler/nodes.go
+++ b/pkg/assembler/nodes.go
@@ -67,8 +67,7 @@ func (an PackageNode) PropertyNames() []string {
 }
 
 func (an PackageNode) IdentifiablePropertyNames() []string {
-	// An artifact can be uniquely identified by digest
-	return []string{"purl", "digest"}
+	return []string{"purl"}
 }
 
 // IdentityNode is a node that represents an identity
--- a/pkg/assembler/nodes.go
+++ b/pkg/assembler/nodes.go
@@ -41,6 +41,36 @@ func (an ArtifactNode) IdentifiablePropertyNames() []string {
 	return []string{"digest"}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/assembler/nodes.go
```diff
@@ -41,6 +41,36 @@ func (an ArtifactNode) IdentifiablePropertyNames() []string {
 	return []string{"digest"}
 }
 
+// PackageNode is a node that represents an artifact
+type PackageNode struct {
+	Name   string
+	Digest string
```
mlieberman85, line 47: How do we ensure the digest is required?

pkg/assembler/nodes.go
```diff
@@ -41,6 +41,36 @@ func (an ArtifactNode) IdentifiablePropertyNames() []string {
...
+	Purl   string
+	CPEs   []string
```
pxp928, line 49: we cant store lists of strings as an properties in GUAC

pkg/assembler/nodes.go
lumjjb: In some of the SPDX documents some of them don't have digests - and may not be defined for them (e.g. if they are just describing package versions). I think this will be an exercise once we ingest and then bring up in a laundry list to fix.

lumjjb: we had more discussions around multiple and missing identifiers and i think we're going forward with using "purl" for now, since it works with the current dataset of syft SBOMs. This decision isn't quite set in stone and we want to discuss it further with the broader group and when @mihaimaruseac gets back. I've created an issue for now to revisit.

pxp928 (approved):
LGTM. We have an open issue to discuss the questions around unique identifiers and lists as properties

mlieberman85 (approved):
LGTM

PR #109, opened 2022-09-27
add spdx guesser

pxp928:

Add spdx guesser for the document Type


pkg/handler/processor/guesser/type_spdx.go
```diff
@@ -0,0 +1,36 @@
...
+	spdx_json "github.com/spdx/tools-golang/json"
+)
+
+type spdxTypeGuesser struct{}
+
+func (_ *spdxTypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
+	reader := bytes.NewReader(blob)
+	spdxDoc, err := spdx_json.Load2_2(reader)
+	if err == nil && format == processor.FormatJSON {
+		if len(spdxDoc.Packages) > 0 {
```
pxp928, line 31: Do we need any other checks here to make sure its valid?

lumjjb: i don't think this check is required.. also probably if anything this would be the job of the processor and not guesser

pkg/handler/processor/guesser/type_spdx_test.go
```diff
@@ -0,0 +1,65 @@
...
+	//go:embed testdata/small-spdx.json
+	spdxExampleSmall []byte
+
+	//go:embed testdata/alpine-spdx.json
+	spdxExampleBig []byte
+
+	// Invalid types for field spdxVersion
+	//go:embed testdata/invalid-spdx.json
+	spdxInvalidExample []byte
+)
```
lumjjb, line 36: let's maybe move theses into an `internal/` package instead of copying the files again? Then just make these public variables to expose?

pkg/handler/processor/guesser/type_spdx.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+	spdx_json "github.com/spdx/tools-golang/json"
+)
+
+type spdxTypeGuesser struct{}
+
+func (_ *spdxTypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
+	reader := bytes.NewReader(blob)
+	_, err := spdx_json.Load2_2(reader)
```
lumjjb, line 29: probably make a switch statement since we will add other formats in the future.

pkg/handler/processor/guesser/type_spdx.go
```diff
@@ -0,0 +1,34 @@
...
+package guesser
+
+import (
+	"bytes"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
+	spdx_json "github.com/spdx/tools-golang/json"
+)
+
+type spdxTypeGuesser struct{}
```
lumjjb, line 25: register guesser  in https://github.com/guacsec/guac/blob/main/pkg/handler/processor/guesser/type_guesser.go as well!

lumjjb (commented):
minor nits, else lgtm

mlieberman85 (approved):
LGTM

SantiagoTorres:
LGTM, but i assume we want to require json format or spdx? should we document that somewhere? (my understanding is that the kv format is the one used the most in the wild)

lumjjb:
@SantiagoTorres it's currently documented here: https://github.com/guacsec/guac/blob/main/pkg/handler/processor/spdx/spdx.go#L26-L27

lumjjb (approved):
lgtm

Issue #102, opened 2022-09-20
task: [processor/ingestor] CycloneDX support


lumjjb:
Add support to ingest CycloneDX documents

cpendery:
@lumjjb Happy to contribute support for this. Can you assign me?

lumjjb:
awesome @cpendery , i would start with the processor and guesser! https://github.com/guacsec/guac/pull/104

lumjjb:
woop! #140 introduces json cyclone DX support, left needing a processor and i think we can close up this issue for now! @cpendery - if you add the parser portion for XML we should be good to go.

cpendery:
@lumjjb Just finished up the PR for the XML support. Happy to contribute anything last minute as you all prep for KubeCon on Monday

lumjjb:
Awesome. We should be good for kubecon for now! Thanks for helping out!!!

PR #114, opened 2022-09-29
feat: implement xml guesser

cpendery:
Adds an XML guesser required for CycloneDX support in https://github.com/guacsec/guac/issues/102 as CycloneDX has both a JSON and XML format

### DIFF
--- a/pkg/handler/processor/guesser/format_xml.go
+++ b/pkg/handler/processor/guesser/format_xml.go
@@ -1,0 +1,40 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package guesser
+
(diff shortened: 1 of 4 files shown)


pkg/handler/processor/guesser/format_xml.go
```diff
@@ -0,0 +1,40 @@
...
+
+type xmlFormatGuesser struct{}
+
+// GuessFormat expects at least 1 XML element in a blob to identify it as an XML
+// formatted document
+func (_ *xmlFormatGuesser) GuessFormat(blob []byte) processor.FormatType {
+	decoder := xml.NewDecoder(bytes.NewReader(blob))
+	for i := 0; ; i++ {
+		if err := decoder.Decode(new(interface{})); err != nil {
+			if err == io.EOF && i != 0 {
```
lumjjb, line 34: isn't this just trying to parse if the entire document is XML? So it should be just 1 document? Also it seems like [Decode](https://pkg.go.dev/encoding/xml#Decoder.Decode) seems to find the start element, so if there are prefixes to the XML element it would still be valid?

lumjjb (commented):
Thanks for the PR! Added a comment on the validity of XML we are expecting

pkg/handler/processor/guesser/format_xml.go
cpendery: Yeah, it is just trying to decode the entire document as XML. I wasn't sure if the XML spec allowed for siblings at the highest level in level. I can convert it to `Unmarshal` which I don't believe looks for the start element & should fix the prefix problem

lumjjb:
Yea let's do that instead and remove the loop for now.

pxp928 (approved):
LGTM

lumjjb (approved):
lgtm

PR #127, opened 2022-10-06
Add scorecard guesser/processor

lumjjb:
Signed-off-by: Brandon Lum <lumjjb@gmail.com>

PR #130, opened 2022-10-07
Scorecards parser

lumjjb:
Depends on #127


lumjjb:
ready for review!

pkg/assembler/nodes.go
```diff
@@ -314,3 +350,46 @@ func (e ContainsEdge) PropertyNames() []string {
...
+}
+
+func (e MetadataForEdge) Type() string {
+	return "MetadataFor"
+}
+
+func (e MetadataForEdge) Nodes() (v, u GuacNode) {
+	uA, uP := isDefined(e.ForArtifact), isDefined(e.ForPackage)
+	if uA == uP {
+		panic("only on of package and artifact dependency node defined for DependsOn relationship")
```
mlieberman85, line 372: nit: "one" is spelled wrong

pkg/ingestor/parser/scorecard/parser_scorecard.go
```diff
@@ -0,0 +1,136 @@
...
+	mnNode := assembler.MetadataNode{
+		MetadataType: "scorecard",
+		ID:           metadataId(s),
+		Details:      map[string]interface{}{},
+	}
+
+	for _, c := range s.Checks {
+		mnNode.Details[strings.ReplaceAll(c.Name, "-", "_")] = c.Score
+	}
+	mnNode.Details["repo"] = s.Repo.Name
```
pxp928, line 110: would be better to make the keys of the map constants defied in assembler.

lumjjb:
addressed comments! @pxp928 @mlieberman85 ready for review!

pxp928 (approved):
LGTM

pkg/assembler/nodes.go
```diff
@@ -318,3 +354,46 @@ func (e ContainsEdge) PropertyNames() []string {
...
+
+// MetadataFor is an edge that represents the fact that an
+// a metadata node represents metadata for an `ArtifactNode/PackageNode`
+// Only one of each side of the edge should be defined.
+type MetadataForEdge struct {
+	// From node
+	MetadataNode MetadataNode
+	// To node
+	ForArtifact ArtifactNode
+	ForPackage  PackageNode
```
mihaimaruseac, line 366: I'm not sure we can have an edge with 3 nodes. Looking at Nodes() below it looks like we either have an artifact or a package, but never both? Should we document this? (more of a nit)

pkg/ingestor/parser/scorecard/parser_scorecard.go
```diff
@@ -0,0 +1,141 @@
...
+		return fmt.Errorf("expected document type: %v, actual document type: %v", processor.DocumentScorecard, doc.Type)
+	}
+
+	switch doc.Format {
+	case processor.FormatJSON:
+		var scorecard sc.JSONScorecardResultV2
+		if err := json.Unmarshal(doc.Blob, &scorecard); err != nil {
+			return err
+		}
+		//p.scorecards = append(s.scorecards, scorecard)
```
mihaimaruseac, line 58: Nit: I think we can remove the comment

mihaimaruseac (approved):

pkg/assembler/nodes.go
lumjjb: hmm - i added a comment in the struct definition - do you think we need to be more explicit?

mihaimaruseac: I think we should be good

Issue #137, opened 2022-10-07
SLSA parser crashes on multiple subjects OR multiple hashes


lumjjb:
SLSA parser crashes with multiple subjects or multiple hashes

Multiple digests error:
SLSA multiple digests example:
```
  "subject": [
        {
      "name": "gs://kubernetes-release/release/v1.25.2/bin/linux/arm64/kube-apiserver",
      "digest": {
        "sha256": "5522c9bcd76863fa24a658d9faeb6fa2ca999d022806e301e922efca747043f6",
        "sha512": "aa989e60525ac208bc1a7469b486eecb02bf4e7ceb3530c97bae5e0cbc8d4361ce040a8899fa7d9eb56f573fdfc605325e4fcaf956f5efa930cf1a52cb5ebb10"
      }
    }
      ],
```

Error:
```
panic: runtime error: index out of range [342] with length 342

goroutine 1 [running]:
github.com/guacsec/guac/pkg/assembler.StoreGraph({{0xc00017c800, 0x156, 0x180}, {0xc000372000, 0x3f9, 0x400}}, {0x1d75098?, 0xc0000c6f20?})
	/Users/lumb/go/src/github.com/guacsec/guac/pkg/assembler/graphdb.go:62 +0xa1d
github.com/guacsec/guac/cmd/guacone/cmd.getAssembler.func1({0xc0005a7a70?, 0x1?, 0xc00019e540?})
	/Users/lumb/go/src/github.com/guacsec/guac/cmd/guacone/cmd/files.go:178 +0xc5
github.com/guacsec/guac/cmd/guacone/cmd.glob..func1.1(0xc00019f020)
	/Users/lumb/go/src/github.com/guacsec/guac/cmd/guacone/cmd/files.go:109 +0x13b
github.com/guacsec/guac/pkg/handler/collector.Collect({0x1d73be8?, 0xc0008001e0}, 0xc00035dcc8, 0xc00035dc58)
	/Users/lumb/go/src/github.com/guacsec/guac/pkg/handler/collector/collector.go:84 +0x2f0
github.com/guacsec/guac/cmd/guacone/cmd.glob..func1(0x2488a80?, {0xc000800180, 0x1, 0x3})
	/Users/lumb/go/src/github.com/guacsec/guac/cmd/guacone/cmd/files.go:125 +0x56a
github.com/spf13/cobra.(*Command).execute(0x2488a80, {0xc000800120, 0x3, 0x3})
[...]

PR #139, opened 2022-10-07
Fix preallocated map size

mlieberman85:
edge_dicts was incorrectly set to be sized to the number of the nodes. This means that if a document contains less nodes than edges it will crash.


lumjjb (approved):
LGTM

lumjjb:
Fixes https://github.com/guacsec/guac/issues/137

pxp928 (approved):
LGTM

PR #143, opened 2022-10-11
add collector source information

pxp928:

Adding collector and source information to the nodes.

### DIFF
--- a/pkg/assembler/helper.go
+++ b/pkg/assembler/helper.go
@@ -18,7 +18,34 @@ package assembler
 import (
 	"reflect"
 	"strings"
+
+	"github.com/guacsec/guac/pkg/handler/processor"
 )
+
+// objectMetadata appends metadata associated with the node
+type objectMetadata struct {
+	// sourceInfo is the file location from which the node was created
+	sourceInfo string
+	// collectorInfo is the collector from which the file that created the node came from
+	collectorInfo string
+}
+
+// NewObjectMetadata creates a new instance to add metadata to nodes
+func NewObjectMetadata(s processor.SourceInformation) *objectMetadata {
+	return &objectMetadata{
+		sourceInfo:    s.Source,
+		collectorInfo: s.Collector,
+	}
+}
+
+func (o *objectMetadata) addProperties(prop map[string]interface{}) {
+	prop["source"] = o.sourceInfo
+	prop["collector"] = o.sourceInfo
+}
+
+func (o *objectMetadata) getProperties() []string {
+	return []string{"source", "collector"}
+}
 
 func isDefined(v interface{}) bool {
 	return !reflect.ValueOf(v).IsZero()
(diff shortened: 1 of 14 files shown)


pxp928:
Changed to metadata struct so that we can add other properties like temporal data in future PRs.

pkg/assembler/nodes.go
```diff
@@ -29,11 +34,12 @@ func (an ArtifactNode) Properties() map[string]interface{} {
 	properties := make(map[string]interface{})
 	properties["name"] = an.Name
 	properties["digest"] = an.Digest
+	properties["source"] = an.Metadata.Source
```
lumjjb, line 37: i think we can make a generic function for this? like populate metadata ? And do the same for PropertyNames() and others as well? instead of manually populating them?

pkg/assembler/nodes.go
```diff
@@ -15,10 +15,15 @@
 
 package assembler
 
+type Metadata struct {
+	Source string
```
lumjjb, line 19: Should this mirror source information with both collector and source ? or ?

lumjjb: let's add a comment for the Metadata and its fields since its not super straightforward what they mean.

pxp928: Added both source and collector.

pkg/assembler/nodes.go
pxp928: made it a map[string]interface{} so can add properties in the future. Added constants for the property names.

lumjjb (approved):
I was thinking more of a defined type rather than a `map[string]interface{}`, but this works for now.

pxp928:
> I was thinking more of a defined type rather than a `map[string]interface{}`, but this works for now.

What did you have in mind for the defined Type. I was thinking this would allow us to add in more values in the future without much additional changes to the nodes.go

lumjjb:
i was thinking like defining

```
type objectMetadata struct{
    SourceInfo string 
    CollectorInfo string
}

func (m *objectMetadata) addProperties (m map[string]interface);
func (m *objectMetadata) getProperties () []string;

// When temporal data is added, it will automatically be set with a call to NewObjectMetadata
func NewObjectMetadata (s processor.SourceInformation) objectMetadata 
```
yea it's minor so it's kind of whatever.. it's a small detail

pxp928:
ah cool I can do that.

pkg/assembler/helper.go
```diff
@@ -18,8 +18,35 @@ package assembler
...
+// objectMetadata appends metadata associated with the node
+type objectMetadata struct {
+	// sourceInfo is the file location from which the node was created
+	sourceInfo string
+	// collectorInfo is the collector from which the file that created the node came from
+	collectorInfo string
+}
+
+// NewObjectMetadata creates a new instance to add metadata to nodes
+func NewObjectMetadata(s processor.SourceInformation) *objectMetadata {
```
lumjjb, line 34: do you think it'd be easier if we just ask for the document itself? do you think that'd be an easier interface? if we do end up storing the information as well, the storage information can be embedded in the document struct as well - thoughts?

mihaimaruseac (approved):

PR #146, opened 2022-10-12
minor spdx bug fixes for heuristics

lumjjb:


pxp928 (approved):
LGTM but I think we should move the `strings.ToLower` to the assembler level. Can be another PR.

mlieberman85 (approved):
LGTM

PR #147, opened 2022-10-12
fix hash algo being uppercase

### LINKED ISSUE #135: Keep consistency in certain identifying properties in GuacNode/Edges
In certain cases, there may be slight variation in identifiers, for example, digest being "sha256:abc..." vs "SHA256:abc...", this should be handled so that common identified nodes are not duplicated.

this can probably be done easily with the `Properties` function (https://github.com/guacsec/guac/blob/main/pkg/assembler/nodes.go#L56)

pxp928:

Fixes #135


pkg/assembler/helper.go
```diff
@@ -15,8 +15,19 @@
...
+import (
+	"reflect"
+	"strings"
+)
 
 func isDefined(v interface{}) bool {
 	return !reflect.ValueOf(v).IsZero()
 }
+
+func toLower(v []string) []string {
```
lumjjb, line 27: how about using varargs for this? will that work well?

pxp928: changed to variadic function but it would still need to return a `[]string`

lumjjb (approved):
LGTM

PR #148, opened 2022-10-12
fix CI unit test issues from PR 146

pxp928:


mlieberman85:
We should merge to get the tests passing, but it might also make sense at some point to have a test to ensure the tolower is happening as expected.

mlieberman85 (approved):
LGTM

pxp928:
I can make a test case in https://github.com/guacsec/guac/pull/147 as we will be pulling the `tolower` to the assembler level

lumjjb (approved):

PR #149, opened 2022-10-12
ingestor plugin registered as func instead of static struct

lumjjb:
Fixes a bug where the ingestor plugin system reuses structs and thus isn't reset between each invocation of the plugin.


pxp928 (approved):
LGTM

mihaimaruseac (approved):

PR #154, opened 2022-10-14
update assembler and processor for CREV and Certify attestation

pxp928:

Update assembler and processor for code review and github review. Another PR will cover the changes to the ingestor

### DIFF
--- a/internal/testing/processor/testdata/github-review.json
+++ b/internal/testing/processor/testdata/github-review.json
@@ -6,7 +6,7 @@
         "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
       }
   ],
-  "predicateType": "http://in-toto.io/attestation/certify",
+  "predicateType": "https://in-toto.io/attestation/certify/v0.1",
   "predicate": {
     "certifier": {
       "name": "John Doe",
--- a/internal/testing/processor/testdata/github-review.json
+++ b/internal/testing/processor/testdata/github-review.json
@@ -6,8 +6,15 @@
         "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
       }
   ],
-  "predicateType": "http://in-toto.io/attestation/review",
+  "predicateType": "http://in-toto.io/attestation/certify",
   "predicate": {
+    "certifier": {
+      "name": "John Doe",
+      "sig": "",
+      "pubKey": "",
+      "url": "person@example.com"
+    },
+    "date": "2022-10-03T12:00:00Z",
     "full_review": "https://github.com/kubernetes/kubernetes/pull/112078#pullrequestreview-1088153270"
   }
 }
--- a/internal/testing/processor/testdata/github-review.json
+++ b/internal/testing/processor/testdata/github-review.json
@@ -1,0 +1,13 @@
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [
+      {
+        "name": "git://github.com/kubernetes/kubernetes", 
+        "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
+      }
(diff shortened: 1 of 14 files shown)


pkg/handler/processor/processor.go
```diff
@@ -53,6 +53,8 @@ type DocumentType string
 // Document* is the enumerables of DocumentType
 const (
 	DocumentITE6SLSA    DocumentType = "SLSA"
+	DocumentITE6CREV    DocumentType = "CREV"
+	DocumentITE6REVIEW  DocumentType = "REVIEW"
```
lumjjb, line 57: I think we can skip these , let `DocumentITE6Unknown` is good, maybe we rename it to `DocumentITE6Generic` instead

internal/testing/processor/testdata/github-review.json
```diff
@@ -0,0 +1,13 @@
...
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [
+      {
+        "name": "git://github.com/kubernetes/kubernetes", 
+        "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
+      }
+  ],
+  "predicateType": "http://in-toto.io/attestation/review",
+  "predicate": {
+    "full_review": "https://github.com/kubernetes/kubernetes/pull/112078#pullrequestreview-1088153270"
```
lumjjb, line 11: i'm not quite sure what the `full_review` here is trying to say ... Maybe we can change this to just a certification attestation "http://in-toto.io/attestation/certify`, and have things like who the certifier is, etc.

lumjjb (commented):
Code LGTM, minor changes.

pkg/handler/processor/processor.go
pxp928: If we do this. How will the ingestor know how to parse? As we want to capture the payload associated with the predicate.

internal/testing/processor/testdata/github-review.json
pxp928: Changed this to be `Certify` and added additional fields. Good to get feedback from @mlieberman85 @SantiagoTorres

pkg/handler/processor/processor.go
lumjjb: i think it would flatten each into a dictionary - at least for now, until we establish the pred dictionary in the future. The fields can just be all included for now for anything that isn't SLSA, so the trade off would be that for now, it only points to the subject.

internal/testing/processor/testdata/github-review.json
mlieberman85: Does it make sense to just make it clear this is just like a test certification type or something? Or like "alpha" attestation or however we want to structure it. I just want to make sure it's clear that this isn't a widely accepted predicate yet.

mihaimaruseac (approved):
LGTM modulo the pending discussion

nealmcb:
I assume that is intended to be "Certify" in this title. ("Update assembler and processor for CREV and Certify attestation") ;)

internal/testing/processor/testdata/github-review.json
```diff
@@ -0,0 +1,20 @@
+{
+  "_type": "https://in-toto.io/Statement/v0.1",
+  "subject": [
+      {
+        "name": "git://github.com/kubernetes/kubernetes", 
+        "digest": {"sha1": "5835544ca568b757a8ecae5c153f317e5736700e"}
+      }
+  ],
+  "predicateType": "http://in-toto.io/attestation/certify",
```
mlieberman85, line 9: One more minor change here, similar to `https://slsa.dev/provenance/v0.2` can we make this something like `https://in-toto.io/attestation/certify/v0.1`?

mlieberman85 (approved):
LGTM

Issue #133, opened 2022-10-07
Add guacone command to add indexes


lumjjb:
Provide a guacone subcommand (or include as part of the files) code to create the indices to help performance.

mihaimaruseac:
Closed by #175

PR #175, opened 2022-10-20
Create indices in guacone

mihaimaruseac:
Solves #133. I'll try to find a solution where the indices are created automatically so we don't need to keep track of this functionality when adding new node types.

### DIFF
--- a/cmd/guacone/cmd/files.go
+++ b/cmd/guacone/cmd/files.go
@@ -175,6 +175,12 @@ func getAssembler(opts options) (func([]assembler.Graph) error, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	err = createIndices(client)
+	if err != nil {
+		return nil, err
+	}
+
 	return func(gs []assembler.Graph) error {
 		combined := assembler.Graph{
 			Nodes: []assembler.GuacNode{},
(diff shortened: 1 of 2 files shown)


cmd/guacone/cmd/files.go
```diff
@@ -190,3 +196,23 @@ func getAssembler(opts options) (func([]assembler.Graph) error, error) {
...
 	}, nil
 }
+
+func createIndices(client graphdb.Client) error {
+	indices := map[string][]string{
+		"Artifact":    {"digest", "name"},
+		"Package":     {"purl", "name"},
+		"Metadata":    {"id"},
+		"Attestation": {"digest"},
+	}
```
lumjjb, line 206: nit: thinking if this should be part of the assmbler

lumjjb (approved):
lgtm - minor comment on placement of index definition

cmd/guacone/cmd/files.go
mihaimaruseac: Most likely will move there in the future. I'm currently looking at usages of `GuacNode` and `GuacEdge` throughout the codebase to try to see which places cause duplication and don't have good UX (e.g., there's a very large duplication in `internal/testing/ingestor/testdata`, as a very stringent example, albeit in testing). I'm planning some refactoring on these after the POC.

mlieberman85 (approved):

PR #176, opened 2022-10-21
Consume results of write-only transactions

fbiville:
While it's not strictly necessary, since the Neo4j server will execute the transaction queries just before commit at latest, consuming results makes sure the queries are run right when the client starts pulling results.

Another advantage is that some query errors will surface when you consume results and not when you commit.


lumjjb (approved):
LGTM! Thanks @fbiville !

mihaimaruseac (approved):
This LGTMs to me, thank you very much

lumjjb:
hi @fbiville noticed that there was an issue with the sign off message (DCO), with your permission can we add the DCO sign off line on your behalf?

fbiville:
@lumjjb I somehow completely missed your message, apologies. No problem to add the sign off as you've probably already done by now

PR #185, opened 2022-10-24
Remove the additional step of creating the indexes

mihaimaruseac:
Now the index is created by `guacone` so it should be handled by the next step already.


lumjjb (approved):
LGTM

mlieberman85 (approved):
LGTM

Issue #206, opened 2022-11-09
Dependencies in CycloneDX SBOM format not parsed


albert0815:
While trying guac with a CycloneDX SBOM I found that no edges are being created. To make sure this is a reproducable case you could run the following commands to create the same SBOM:

```
git clone https://github.com/quarkusio/quarkus-quickstarts.git
cd quarkus-quickstarts
mvn org.cyclonedx:cyclonedx-maven-plugin:2.7.1:makeBom
.../bin/guacone files --creds neo4j:s3cr3t target/bom.json
```

It seems no edges are identified in that BOM and therefore no dependency graph is built in neo4j, only single nodes are created. But as far as I understand the SBOM actually contains the dependency information which could be used to build the graph. In the SBOM it looks like this:

```
{
  components: [...],
  dependencies: [
    {
      ref: getting-started
      dependsOn: [resteasy-reactive]
    },
    {
      ref: resteasy-reactive
      dependsOn: [resteasy-common, ...]
    }
...
  ]
}
```

So based on above sample I do not think it would be required to use information of the language ecosystem or dep.dev. The details are in the SBOM but not being parsed.

Posted this initially in #200 but created a new issue as suggested by @lumjjb

lumjjb:
hi @albert0815 can you share your bom file so we can add it to the tests?

lumjjb:
running into errors trying to generate this @albert0815 , could you share the copy you have - while we try to fix the `mvn` errors

albert0815:
here you are: [bom.zip](https://github.com/guacsec/guac/files/9974784/bom.zip)

lumjjb:
awesome thanks!  @nadgowdas is working on a fix for this!

albert0815:
Cool, thank you. I was working on a system to manage CycloneDX BOMS and stumbled across the GUAC announcement. I thought it would be worth a try to replace my own storing routine with GUAC. I will take another look when this bug is solved.

Created another [BOM for testing, for GUAC](https://github.com/guacsec/guac/files/9978951/guac-bom.zip), using this command:

```
cyclonedx-gomod mod -json .>guac-bom.json
```

Maybe it helps :)

nadgowdas:
Created a PR for fix: https://github.com/guacsec/guac/pull/224

lumjjb:
Hi @albert0815 does the PR fix this for you? if it does, we can close this :)

pxp928:
[...]

PR #224, opened 2022-11-14
Bug#206: Parse and include dependencies of CDX format

nadgowdas:
Fixes Bug #206 

If CDX Document has dependencies section, it now parses those and adds to the graph accordingly.

Sample graph for bom from : https://github.com/guacsec/guac/issues/206#issuecomment-1309327095
Screenshot (4)

While running guac for this bom, it took about 35 seconds, which I found little slow. But, most of this time was spent in DB transactions https://github.com/guacsec/guac/blob/main/pkg/assembler/graphdb.go#L77-L89

### DIFF
--- a/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
+++ b/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
@@ -47,15 +47,11 @@ func NewCycloneDXParser() common.DocumentParser {
 
 func (c *cyclonedxParser) CreateNodes(ctx context.Context) []assembler.GuacNode {
 	nodes := []assembler.GuacNode{}
-	addNodes(c.rootComponent, &nodes)
-	return nodes
-}
-
-func addNodes(curPkg component, nodes *[]assembler.GuacNode) {
-	*nodes = append(*nodes, curPkg.curPackage)
-	for _, d := range curPkg.depPackages {
-		addNodes(*d, nodes)
+	nodes = append(nodes, c.rootComponent.curPackage)
+	for _, p := range c.rootComponent.depPackages {
+		nodes = append(nodes, p.curPackage)
 	}
+	return nodes
 }
 
 func addEdges(curPkg component, edges *[]assembler.GuacEdge) {
@@ -99,6 +95,7 @@ func (c *cyclonedxParser) addRootPackage(cdxBom *cdx.BOM) {
 	if cdxBom.Metadata.Component != nil {
 		rootPackage := assembler.PackageNode{}
 		rootPackage.Name = cdxBom.Metadata.Component.Name
+		// rootPackage.CPEs = nil
 		rootPackage.NodeData = *assembler.NewObjectMetadata(c.doc.SourceInformation)
 		if cdxBom.Metadata.Component.PackageURL != "" {
 			rootPackage.Purl = cdxBom.Metadata.Component.PackageURL
@@ -127,8 +124,10 @@ func (c *cyclonedxParser) addPackages(cdxBom *cdx.BOM) {
 			// Digest: []string{comp.Version},
 			Purl:     comp.PackageURL,
 			Version:  comp.Version,
-			CPEs:     []string{comp.CPE},
 			NodeData: *assembler.NewObjectMetadata(c.doc.SourceInformation),
+		}
+		if comp.CPE != "" {
+			curPkg.CPEs = []string{comp.CPE}
 		}
 		parentPkg := component{
 			curPackage:  curPkg,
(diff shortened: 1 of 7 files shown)


pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -95,24 +95,31 @@ func (c *cyclonedxParser) CreateEdges(ctx context.Context, foundIdentities []ass
...
 		}
+		c.rootPackage = parentPackages{
+			curPackage:  rootPackage,
+			depPackages: []*parentPackages{},
+		}
 	}
 }
 
 func (c *cyclonedxParser) addPackages(cdxBom *cdx.BOM) {
+	pkgMap := map[string]*parentPackages{}
```
lumjjb, line 122: i think it may be helpful to have this pkgMap be in the parser struct since the references will likely will be useful in other aspects in the future.

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -122,10 +129,24 @@ func (c *cyclonedxParser) addPackages(cdxBom *cdx.BOM) {
 			CPEs:     []string{comp.CPE},
 			NodeData: *assembler.NewObjectMetadata(c.doc.SourceInformation),
 		}
-		c.rootPackage.depPackages = append(c.rootPackage.depPackages, parentPackages{
+		parentPkg := parentPackages{
```
lumjjb, line 132: hmm can we rename this to `package` or `component` node, or something, just parentPackage doesn't seem to match the description anymore?

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -122,10 +129,24 @@ func (c *cyclonedxParser) addPackages(cdxBom *cdx.BOM) {
...
+		c.rootPackage.depPackages = append(c.rootPackage.depPackages, &parentPkg)
+		pkgMap[comp.BOMRef] = &parentPkg
+	}
+
+	if cdxBom.Dependencies != nil {
+		for _, deps := range *cdxBom.Dependencies {
+			if topPkg, found := pkgMap[deps.Ref]; found {
+				for _, depPkg := range *deps.Dependencies {
+					if depPkg, exist := pkgMap[depPkg]; exist {
+						topPkg.depPackages = append(topPkg.depPackages, depPkg)
```
lumjjb, line 145: yea it's a little difficult to reason about thinking about it as a "top" package, probably needs some renaming..  This loop is pretty nested as well, can refactor this or make some of these continues so it doesn't next as much?

lumjjb (commented):
minor comments around renaming since the purpose of the variables are changed now.

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -122,10 +129,24 @@ func (c *cyclonedxParser) addPackages(cdxBom *cdx.BOM) {
...
 			curPackage:  curPkg,
-			depPackages: []parentPackages{},
-		})
+			depPackages: []*parentPackages{},
+		}
+		c.rootPackage.depPackages = append(c.rootPackage.depPackages, &parentPkg)
+		pkgMap[comp.BOMRef] = &parentPkg
+	}
+
+	if cdxBom.Dependencies != nil {
```
pxp928, line 140: Can we also include a smaller example with dependencies that we can add to the unit test here? The above example is massive and would have tons of nodes to map out.

nadgowdas: done!

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
nadgowdas: done!

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
nadgowdas: done!

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
nadgowdas: done!

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
pxp928: Sorry for not being clear but I think need to add to `parser_cyclonedx_test.go` to use the smaller example to test out this new functionality.

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -28,43 +28,45 @@ import (
 )
 
 type cyclonedxParser struct {
-	doc         *processor.Document
-	rootPackage parentPackages
+	doc           *processor.Document
+	rootComponent component
+	pkgMap        map[string]*component
```
lumjjb, line 33: I think as to @pxp928 's comment we would want a parser test here! That should round up the PR, the rest LGTM.

pxp928 (approved):
LGTM

lumjjb (approved):
LGTM

PR #229, opened 2022-11-15
remove all logrus calls. Add ctx

### LINKED ISSUE #232: Fix key_test unit test
During cleanup to remove `logurs` from GUAC (as it is not being used) and pass `context` to allow for the logger to be initialized later as needed, the unit test in `key_test.go` failed. 

```
--- FAIL: TestRetrieve (0.22s)
    key_test.go:390: the key provider is being overwritten: mock1
    key_test.go:394: the key provider is being overwritten: mock2
    --- FAIL: TestRetrieve/one_provider (0.00s)
        key_test.go:169: DSSEProcessor.Unpack() = &{SHA256:DoOlWGZKZUoygFf7dJY0aQixmdMf7JNZKlBS6LQmM8A ecdsa 0x140001bd530 ecdsa-sha2-nistp256}, expected &{SHA256:U1OxUJEVw9d+oxhMBqLuqRdj2MsabgK9kb3EgCuTFCM ecdsa 0x140001bd620 ecdsa-sha2-nistp256}
    --- FAIL: TestRetrieve/two_provider (0.00s)
        key_test.go:169: DSSEProcessor.Unpack() = &{SHA256:5L+wqCcJwRO0lulSRokcLbWV6bWgy1l3Ffy1YFfzthY rsa 0x1400010c7e0 rsassa-pss-sha256}, expected &{SHA256:REWE5MQXyLXMaTon1xo2P0DrCRdxA8Kx+Qr2Q4VVtQ4 rsa 0x1400009c240 rsassa-pss-sha256}
```

The issue is that the testing is done via two key providers and where the keys got replaced before the test completes. Fixed via not re-initializing the mocked key providers if they already exist within the `keyProviders` dictionary.

pxp928:
Remove all traces of Logrus that is not being used. Also, pass context to get logger from context. 

Fixed unit test bug in key_test.go

Edit:

Fixes #232

### DIFF
--- a/pkg/ingestor/key/key_test.go
+++ b/pkg/ingestor/key/key_test.go
@@ -16,18 +16,21 @@
 package key
 
 import (
+	"context"
 	"crypto"
 	"reflect"
 	"strings"
 	"testing"
 
 	"github.com/guacsec/guac/internal/testing/ingestor/keyutil"
+	"github.com/guacsec/guac/pkg/logging"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 )
 
 func TestFind_OneProvider(t *testing.T) {
+	ctx := logging.WithLogger(context.Background())
 	provider, _, wantKey := setupOneProvider(t)
-	provider.collector["ecdsa"] = *wantKey
+	provider.collector["ecdsa"] = wantKey
 	tests := []struct {
 		name    string
 		id      string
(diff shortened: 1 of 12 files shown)


pkg/ingestor/key/key_test.go
```diff
@@ -297,28 +310,28 @@ func TestDelete(t *testing.T) {
 }
 
 type mockKeyProvider struct {
-	collector map[string]Key
+	collector map[string]*Key
```
lumjjb, line 313: hmm quick question on the change here? is this related to ctx change or just a drive by?

pxp928: So as I was adding in the context, the unit test started to fail. Some keys were being overwritten before the test could complete. That was the reason for the changes. Added issue #232 to capture this.

lumjjb (approved):
LGTM

pkg/ingestor/key/key_test.go
lumjjb: awesome - tks!

Issue #222, opened 2022-11-10
OSV Certifier for GUAC


lumjjb:
Design and develop an OSV certifier that will be able to create attestations on artifacts in GUAC such that they are usable for vulnerability queries

lumjjb:
Completed with #245

PR #230, opened 2022-11-15
add VulnerabilityNode and associated edges for certifier

pxp928:

Add VulnerabilityNode, AttestationForPackage, and VulnerableEdge to be used by the certifier. 

Part of issue #222

### DIFF
--- a/pkg/assembler/nodes.go
+++ b/pkg/assembler/nodes.go
@@ -229,6 +229,34 @@ func (mn MetadataNode) IdentifiablePropertyNames() []string {
 	return []string{"metadata_type", "id"}
 }
 
+// VulnerabilityNode is a node that represents a vulnerability associated with the certifier attestation
+type VulnerabilityNode struct {
+	ID       string
+	NodeData objectMetadata
+}
+
+func (vn VulnerabilityNode) Type() string {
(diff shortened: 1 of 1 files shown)


mihaimaruseac (approved):

mihaimaruseac (approved):

pkg/assembler/nodes.go
```diff
@@ -431,3 +459,58 @@ func (e MetadataForEdge) PropertyNames() []string {
 func (e MetadataForEdge) IdentifiablePropertyNames() []string {
 	return []string{}
 }
+
+// AttestationForPackage is an edge that represents the fact that an
+// `AttestationNode` is an attestation for an `PackageNode`.
+type AttestationForPackage struct {
```
lumjjb, line 465: hmm wondering if we should do this as extending `AttestationFor` to be able to take either package or artifact... kind of like what we did for https://github.com/guacsec/guac/blob/main/pkg/assembler/nodes.go#L404-L413

pxp928: Yea I like that idea. Let me update.

pxp928: done.

PR #231, opened 2022-11-15
Fix package overwite cipher bug

mihaimaruseac:
When running a separate ingestion step that only gives a purl for a package, the previous node in the DB would get overriden with empy fields for all other properties.

This PR attempts a workaround while I'm working on a better solution that automates all this boilerplate.

### DIFF
--- a/pkg/assembler/nodes.go
+++ b/pkg/assembler/nodes.go
@@ -66,12 +66,24 @@ func (pn PackageNode) Type() string {
 
 func (pn PackageNode) Properties() map[string]interface{} {
 	properties := make(map[string]interface{})
-	properties["name"] = pn.Name
-	properties["purl"] = pn.Purl
-	properties["version"] = pn.Version
-	properties["cpes"] = pn.CPEs
-	properties["digest"] = toLower(pn.Digest...)
-	properties["tags"] = pn.Tags
+	if len(pn.Name) > 0 {
+		properties["name"] = pn.Name
+	}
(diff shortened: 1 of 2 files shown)


lumjjb (approved):
lgtm, would it be possible to add a test to have this in regression?

pkg/assembler/nodes.go
```diff
@@ -66,12 +66,24 @@ func (pn PackageNode) Type() string {
...
+	if len(pn.CPEs) > 0 {
+		properties["cpes"] = pn.CPEs
+	}
+	if len(pn.Digest) > 0 {
+		properties["digest"] = toLower(pn.Digest...)
+	}
+	if len(pn.Tags) > 0 {
+		properties["tags"] = pn.Tags
+	}
 	pn.NodeData.addProperties(properties)
```
pxp928, line 87: testing this out. Looks like the collector and source still get replaced with empty strings

mihaimaruseac: Oh, those come from  `addProperties`. Sending a fix shortly

mihaimaruseac: Can you try now? I seem to have also found another bug, see 224ec1b

mihaimaruseac:
Regarding testing, we don't have yet a function that will retrieve nodes from within the test since current implementation is generic in the returned query. I'll add tests with the query refactor

pxp928 (approved):
LGTM!

PR #245, opened 2022-11-22
add vulnerability certifier

pxp928:
Signed-off-by: pxp928 [parth.psu@gmail.com](mailto:parth.psu@gmail.com)

Adds vulnerability certifier interface that allows for other certifiers to be pluggable. Currently added OSV support.
The certifier asynchronously queries the graph DB recursively for artifacts to scan. It passes these to the certifiers asynchronously to scan for vulnerabilities and generates vulnerability attestations. The generated attestations are passed asynchronously to the emitter to be ingested.

Currently, we are missing integration testing for the certifier as it needs to query a graph DB for artifacts.

Added internal OSV_query package until the upstream https://github.com/google/osv-scanner/blob/main/internal/osv/osv.go becomes public or another method is used.

`ReadQuery` is a modified version of `ReadQueryForTesting` for this use case but this will be replaced once we refactor graph DB functions and API are generated.

Part of Issue https://github.com/guacsec/guac/issues/222


mlieberman85:
So overall I like the approach. My only major concern is it being unclear how the graphdb stuff works with a bunch of this all parallelized. However, I'm not a neo4j expert here.

pkg/certifier/certifier.go
```diff
@@ -0,0 +1,50 @@
...
+// CertfierType describes the type of the certifier
+type CertfierType string
+
+const (
+	CertifierOSV CertfierType = "OSV"
+)
+
+// Component represents the top level package node and its dependencies
+type Component struct {
+	CurPackage  assembler.PackageNode
```
lumjjb, line 48: nit: maybe just package?

pkg/certifier/certify/certify.go
```diff
@@ -0,0 +1,188 @@
...
+
+func getCompHelper(ctx context.Context, client graphdb.Client, parentPurl string) ([]*certifier.Component, error) {
+	dependencies, err := graphdb.ReadQuery(client, "MATCH (p:Package) WHERE p.purl = $rootPurl WITH p MATCH (p)-[:DependsOn]->(p2:Package) return p2",
+		map[string]any{"rootPurl": parentPurl})
+	if err != nil {
+		return nil, err
+	}
+	depPackages := []*certifier.Component{}
+	for _, dep := range dependencies {
+		foundDep := dep.(dbtype.Node)
```
lumjjb, line 175: nit: check cast

pkg/certifier/certify/certify.go
```diff
@@ -0,0 +1,188 @@
...
+		foundDepPack := assembler.PackageNode{}
+		foundDepPack.Purl = foundDep.Props["purl"].(string)
```
lumjjb, line 177: nit: check cast

pkg/certifier/osv/internal/osv_query/osv_query.go
```diff
@@ -0,0 +1,204 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// This package is currently copied from
+// https://github.com/google/osv-scanner/blob/main/internal/osv/osv.go.
+// Currently this is not exposed by the upstream project. Once it is
+// exposed, this will be removed.
+package osv_query
```
lumjjb, line 20: @oliverchang does OSV have a go library that we can use?

pkg/certifier/osv/osv.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (o *osvCertifier) certifyHelper(ctx context.Context, topLevel *certifier.Component, depPackages []*certifier.Component, docChannel chan<- *processor.Document) ([]osv_scanner.Entry, error) {
+	packNodes := []assembler.PackageNode{}
+	totalDepVul := []osv_scanner.Entry{}
+	for _, depPack := range depPackages {
+		if len(depPack.DepPackages) > 0 {
+			depVulns, err := o.certifyHelper(ctx, depPack, depPack.DepPackages, docChannel)
+			if err != nil {
+				return nil, nil
```
lumjjb, line 69: hmm should this return an error here?

pkg/certifier/osv/osv.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+			totalDepVul = append(totalDepVul, depVulns...)
+		} else {
+			packNodes = append(packNodes, depPack.CurPackage)
```
lumjjb, line 73: hmm, would refactoring this to recursively return the entries for a component streamline the function a bit more? I found it a bit confusing to follow.. or unless i'm looking at it a different way, maybe a comment to help out what's going on.

pkg/certifier/osv/osv_test.go
```diff
@@ -0,0 +1,336 @@
...
+		   "metadata":{
+			  "scannedOn":"2022-11-22T13:18:31.607996-05:00"
+		   }
+		}
+	 }`
+)
+
+func TestOSVCertifier_CertifyVulns(t *testing.T) {
+	ctx := logging.WithLogger(context.Background())
+	rootPackage := assembler.PackageNode{}
```
lumjjb, line 194: nit; may be easier to read if use inline struct decalrations

pkg/certifier/osv/osv_test.go
```diff
@@ -0,0 +1,336 @@
...
+						return
+					}
+					certifiersDone += 1
+				}
+			}
+			// Drain anything left in document channel
+			for len(docChan) > 0 {
+				d := <-docChan
+				collectedDocs = append(collectedDocs, d)
+			}
```
lumjjb, line 300: only continue if err != nil?

pkg/certifier/osv/osv_test.go
```diff
@@ -0,0 +1,336 @@
...
+			for i := range collectedDocs {
+				if !docEqual(t, collectedDocs[i], tt.want[i]) {
+					t.Errorf("g.RetrieveArtifacts() = %v, want %v", string(collectedDocs[i].Blob), string(tt.want[i].Blob))
```
lumjjb, line 303: hmm would the datetime not match? might me missing somethign

pkg/certifier/osv/osv_test.go
pxp928: added to return if err :
```
				case err := <-errChan:
					if (err != nil) != tt.wantErr {
						t.Errorf("g.RetrieveArtifacts() error = %v, wantErr %v", err, tt.wantErr)
						return
					}
					if err != nil {
						t.Errorf("collector ended with error: %v", err)
						return
					}
					certifiersDone += 1

				}
```

pkg/certifier/osv/osv_test.go
pxp928: Depending on when the unit test is run, the datatime would be different. So this just changes them to a hardcoded value so that the comparison passes.

pkg/certifier/osv/osv.go
pxp928: Added the following comment:
```
// certifyHelper recursively checks each component for dependencies. If it has dependencies, certifyHelper is re-called until no more dependencies are found.
// The dependency node is appended to the package node array and sent to be queried by OSV
// Once the vulnerabilities are found, an attestation is generated for that package node
// All the vulnerabilities for each dependent package node are collected and the parent package node's
// attestation is generated containing all the vulnerabilities of its dependencies
// these vulnerabilities are passed up until it reaches the root level node which contains an attestation
// with all the aggregate vulnerabilities
```

lumjjb (approved):
lgtm

pkg/certifier/certify/certify.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	documentCertifier[certifierType] = c
+
+	return nil
+}
+
+// Certify queries the graph DB to get the packages to scan. Utilizing the registered certifiers,
+// it scans and generate vulnerability attestation for each package. Aggregating the results to the
+// top/root level package
+func Certify(ctx context.Context, client graphdb.Client, emitter certifier.Emitter, handleErr certifier.ErrHandler) error {
```
lumjjb, line 57: hmm why is there a need for the graph DB client here, i think that should be part of the emitter.

lumjjb (changes requested):

mihaimaruseac (approved):

pkg/certifier/certify/certify.go
mihaimaruseac: It is used in lines 139 (`getCompoents`) and 173(`getCompHelper`) to issue Cipher queries directly before we get to a GraphQL interface

lumjjb: gotcha- we just had a chat about this and were talking about abstracting that into an interface instead so that we can more easily customize the components that we get (e.g.so we can configure a list of PURLs) - so moving those 2 functions behind an interface instead.

pxp928: @lumjjb Updated to add in interface to abstract out getting the components.

lumjjb (approved):
lgtm - create issue to track test creation

pxp928:
created Issue https://github.com/guacsec/guac/issues/254 to track adding tests

pkg/certifier/osv/internal/osv_query/osv_query.go
oliverchang: Sorry for the **very** late reply! 

We recently did some refactoring and now there's a bunch of public package available here: https://github.com/google/osv-scanner/tree/main/pkg

Are these usable here?

pxp928: @oliverchang I looked through the packages but primarily I create a bunch of batchedQueries based on the purls and make the request to OSV endpoint. Is there a way to expose that for use?

PR #246, opened 2022-11-22
Add Certifier cmd to guacone

pxp928:
Add Certifier command to guacone to manually call the certifier and kick off the vulnerability certification process for testing.

This will run through the current graphDB and look for all packages to run the OSV query against. If a vulnerability is found, an attestation along with a vulnerability node is generated.


cmd/guacone/cmd/certifier.go
```diff
@@ -0,0 +1,128 @@
...
+			os.Exit(1)
+		}
+
+		totalNum := 0
+		// Set emit function to go through the entire pipeline
+		emit := func(d *processor.Document) error {
+			totalNum += 1
+			start := time.Now()
+
+			docNode := &processor.DocumentNode{
```
lumjjb, line 71: hmm should this be sent through the `processor.Process` instead? that should output the document tree for you

lumjjb: also, i think you could reuse some of these functions: https://github.com/guacsec/guac/blob/main/cmd/guacone/cmd/files.go#L166

cmd/guacone/cmd/certifier.go
```diff
@@ -0,0 +1,128 @@
...
+			return false
+		}
+
+		authToken := graphdb.CreateAuthTokenWithUsernameAndPassword(opts.user, opts.pass, opts.realm)
+		client, err := graphdb.NewGraphClient(opts.dbAddr, authToken)
+		if err != nil {
+			logger.Errorf("error: %v", err)
+			os.Exit(1)
+		}
+		packageQuery := root_package.NewPackageQuery(client)
```
lumjjb, line 109: would be nice to refactor this into a function.

lumjjb (commented):
minor nits, else LGTM

cmd/guacone/cmd/certifier.go
pxp928: done.

mihaimaruseac (approved):

lumjjb (approved):
lgtm

PR #255, opened 2022-12-05
added NATS with Jet Stream enabled (NATS PR: 1 of 4)

pxp928:

Added NATS Jetstream implementation to collect, process, and ingest the documents. Pull-based consumers are implemented that allow for scaling and allow multiple processors and ingestors to run simultaneously.

This is 1 of 4 PRs related to NATS. This adds in the NATS emitter to allow for connection to NATS and initialization of Jet Stream.


internal/testing/nats/nats.go
```diff
@@ -0,0 +1,84 @@
...
+func getFreePort() (port int, err error) {
+	var a *net.TCPAddr
+	if a, err = net.ResolveTCPAddr("tcp", "localhost:0"); err == nil {
+		var l *net.TCPListener
+		if l, err = net.ListenTCP("tcp", a); err == nil {
+			defer l.Close()
+			return l.Addr().(*net.TCPAddr).Port, nil
+		}
+	}
+	return
```
lumjjb, line 83: nit: know its totally valid, but empty return statements feel kind of weird to me..

internal/testing/nats/nats.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return "", err
+	}
+	s, err := runServerOnPort(port)
+	if err != nil {
+		return "", err
+	}
+	err = s.EnableJetStream(&server.JetStreamConfig{})
+	if err != nil {
+		return "", fmt.Errorf("unexpected error initializing test NATS: %v", err)
```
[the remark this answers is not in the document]
pxp928: added `s.Shutdown()` to shutdown server before returning error

pkg/emitter/nats_emitter.go
```diff
@@ -0,0 +1,139 @@
...
+func NewJetStream(url string, creds string, nKeyFile string) *jetStream {
+	return &jetStream{
+		url:      url,
+		creds:    creds,
+		nKeyFile: nKeyFile,
+	}
+}
+
+// JetStreamInit initializes NATS and enabled Jet Stream to be used for GUAC
+func (j *jetStream) JetStreamInit(ctx context.Context) (context.Context, error) {
```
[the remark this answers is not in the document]
pxp928: returning original context if there is an error.

pkg/emitter/nats_emitter.go
```diff
@@ -0,0 +1,139 @@
...
+	// Use UserCredentials
+	if j.creds != "" {
+		opts = append(opts, nats.UserCredentials(j.creds))
+	}
+
+	// Use Nkey authentication.
+	if j.nKeyFile != "" {
+		opt, err := nats.NkeyOptionFromSeed(j.nKeyFile)
+		if err != nil {
+			logger.Fatal(err)
```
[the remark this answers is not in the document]
pxp928: this should be returning an error as if authentication is used, there is no reason to continue if it fails. Changed to return original context and error.

pkg/emitter/nats_emitter.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		opts = append(opts, opt)
+	}
+
+	// Connect to NATS
+	nc, err := nats.Connect(j.url, opts...)
+	j.nc = nc
```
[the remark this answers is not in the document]
pxp928: moved to the bottom and added `nc.Close()` on err cases.

pkg/emitter/nats_emitter_test.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return fmt.Errorf("failed marshal of document: %w", err)
+	}
+	_, err = js.Publish(SubjectNameDocCollected, docByte)
+	if err != nil {
+		return fmt.Errorf("failed to publish document on stream: %w", err)
+	}
+	logger.Infof("doc published: %+v", d)
+	return nil
+}
```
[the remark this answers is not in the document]
pxp928: working on this now.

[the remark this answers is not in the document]
pxp928: added a test subscribe to check for the published data.

pxp928:
Note: DSSE/SLSA Testdata was moved from ingestor testdata into processor testdata as it was causing an import cycle error in the emitter package.

internal/testing/nats/nats.go
```diff
@@ -0,0 +1,84 @@
...
+	}
+	time.Sleep(time.Second * 5)
```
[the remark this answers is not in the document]
pxp928: removed sleep as it was not needed on further testing.

pxp928:
```
// RecreateStream deletes the current existing stream and recreates it
func (j *jetStream) RecreateStream(ctx context.Context) error {
```
add to delete and recreate the stream to empty out the queue if needed.

pxp928:
rebased and ready for re-review

pkg/emitter/nats_emitter_test.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	defer jetStream.Close()
+	err = testPublish(ctx, &ite6SLSADoc)
+	if err != nil {
+		t.Fatalf("unexpected error on emit: %v", err)
+	}
+
+	var cancel context.CancelFunc
+
+	ctx, cancel = context.WithTimeout(ctx, time.Second)
```
lumjjb, line 107: if back off for subscribe is 1 second , perhaps the timeout should be longer?

pkg/emitter/nats_emitter_test.go
```diff
@@ -0,0 +1,197 @@
...
+	ctx := context.Background()
+	jetStream := NewJetStream(url, "", "")
+	ctx, err = jetStream.JetStreamInit(ctx)
+	if err != nil {
+		t.Fatalf("unexpected error initializing jetstream: %v", err)
+	}
+	err = jetStream.RecreateStream(ctx)
+	if err != nil {
+		t.Fatalf("unexpected error recreating jetstream: %v", err)
+	}
```
lumjjb, line 98: nit: probably better as its own test, since it's weird sticking in the context of the pub/sub test... unless i'm missing something.

pkg/emitter/nats_emitter_test.go
```diff
@@ -0,0 +1,197 @@
...
+		logger.Errorf("processor subscribe failed: %s", err)
+		return err
+	}
+	for {
+		// if the context is canceled we want to break out of the loop
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		msgs, err := sub.Fetch(1)
+		if err != nil {
```
lumjjb, line 166: hmm not that it matters too much, but probably useful in the future, is there an error we can check for?

lumjjb (commented):
last round minor nits!! This is awesome though, can't wait to try this out!

pkg/emitter/nats_emitter_test.go
pxp928: changed to 3 seconds.

pkg/emitter/nats_emitter_test.go
pxp928: This is used in the test to ensure any lingering documents do get added to the test. I did create another separate test just for `RecreateStream` also.

pkg/emitter/nats_emitter_test.go
pxp928: added the following to check for the error:
```
		msgs, err := sub.Fetch(1)
		if err != nil {
			if errors.Is(err, nats.ErrTimeout) {
				logger.Infof("[processor: %s] error consuming, backing off for a second: %v", id, err)
				time.Sleep(1 * time.Second)
				continue
			} else {
				return err
			}
		}
```

pkg/emitter/nats_emitter_test.go
```diff
@@ -0,0 +1,256 @@
...
+	"github.com/guacsec/guac/internal/testing/dochelper"
+	nats_test "github.com/guacsec/guac/internal/testing/nats"
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/guacsec/guac/pkg/logging"
+	"github.com/nats-io/nats.go"
+	uuid "github.com/satori/go.uuid"
+)
+
+var (
+	// Taken from: https://slsa.dev/provenance/v0.1#example
```
mlieberman85, line 35: Probably fine for this test, but the examples under slsa.dev aren't full examples of the SLSA document. For example the digests aren't real digests `{"sha256": "5678..."}`

mlieberman85 (approved):
Minor FYI comment about slsa.dev examples not being valid documents but overall seems reasonable. I do want someone other than me who understands NATS a bit more to take a look though.

pkg/emitter/nats_emitter_test.go
pxp928: Yup at this level of test it does not matter. For the larger integration test a valid doc that can be ingested will be used.

lumjjb (approved):
lgtm

Issue #31, opened 2022-08-25
Design and implement full integration with pub/sub for GUAC flow for collector/processor/assembler


lumjjb:
Collectors that obtain documents need somewhere to emit them to. The processor, which is the next part of the pipeline needs to gather the documents and process them..

There are a couple options naturally:

1. Processor runs as a gRPC server
2. Processor obtains documents from a Pub/Sub queue (e.g. kafka, nats.io, etc.)
3. Processor ingests from STDIN or file
4. Processor and Collector are part of the same process.

This boils down to we collectors and processors want to be run in the architecture. The ingestor will most likely be tied to the assembler. 

Deliberation:
- Will all the collectors be run in a single executable? I.e. the processor will cache duplicate documents so it is beneficial to have an n:m relationship (where n>m) between collectors and executables. If the answer is no, this excludes option 3 and 4.
  - I think it is likely that this answer is no, given the access of collectors to need credentials and not a single account/team would have all credentials
- Options 1 and 2 are similar, with a trade-off between simplicity and scale.

lumjjb:
@trmiller this may be interesting to you

PR #256, opened 2022-12-05
added marshal and unmarshal for guac nodes and edge interfaces (NATS PR: 2 of 4)

pxp928:

Added `UnmarshalJSON` and `MarshalJSON` for GuacNode and GuacEdge interface. This will allow for byte arrays to be generated that can be used by NATS to publish. These will be unmarshalled back to a GuacNode and GuacEdge to store into the graphDB. 

Note: ID was changed to be more descriptive to ensure proper marshal and unmarshal if a node or edge uses a free-form property like `Details   map[string]interface{}` that can take in any string parameter for the key.

Part 2 of 4 for issue https://github.com/guacsec/guac/issues/31

### DIFF
--- a/pkg/assembler/assembler.go
+++ b/pkg/assembler/assembler.go
@@ -15,6 +15,12 @@
 
 package assembler
 
+import (
+	"encoding/json"
+	"errors"
+	"reflect"
+)
+
 type assembler struct{} //nolint: unused
 
 // NOTE: `GuacNode` and `GuacEdge` interfaces are very experimental and might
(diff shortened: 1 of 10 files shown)


pkg/assembler/assembler.go
```diff
@@ -140,3 +146,118 @@ func (g *Graph) AppendGraph(gs ...Graph) {
 
 // AssemblerInput represents the inputs to add to the graph
 type AssemblerInput = Graph
+
+// UnmarshalJSON deserializes back to graph struct with GuacNode and GuacEdge
+func (g *Graph) UnmarshalJSON(b []byte) error {
```
lumjjb, line 151: nice

pkg/assembler/assembler.go
```diff
@@ -140,3 +146,118 @@ func (g *Graph) AppendGraph(gs ...Graph) {
...
+		return err
+	}
+
+	g.Nodes = foundNodes
+	g.Edges = foundEdges
+
+	return nil
+}
+
+func unmarshalNodeType(rawMessagesForGuacNodes []*json.RawMessage, customTypes map[string]reflect.Type) ([]GuacNode, error) {
```
lumjjb, line 206: i'm wondering since we do control both ends here, would it help to define a serialization format struct instead? May make things a bit cleaner. Thoughts?

Maybe something like

```
type GuacNodeSerialization struct {
    Type string `json:"type"`
    MetdataNode MetadataNode `json:"metadataNode"`
    ArtifactNode ArtifactNode `json:"artifactNode"`
...
}
```

or potentially just on the graph level (although I think it probably is helpful to have it on the node/edge level as well.

pkg/assembler/nodes.go
```diff
@@ -100,32 +147,44 @@ func (pn PackageNode) IdentifiablePropertyNames() []string {
...
-	KeyScheme string
-	NodeData  objectMetadata
+	Key       string         `json:"key"`
+	KeyType   string         `json:"key_type"`
+	KeyScheme string         `json:"key_scheme"`
+	NodeData  objectMetadata `json:"nodedata"`
+}
+
+func (in IdentityNode) MarshalJSON() ([]byte, error) {
+	marhsalProperties := make(map[string]interface{})
```
lumjjb, line 160: typo: marshal

pkg/assembler/nodes.go
```diff
@@ -100,32 +147,44 @@ func (pn PackageNode) IdentifiablePropertyNames() []string {
...
+
+	for k, v := range in.Properties() {
+		marhsalProperties[k] = v
+	}
+
+	marhsalProperties["type"] = in.Type()
```
lumjjb, line 166: yea thoughts on serialization format, may save some complexity here.

pkg/assembler/assembler.go
pxp928: hmm not sure I am following completely. Can you elaborate a bit more?

lumjjb: going to try my best to describe it in prose, but may fail terribly ...

The thought is having `func (g Graph) MarshalJSON`  to take the list of Guac nodes and then based on a switch statement on their type, populate a new struct with a list of each type of node.. So the serializable wire format of the `Graph` is different from the one in golang. 

So Marshal would:
- convert Graph to GuacNodeSerialization
- Serialize GuacNodeSerialization struct to JSON

Unmarshal would:
- Unmarshal JSON to GuacNodeSerialization struct
- convert GuacNodeSerialization to Graph

Hope this articulates the idea better.. will be happy to find some time to hop on the call for more clarity

pxp928:
After discussion. Closing this PR as it will soon be superseded by the GraphQL work.

pkg/assembler/assembler.go
mihaimaruseac: I'd say let's not focus on this, since the GraphQL interface will also serialize stuff that goes to the DB

pxp928: @mihaimaruseac let's chat about this. I can re-open the PR.

Issue #261, opened 2022-12-06
Reorganize internal testing data


pxp928:
The internal testing folder causes some import cycle errors when testing with NATS. Moving the data and reorg of the internal folder is needed to remove these and allow for it easier testing.

pxp928:
Closing PR as not a bug/issue. Can revisit in future if needed.

PR #262, opened 2022-12-06
reorg internal folder to consolidate

pxp928:

The internal folder was broken into processor and ingestor which caused confusion when importing the packages for testing. Reorganized the internal folder to be more consolidated and remove import cycle errors.

- Moved `DocTreeEqual` function into testdata.go such that it can be used by other packages when testing.

Closes issue #261


pxp928:
Closing this PR as cycle error still persists.

Issue #242, opened 2022-11-22
Testing requires more granular graphDB queries


lumjjb:
While writing some tests, having the ability to have more granular database queries is very helpful! However, the issue it brings about is that this muddles the abstraction of the graph DB being transparent (vs having graphQL queries for each thing).

The current approach is to continue with direct database queries for tests for now, and if/when the a new DB gets added to abstract these queries into debuggable graphQL interfaces.

pxp928:
Add missing unit tests to `graphdb.go` for `StoreGraph` and `Subscribe`

pxp928:
This is being reworked to create unit tests for the various backend resolvers.

pxp928:
Added unit tests to the in-memory backend resolvers (that can be used for conformance tests for others). @mihaimaruseac Close this issue as completed or is there more associated with this issue?

mihaimaruseac:
I think this is done. At the time we had the ability to get an entire subgraph out of neo4j and then the test needed to parse it, but now with graphql we can get all data we need.

PR #268, opened 2022-12-07
NATS Pub/Sub (NATS PR: 3 of 4)

pxp928:

Adds NATS Pub/Sub to assembler, collector, processor, parser and certifier along with respective unit tests.

Note: graphdb.go is missing unit test that will need to be added once the client abstraction is done. Issue https://github.com/guacsec/guac/issues/242

Currently in draft waiting for #255 and #256 to merge

Part 3 of 4 for issue https://github.com/guacsec/guac/issues/31


pkg/handler/collector/collector_test.go
```diff
@@ -80,3 +88,108 @@ func TestCollect(t *testing.T) {
...
+	}
+
+	var cancel context.CancelFunc
+
+	ctx, cancel = context.WithTimeout(ctx, 2*time.Second)
+	defer cancel()
+
+	docChan := make(chan processor.DocumentTree, 1)
+	errChan := make(chan error, 1)
+	defer close(errChan)
```
lumjjb, line 125: defer close(docChan)

pkg/handler/collector/collector_test.go
```diff
@@ -80,3 +88,108 @@ func TestCollect(t *testing.T) {
...
+				t.Errorf("doc tree did not match up, got\n%s, \nexpected\n%s", dochelper.StringTree(d), dochelper.StringTree(expectedDocTree))
+			}
+		case err := <-errChan:
+			if err != nil && !errors.Is(err, context.DeadlineExceeded) {
+				t.Errorf("nats emitter Subscribe test errored = %v", err)
+			}
+			subscribersDone += 1
+		}
+	}
+
```
lumjjb, line 146: need to drain the channel here as well, since the loop may end without deref'ing docChan

lumjjb (commented):
initial comments - love the tests!

pkg/handler/processor/process/process.go
```diff
@@ -51,12 +57,69 @@ func RegisterDocumentProcessor(p processor.DocumentProcessor, d processor.Docume
 	return nil
 }
 
+// Subscribe is used by NATS JetStream to stream the documents received from the collector
+// and process them them via Process
+func Subscribe(ctx context.Context) error {
```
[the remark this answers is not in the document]
pxp928: Could you elaborate on this? I like the idea of making it more generic.

pkg/handler/collector/collector_test.go
```diff
@@ -80,3 +88,108 @@ func TestCollect(t *testing.T) {
...
+		// if the context is canceled we want to break out of the loop
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		msgs, err := sub.Fetch(1)
+		if err != nil && errors.Is(err, nats.ErrTimeout) {
+			logger.Infof("[processor: %s] error consuming, backing off for a second: %v", id, err)
+			time.Sleep(1 * time.Second)
+			continue
+		}
```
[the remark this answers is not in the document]
pxp928: good catch. Added check.

pxp928:
Looking into fixing the broken unit test. Seems like the context is getting canceled before the test completes...

pxp928:
ready for re-review after refactor and abstraction

mlieberman85 (approved):
LGTM

pkg/emitter/nats_data.go
```diff
@@ -0,0 +1,52 @@
...
+	for {
+		select {
+		case d := <-psub.dataChan:
+			if err := processFunc(d); err != nil {
+				return err
+			}
+		case err := <-psub.errChan:
+			return err
+		}
+	}
```
lumjjb, line 51: I think there’s an edge case here, if an error is hit, we want it to end, but it may be missing some documents that it hasn’t processed…. so we would to store the error, do the drain and then return the error.

lumjjb: nit: Another thought here, as an alternative if you like it, is to have this just return an `interface{}, error` and then the wrapper around this function which is implemented for each of the modules would do the transformation

lumjjb (commented):
minor edge case else LGTM!

pkg/emitter/nats_data.go
pxp928: So NATS via the subscribe returns it via a `[]bytes`. So I didn't see a need to return an `interface{}` as this will always be consistent. The unmarshal would happen in each module.

pxp928: Agreed on the error drain. Fixing it now.

pxp928: @lumjjb thoughts on the above comment?

lumjjb (approved):
lgtm

PR #282, opened 2022-12-14
Add a GitHub collector

shafeeshafee:
WIP: Added base logic.


pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,140 @@
...
+		dir:           dir,
+		poll:          poll,
+		interval:      interval,
+		fileCollector: fileCollector,
+	}
+}
+
+func (g *githubDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	// Replace with your own personal access token, export to use
+	token := os.Getenv("API_KEY")
```
JonZeolla, line 47: Shouldn't this be more specific to a github api key / PAT?

shafeeshafee: @JonZeolla you're right it should. I'm going to make that more apparent as I push this scaffolding further. Thanks for the headsup!

mlieberman85: I would probably also pass this into the `githubDocumentCollector` and either call the environment variable, config file, etc. from what calls the githubDocumentCollector's constructor. I could also see falling back to looking for environment variable in the constructor if a token is not passed to it.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,140 @@
...
+type githubDocumentCollector struct {
+	url           string
+	dir           string
+	lastChecked   time.Time
+	poll          bool
+	interval      time.Duration
+	fileCollector collector.Collector
+}
+
+func NewGitHubDocumentCollector(ctx context.Context, url string, dir string, poll bool, interval time.Duration) *githubDocumentCollector {
```
mlieberman85, line 33: Add a comment here about what the GH token's permissions need to be, when you're ready.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,140 @@
...
+func (g *githubDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	// Replace with your own personal access token, export to use
+	token := os.Getenv("API_KEY")
+
+	// Replace with the owner and name of the repository
+	owner := "slsa-framework"
+	repo := "slsa-github-generator"
+
+	// Replace with the path of the directory where you want to download the assets
+	dir := "temp"
```
mlieberman85, line 54: Yep, this should be a dir in the struct.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (g *githubDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	// Replace with your own personal access token, export to use
+	token := os.Getenv("API_KEY")
+
+	// Replace with the owner and name of the repository
+	owner := "slsa-framework"
+	repo := "slsa-github-generator"
+
+	// Replace with the path of the directory where you want to download the assets
```
mlieberman85, line 53: Just a note it's often useful to add a comment like `TODO: ...` or `FIXME: ...` for these sorts of comments as you can easily search them while working so you know what you have to go back and fix, vs. what comments are just for readers to better understand what you're doing.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+	"go.uber.org/zap"
+	"golang.org/x/oauth2"
+)
+
+const (
+	CollectorGitHubDocument = "GitHubCollector"
+)
+
+type githubDocumentCollector struct {
+	dir           string
```
pxp928, line 24: remote dir as not used.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+	lastChecked   time.Time
```
pxp928, line 25: remove lastChecked as not used.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+	poll          bool
+	interval      time.Duration
+	token         string
+	owner         string
+	repo          string
+	tag           string
+	tagList       []string
+	latestRelease bool
```
pxp928, line 33: remove latestRelease as not used

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+
+func NewGitHubDocumentCollector(ctx context.Context, dir string, poll bool, interval time.Duration, logger *zap.SugaredLogger, token string, owner string, repo string, tag string, latestRelease bool) *githubDocumentCollector {
+	return &githubDocumentCollector{
+		dir:           dir,
+		poll:          poll,
+		interval:      interval,
+		token:         token,
+		owner:         owner,
+		repo:          repo,
+		tag:           tag,
```
pxp928, line 44: `tagList` needs to be initialized as an empty slice. `[]string{}`

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+	if g.poll {
+		for {
+			if ctx.Err() != nil {
+				return nil
+			}
+			err := g.fetchAssets(ctx, logger, docChannel)
+			if err != nil {
+				return err
+			}
+			g.lastChecked = time.Now()
```
pxp928, line 60: dont need `lastChecked`

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,163 @@
...
+			defer resp.Body.Close()
+
+			bytes, err := io.ReadAll(resp.Body)
+			if err != nil {
+				return err
+			}
+
+			var sourceString string
+
+			if g.latestRelease {
```
pxp928, line 142: use tag == nil like above.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,159 @@
...
+
+func NewGitHubDocumentCollector(ctx context.Context, poll bool, interval time.Duration, logger *zap.SugaredLogger, token string, owner string, repo string, tag string, tagList []string) *githubDocumentCollector {
+	return &githubDocumentCollector{
+		poll:     poll,
+		interval: interval,
+		token:    token,
+		owner:    owner,
+		repo:     repo,
+		tag:      tag,
+		tagList:  tagList,
```
pxp928, line 41: initialize via []string{}. It will not be an input from the user.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (g *githubDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	logger := logging.FromContext(ctx)
+	if g.poll {
+		for {
+			if ctx.Err() != nil {
+				return nil
+			}
+			err := g.fetchAssets(ctx, logger, docChannel)
```
pxp928, line 52: Need to check that a tag is not passed in. As the poll should always be querying for the latest release.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,159 @@
...
+	} else {
+		// get the release with the specified tag
+		release, _, err = client.Repositories.GetReleaseByTag(ctx, g.owner, g.repo, g.tag)
+	}
+	if err != nil {
+		logger.Debug(err)
+		return err
+	}
+	// Add the current tag to the tagList if it has not been seen before
+	currentTag := release.GetTagName()
```
pxp928, line 98: currentTag and g.tag should be the same thing so need for this?

PR #284, opened 2022-12-14
fix: viper flag name missmatch

pxp928:

Fix the flag name miss-match that was causing the flags to never get set to the based on the input from the cli.


mlieberman85 (approved):
Lgtm

PR #314, opened 2023-01-15
feat: L1 Policy for checking duplicates

pxp928:

TODO: Missing unit test and doc comments


pkg/cache/cache.go
```diff
@@ -0,0 +1,89 @@
+//
```
mlieberman85, line 1: Should we put the redis cache behind an interface in case we end up going with something else>

pxp928: good call. Added interface.

pxp928:
Closing for more research

Issue #322, opened 2023-01-17
Is There a Difference Between Structs


nathannaveen:
Hi, I was trying to test `pkg/certifier/osv/internal/osv_query/osv_query.go` and saw that there are the structs, `BatchedResponse` and `HydratedBatchedResponse`:

https://github.com/guacsec/guac/blob/3dd89315ddecf795fef1c5374cea77246d37aaf2/pkg/certifier/osv/internal/osv_query/osv_query.go#L80-L88

`HydratedBatchedResponse` only has ``Results []Response `json:"results"` `` in it, and `BatchedResponse` has ``Results []MinimalResponse `json:"results"` `` in it.

`Response` is:

``` go
type Response struct {
	Vulns []osv.Entry `json:"vulns"`
}
```

And `MinimalResponse` is:

``` go
type MinimalResponse struct {
	Vulns []osv.Entry `json:"vulns"`
}
```

So what makes `HydratedBatchedResponse` and `BatchedResponse` different, is there a reason they are two different structs?

nathannaveen:
cc @pxp928, 👀

pxp928:
@nathannaveen This file was copied over (and slightly modified) from the upstream project (now OSV Scanner). I did create another issue #323 to track that we eventually remove this from GUAC. Feel free to remove the redundant structs but overall the whole file will be removed in the near future once I speak with the OSV Scanner maintainers.

PR #325, opened 2023-01-18
Refactored osv_query

nathannaveen:
- Refactored the file `pkg/certifier/osv/internal/osv_query/osv_query.go`
- Removed redundant structs which fixes: https://github.com/guacsec/guac/issues/322
- Removed the `MakeCommitRequest` struct because it wasn't ever used
- Removed the `Hydrate()` function because it isn't ever used
- Also removed `Get()` because it was only used in `Hydrate()`


lumjjb (approved):
lgtm

PR #337, opened 2023-01-23
Ensure recursion termination in certifyHelper in osv

### LINKED ISSUE #336: Stack overflow in certifyHelper function in osv
## Description

I have identified a potential issue with the certifyHelper https://github.com/guacsec/guac/blob/86e83206e7df9a21ac5da35636b819d6152a1cba/pkg/certifier/osv/osv.go#L70-L83 function in our codebase. The function has the potential to stack overflow if the dependencies of the input component are cyclical. This can happen if a component's dependencies include itself, or if two or more components mutually depend on each other.

A test to check for stack overflow can be created by creating a cyclical dependency between two or more components. One way to do this is to create a component, A, that has a dependency on B, and then create B with a dependency on A. Then call o.certifyHelper(ctx, A, docChannel). If a stack overflow occurs, the function will not return and the program will crash.

It is essential to check that the function returns an error when it detects a cyclical dependency to let the user know that the function has not been executed.

## Steps to Reproduce

- Create a cyclical dependency between two or more components

- Call the certifyHelper function with one of the components as the input
- Observe that the function does not return, and the program crashes.
``` go
func TestCertifyHelperStackOverflow(t *testing.T) {
	var A, B, C *certifier.Component
	// Create a cyclical dependency between two components
	A = &certifier.Component{
		Package: assembler.PackageNode{
			Name: "example.com/A",
		},
	}

	B = &certifier.Component{
		Package: assembler.PackageNode{
			Purl: "example.com/B",
		},
	}
	C = &certifier.Component{
		Package: assembler.PackageNode{
			Purl: "example.com/C",
		},
	}
	A.DepPackages = []*certifier.Component{B, C}
	B.DepPackages = []*certifier.Component{C, A}
	C.DepPackages = []*certifier.Component{A, B}
	// Create a channel to receive the generated documents
	docChannel := make(chan *processor.Document)
	o := NewOSVCertificationParser()
	// Create a context to cancel the function if it takes too long
	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
	defer cancel()
	err := o.CertifyComponent(ctx, A, docChannel)
	// Call certifyHelper with the cyclical dependency
	if err != nil {
		t.Errorf("Expected no error, got %v", err)
	}
	if err == context.DeadlineExceeded {
		t.Errorf("Function did not return an error, but it took too long to execute, which indicates stack overflow")
	}
	if err == nil {
		t.Errorf("Function did not return an error, but it should have, since a cyclical dependency was detected")
	}
}
```
## Expected behavior

The function should detect the cyclical dependency and return an error to let the user know that the function has not been executed.

## Actual behavior

The function does not return and the program crashes.

naveensrinivasan:
- Fixed the recursion bug
- Included tests for recursion bug.
- Fixes #336 

### DIFF
--- a/pkg/certifier/attestation/attestation_vuln.go
+++ b/pkg/certifier/attestation/attestation_vuln.go
@@ -21,10 +21,10 @@ import (
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 )
 
-// This is a new predicate type for vulnerabilities based off
+// PredicateVuln This is a new predicate type for vulnerabilities based off
 // https://github.com/sigstore/cosign/blob/main/specs/COSIGN_VULN_ATTESTATION_SPEC.md.
 // This is used by the certifier to attest to vulnerabilities in an artifact.
-// Currently the predicate is defined here but the intention is to upstream this to
+// Currently, the predicate is defined here but the intention is to upstream this to
 // https://github.com/in-toto/attestation in the near future once the quirks are worked out.
 const (
 	PredicateVuln = "https://in-toto.io/attestation/vuln/v0.1"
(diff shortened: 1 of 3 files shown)


pxp928 (approved):
LGTM! Thanks @naveensrinivasan!

lumjjb (approved):

Issue #350, opened 2023-01-23
Potential - Stack Overflow in fetchOCIArtifacts function


naveensrinivasan:
Description:
The `fetchOCIArtifacts` function in the OCI collector is prone to a stackoverflow when processing images with many platform-specific digests and layers.
https://github.com/guacsec/guac/blob/29653250d1d5c81163b09e2d91620f6be371a9e7/pkg/handler/collector/oci/oci.go#L142-L163

pxp928:
@naveensrinivasan should is issue close also?

PR #351, opened 2023-01-24
Avoid recursion for OCI

naveensrinivasan:
- Implemented non recursive function calls for OCI


pxp928:
@sudo-bmitch would like your thoughts on this PR. Related to issue https://github.com/guacsec/guac/issues/350

naveensrinivasan:
I am unable to create a stackoverflow, and as such, this hypothesis does not provide a viable solution. Therefore, I am closing this PR.

PR #355, opened 2023-01-25
Replaced returning err with fmt.Errorf

nathannaveen:
- Replaced the `return err` with `return fmt.Errorf()`

Returning `fmt.Errorf()` instead of `return err` in Go is generally considered to be best practice because it allows for more detailed and informative error messages. `fmt.Errorf()` is a function that formats a string and returns an error, which means that developers can include additional context and information in the error message that can be helpful for debugging and troubleshooting. In contrast, simply returning `err` without any additional context can make it more difficult to understand the cause of the error and how to fix it. Additionally, by returning a formatted error message, it makes error handling and debugging more consistent.

### DIFF
--- a/pkg/handler/collector/git/git.go
+++ b/pkg/handler/collector/git/git.go
@@ -17,6 +17,7 @@ package git_collector
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"time"
 
@@ -73,7 +74,7 @@ func (g *gitDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel
 			}
 			err := g.createOrPull(ctx, logger, docChannel)
 			if err != nil {
-				return err
+				return fmt.Errorf("error creating git repo: %v", err)
 			}
 			g.lastChecked = time.Now()
 			time.Sleep(g.interval)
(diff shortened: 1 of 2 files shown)


pkg/handler/collector/git/git.go
```diff
@@ -73,15 +74,15 @@ func (g *gitDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel
 			}
 			err := g.createOrPull(ctx, logger, docChannel)
 			if err != nil {
-				return err
+				return fmt.Errorf("error creating git repo: %v", err)
```
pxp928, line 77: Change to use `%w` instead of `%v` to wrap the error. Similar for the others.

pkg/handler/collector/git/git.go
```diff
@@ -73,15 +74,15 @@ func (g *gitDocumentCollector) RetrieveArtifacts(ctx context.Context, docChannel
...
 			}
 			g.lastChecked = time.Now()
 			time.Sleep(g.interval)
 		}
 	} else {
 		err := g.createOrPull(ctx, logger, docChannel)
 		if err != nil {
-			return err
+			return fmt.Errorf("error pulling git repo: %v", err)
```
pxp928, line 85: this should be the same error message as the one above?

pxp928 (commented):
few nits. Also I think the tests might have to be fixed up based on these changes.

pxp928 (approved):
LGTM

lumjjb (approved):
LGTM! Thanks!

PR #356, opened 2023-01-25
Metrics - Prometheus

naveensrinivasan:
This PR adds support for Prometheus metrics.

 The implementation includes the following changes:
- Defining an interface `Metrics` that defines the methods for creating and incrementing the Prometheus metrics
- Create a struct `prom` that implements the `Metrics` interface
- Includes tests 

This change allows for decoupling the implementation of the metrics from the main implementation and easily swap out the Prometheus implementation for another metrics implementation if needed.


naveensrinivasan:
@pxp928 @lumjjb @mihaimaruseac  👀

lumjjb (approved):
LGTM! I think something like that - for metrics and counters we want to store, we will need to have a design doc before additional impl. However, this PR is enough of an isolated library. 

I think one thing that sticks out to me, is whether this should be in `internal/`.

mihaimaruseac (approved):

naveensrinivasan:
> LGTM! I think something like that - for metrics and counters we want to store, we will need to have a design doc before additional impl. However, this PR is enough of an isolated library.



I've designed this in a GoKit style so that any package can use it. 
Could you explain why we need a design document? 

It makes sense to create a design document for larger implementations. As we design, the metrics counters and summary will likely evolve, which is the natural process of software development.

> I think one thing that sticks out to me, is whether this should be in `internal/`. 

Keeping this package internal would prevent consumers from being able to use the  create their own implementation `metrics`implementation. What are your thoughts on this?

In my opinion, it would be beneficial to keep the number of internal packages to a minimum. Thoughts?

mihaimaruseac:
So, regarding design docs: for each major piece of infrastructure, it is better to have a design doc to see how it is planned to be deployed, what alternatives are considered, what are the benefits of picking this approach compared to the others, what are the downsides. Without this, it is very likely that another PR could implement a similar functionality or that a few months after the feature lands we discover that it is not doing a key critical job and needs to be scrubbed.

Like, for this PR alone, I am wondering what metrics we'd like to collect and why collecting them from Go vs collecting them from elsewhere (e.g. from database backend on GCP - when that exists) is better. I'm sure a design review can uncover even more questions and can crystallize usage pattens. Plus, it serves as documentation!

Issue #281, opened 2022-12-13
 call the collector.Collect function during collector tests.


shafeeshafee:
This issue was brought up in this discussion: https://github.com/guacsec/guac/pull/228#discussion_r1043236451

pxp928:
Update `collector.go` to output an specific error message that can be compared for `RegisterDocumentCollector`:
```go
var (
	documentCollectors    = map[string]Collector{}
	ErrCollectorOverwrite = fmt.Errorf("the document collector is being overwritten")
)

func collectorTypeOverwriteError(collectorType string) error {
	return fmt.Errorf("%w: %s", ErrCollectorOverwrite, collectorType)
}

func RegisterDocumentCollector(c Collector, collectorType string) error {
	if _, ok := documentCollectors[collectorType]; ok {
		documentCollectors[collectorType] = c
		return collectorTypeOverwriteError(collectorType)
	}
	documentCollectors[collectorType] = c

	return nil
}
```
This needs to be done for all the collector types (git, gcs, file..etc) but for now this example will show how to do it for OCI collector. So in `oci_test.go`, register the oci collector, create an `emit` function that collects the documents, and an `errhandler` function to handle errors. Pass the context, `emit` function, and `errhandler` function into `collector.collect`:
```go
		t.Run(tt.name, func(t *testing.T) {
			ctx := context.Background()
			g := NewOCICollector(ctx, tt.fields.repoTags, tt.fields.poll, tt.fields.interval)

			// Register collector
			err := collector.RegisterDocumentCollector(g, OCICollector)
			if err != nil && !errors.Is(err, collector.ErrCollectorOverwrite) {
				t.Errorf("unexpected error: %v", err)
			}

			var cancel context.CancelFunc
			if tt.fields.poll {
				ctx, cancel = context.WithTimeout(ctx, 10*time.Second)
				defer cancel()
			}

			collectedDocs := []*processor.Document{}

			emit := func(d *processor.Document) error {
				collectedDocs = append(collectedDocs, d)
				return nil
			}

			// Collect
			errHandler := func(err error) bool {
				if err != nil {
					if !tt.wantErr {
						t.Errorf("g.RetrieveArtifacts() error = %v, wantErr %v", err, tt.wantErr)
						return false
					}
					if !strings.Contains(err.Error(), tt.errMessage.Error()) {
						t.Errorf("g.RetrieveArtifacts() error = %v, wantErr %v", err, tt.errMessage)
						return false
					}
				}
				return true
			}
[...]

PR #358, opened 2023-01-26
Cleaned Up Code in git_test

nathannaveen:
- Cleaned up some code in `git_test.go`
- There is no need for `numCollectors` and `collectorsDone`, because `numCollectors` is always 1 and `collectorsDone` is only incremented when `case err := <-errChan`

### DIFF
--- a/pkg/handler/collector/git/git_test.go
+++ b/pkg/handler/collector/git/git_test.go
@@ -105,12 +105,10 @@ func Test_gitCol_RetrieveArtifacts(t *testing.T) {
 				errChan <- g.RetrieveArtifacts(ctx, docChan)
 			}()
 
-			numCollectors := 1
-			collectorsDone := 0
-
+			hasErrored := false
 			collectedDocs := []*processor.Document{}
 
-			for collectorsDone < numCollectors {
+			for !hasErrored {
 				select {
 				case d := <-docChan:
 					collectedDocs = append(collectedDocs, d)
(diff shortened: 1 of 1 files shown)


mihaimaruseac (commented):
I think I prefer the first version since it is easier to update to run multiple collectors in parallel during the test.

pxp928 (commented):
Hey @nathannaveen! Thanks for the PR. So there is issue https://github.com/guacsec/guac/issues/281 where we wanted to call the `collector.Collect` in all the collector tests (i.e. file, gcs, git...etc). That would be a better approach to cleaning up the collector tests. Let me know and I can expand on the issue if you need more guidance.

naveensrinivasan:
> I think I prefer the first version since it is easier to update to run multiple collectors in parallel during the test.

Wouldn't it better to do that when we have multiple runners? Thoughts?

nathannaveen:
> Hey @nathannaveen! Thanks for the PR. So there is issue #281 where we wanted to call the `collector.Collect` in all the collector tests (i.e. file, gcs, git...etc). That would be a better approach to cleaning up the collector tests. Let me know and I can expand on the issue if you need more guidance.

@pxp928 I can take it up, but I don't fully understand, could you explain? Thanks

pxp928:
> @pxp928 I can take it up, but I don't fully understand, could you explain? Thanks

@nathannaveen I updated the issue with an example that should make it easier to understand. That will have to be done for all the collector tests (file, gcs, git and oci). Let me know if you need more guidance

pxp928:
Closing PR as there is a better path forward on issue #281

PR #377, opened 2023-01-28
Included dependatbot for tools

naveensrinivasan:
- Included dependabot for tools

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,11 @@ updates:
     schedule:
       interval: weekly
 
+  - package-ecosystem: gomod
+    directory: "/tools"
+    schedule:
+      interval: weekly
+
   # Maintain dependencies for GitHub Actions.
   - package-ecosystem: "github-actions"
     directory: "/"


.github/dependabot.yml
```diff
@@ -6,6 +6,11 @@ updates:
     schedule:
       interval: weekly
 
+  - package-ecosystem: gomod
+    directory: "/tools"
+    schedule:
+      interval: weekly
+
```
mihaimaruseac, line 13: Isn't this already covered by the previous stanza?

mihaimaruseac (changes requested):

.github/dependabot.yml
mihaimaruseac: It should be covered, see https://github.com/guacsec/guac/pull/368 which is for a dependency from there.

naveensrinivasan: No, it only updates the go.mod in the root dir and not within the tools dir. That's why you didn't see a PR for the tools dir go.mod

mihaimaruseac: There is no `tools/go.mod`

naveensrinivasan: OK, thanks. why tools.go without a separate module?

mihaimaruseac: It doesn't need one. It is there only to enable `go generate ./...` do the right thing, but the dependencies it has have to be in the root `go.mod`.

naveensrinivasan: I think it's a great idea to keep tool dependencies in a separate module; it helps with CVEs, maintenance, and even bringing in additional dependencies. For example, projects like Scorecard and Rekor follow this pattern.

Thoughts?

mihaimaruseac: The same dependencies are used in the rest of the repository. I don't think it is worthwhile to separate.

Looking at Scorecards, it seems the separation also originated from you (https://github.com/ossf/scorecard/commit/2b15b1353bc5c759b85871445027fadc9d5bc3c1). None of the reasons presented there apply here.

naveensrinivasan: OK.

Issue #376, opened 2023-01-28
New package for nodes.go


nathannaveen:
In my opinion, [`pkg/assembler/nodes.go`](https://github.com/guacsec/guac/blob/main/pkg/assembler/nodes.go) should be split up into multiple files, and all those files should be in a separate package.

For example, we could have a package called `nodes` in `pkg/assembler`, and `nodes` should have the files, `artifactnode.go`, `packagenode.go`, `identitynode.go`, and so on, inside it.

nathannaveen:
On second thought, splitting up `nodes.go` into multiple files without the package should be a good idea whether or not it is put into a different package.

mihaimaruseac:
I would say don't focus on this package given that it will be obsolete soon by [the GraphQL assembler work](https://docs.google.com/document/d/1yZ3-ZcfnRDWgw9uZlPuLmIHS9pNMr3DO_AEbHsDXmN8/edit)

pxp928:
Closing this issue as its no longer relevant with the changes made with graphQL

PR #380, opened 2023-01-28
Refactor node.go into multiple files

nathannaveen:
- Refactored `node.go` into multiple files because it was too big and hard to manage.
- Also wrote some tests for a couple of the refactored files.
- Kind of works in with this issue: https://github.com/guacsec/guac/issues/376


mihaimaruseac:
I would say don't focus on this package given that it will be obsolete soon by [the GraphQL assembler work](https://docs.google.com/document/d/1yZ3-ZcfnRDWgw9uZlPuLmIHS9pNMr3DO_AEbHsDXmN8/edit)

nathannaveen:
> I would say don't focus on this package given that it will be obsolete soon by [the GraphQL assembler work](https://docs.google.com/document/d/1yZ3-ZcfnRDWgw9uZlPuLmIHS9pNMr3DO_AEbHsDXmN8/edit)

Ok, makes sense.

Issue #323, opened 2023-01-18
Certifier - OSV Scanner API for batch querying


pxp928:
Currently, the OSV Scanner querying API is copied as an internal file. Working with upstream maintainers to determine a path forward to remove this [file](https://github.com/guacsec/guac/blob/main/pkg/certifier/osv/internal/osv_query/osv_query.go).

Continuation from PR comment:
https://github.com/guacsec/guac/pull/245#discussion_r1047996505

cc @oliverchang

pxp928:
Currently, OSV Scanner does not allow for a direct query for a package based on a purl. See: https://github.com/google/osv-scanner/blob/main/pkg/osvscanner/osvscanner.go#L261. Determine if there is an alternative method or if this can be added to upstream OSV scanner.

pxp928:
FYI @lumjjb

oliverchang:
Currently our actual OSV querying code is in an internal package: https://github.com/google/osv-scanner/blob/main/internal/osv/osv.go

@another-rex would it make sense to make some of these public so other tools can call the OSV API through our client?

another-rex:
That can work, we can put it under v1 folder under pkgs so that it's public and if there are any breaking changes it easy to update.

pxp928:
Awesome thanks @another-rex!

another-rex:
This is now merged: https://github.com/google/osv-scanner/pull/167, so the library should now be public:
> "github.com/google/osv-scanner/pkg/osv"

pxp928:
Thanks, @another-rex and @oliverchang!

PR #383, opened 2023-01-30
remove internal OSV query and pull from upstream OSV_scanner project

pxp928:

removes the internal osv query file now that the pkg is available in the upstream OSV scanner. Fix error message mismatch causing the CI to fail. 

closes https://github.com/guacsec/guac/issues/323


lumjjb (approved):
lgtm!

mihaimaruseac (approved):

PR #389, opened 2023-01-31
Make sure every PR syncs generated files and generate schema again

mihaimaruseac:
We bumped the gqlgen dependency in https://github.com/guacsec/guac/pull/368, so the generated files changed. We are bringing them back to date now. Also making sure that future CI will test that generated files are up to date.

Cleaned up Makefile formatting while here, making it more consistent.


Makefile
```diff
@@ -8,34 +8,46 @@ LDFLAGS="-X $(PKG).version=$(VERSION) -X $(PKG).commit=$(COMMIT) -X $(PKG).date=
 .DEFAULT_GOAL := build
 
 .PHONY: all
-all: test cover fmt lint ci build generate
+all: test cover fmt lint ci build generate generated_up_to_date
```
lumjjb, line 11: hmm i think this wouldn't need it, since it is part of the `ci`?

lumjjb (approved):
minor nit, else lgtm

Makefile
mihaimaruseac: Right, let me remove.

pxp928 (approved):
LGTM

PR #391, opened 2023-02-01
Add package trie GraphQL schema (before qualifiers)

mihaimaruseac:
I'm adding the package trie from the ontology all the way to the qualifiers. Stopping at the qualifiers as we need to resolve a few questions for that, will submit a new PR with those when fixed.

Moving all schema files to a `schema/` directory to ease readability. Will delete the old schema as soon as a new schema file gets added for other parts of the ontology.

Brought back the ability to run 2 backends. To run the neo4j backend you have to `go run pkg/assembler/graphql/main/server.go` once the neo4j container is set up, data ingested. To run the test backend (which now has some hardcoded testdata based on example pURLs -- easier to hardcode this data than to ingest into Neo4j at this moment, but will do that too shortly) you need to define the env variable, so `BACKEND=testing go run pkg/assembler/graphql/main/server.go`

To test with the hardcoded backend, here are the queries I've used

```gql
query Q1 {
  artifacts {
    name
    sourceInfo
    collectorInfo
  }
}

fragment allPkgTree on Package {
  type
  namespaces {
    namespace
    names {
      name
      versions {
        version
      }
    }
  }
}

# all packages
query Q2 {
  packages(pkgSpec: {}) {
    ...allPkgTree
  }
}

# only PyPI packages
query Q3 {
  packages(pkgSpec: {type: "pypi"}) {
    ...allPkgTree
  }
}

# PyPI packages in "debian" namespace (empty)
query Q4 {
  packages(pkgSpec: {type: "pypi", namespace: "debian"}) {
    ...allPkgTree
  }
}

# debian packages in "debian" namespace
query Q5 {
  packages(pkgSpec: {type: "deb", namespace: "debian"}) {
    ...allPkgTree
  }
}

# "dpkg" package from Ubuntu ("ubuntu" namespace on "deb" type)
query Q6 {
  packages(pkgSpec: {type: "deb", namespace: "ubuntu", name: "dpkg"}) {
    ...allPkgTree
  }
}

# All "deb" type packages named "dpkg"
query Q7 {
  packages(pkgSpec: {type: "deb", name:"dpkg"}) {
    ...allPkgTree
  }
}

# All occurences of openssl-3.0.3
query Q8 {
  packages(pkgSpec: {name:"openssl", version:"3.0.3"}) {
    ...allPkgTree
  }
}

# All occurences of openssl
query Q9 {
  packages(pkgSpec: {name:"openssl"}) {
    ...allPkgTree
  }
}
```

Each commit can be reviewed individually if so desired, I tried to group the changes for readability.


pkg/assembler/graphql/model/nodes.go
```diff
@@ -184,17 +184,87 @@ type Package struct {
...
 
-func (Package) IsArtifactOrPackage() {}
+// PackageNamespace is a namespace for packages.
+//
+// In the pURL representation, each PackageNamespace matches the
+// `pgk:<type>/<namespace>/` partial pURL.
+//
+// Namespaces are optional and type specific. Because they are optional, we use
+// empty string to denote missing namespaces.
+type PackageNamespace struct {
```
pxp928, line 237: nit: swap this with `PackageName` for readability

pxp928 (approved):
LGTM just a small nit.

pkg/assembler/graphql/model/nodes.go
pxp928: generated file. NVM

pkg/assembler/backends/testing/backend.go
```diff
@@ -25,12 +25,330 @@ import (
...
+	packages []*model.Package
+}
+
+// Define some demo packages to test the query without also ingesting into Neo4j
+var demo demoClient
 
 func GetBackend(args backends.BackendArgs) (backends.Backend, error) {
-	return &demoClient{}, nil
+	demo = demoClient{
+		packages: []*model.Package{
```
lumjjb, line 37: hmm, would it be possible to make a helper function for this, to create the structure, may make it a bit more readable.

lumjjb (approved):
minor nits, else LGTM

pkg/assembler/backends/testing/backend.go
mihaimaruseac: Moved to a separate file with utilities. Should be more readable now. PTAL

pxp928: the refactor looks good. Easier to read and keep adding more packages as needed.

mihaimaruseac: Thanks, resolving this to get this merged

PR #393, opened 2023-02-01
Delete old schema.

mihaimaruseac:
Now that we have a new schema in place, we should be able to delete the old schema and avoid name conflicts.

I'm deleting everything and pushing the generated code all in the same commit since this is a no-functional-change type of PR.

Depends on #391, it includes those changes. Will rebase once that lands.


pxp928 (approved):
LGTM. good to delete

lumjjb (approved):
lgtm

PR #395, opened 2023-02-02
add heuristic for SLSA subject names with PURL

lumjjb:
add heuristic for SLSA subject names with PURL. In the case where a PURL is present in SLSA, provide a package node to link so that it links with other metadata with the PURL. Right now it doesnt do this as it is stored as an artifact nodes and won't match against a package node.

### DIFF
--- a/pkg/ingestor/parser/slsa/parser_slsa.go
+++ b/pkg/ingestor/parser/slsa/parser_slsa.go
@@ -34,11 +34,13 @@ const (
 )
 
 type slsaParser struct {
-	doc          *processor.Document
-	subjects     []assembler.ArtifactNode
-	dependencies []assembler.ArtifactNode
-	attestations []assembler.AttestationNode
-	builders     []assembler.BuilderNode
+	doc                     *processor.Document
+	subjects                []assembler.ArtifactNode
+	dependencies            []assembler.ArtifactNode
+	attestations            []assembler.AttestationNode
+	builders                []assembler.BuilderNode
+	heuristicSubjectPackage []assembler.PackageNode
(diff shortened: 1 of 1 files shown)


mihaimaruseac (approved):

lumjjb:
@ianlewis

pkg/ingestor/parser/slsa/parser_slsa.go
```diff
@@ -68,10 +70,28 @@ func (s *slsaParser) Parse(ctx context.Context, doc *processor.Document) error {
...
+		var nodes []assembler.ArtifactNode
 		for alg, ds := range sub.Digest {
-			s.subjects = append(s.subjects, assembler.ArtifactNode{
+			nodes = append(nodes, assembler.ArtifactNode{
 				Name: sub.Name, Digest: alg + ":" + strings.Trim(ds, "'"), NodeData: *assembler.NewObjectMetadata(s.doc.SourceInformation)})
 		}
+		s.subjects = append(s.subjects, nodes...)
+
+		// Add heuristic to link an artifact to a package if the subject name starts with "pkg:"
+		if strings.HasPrefix(sub.Name, "pkg:") {
```
ianlewis, line 81: Unless the subject is specially crafted I don't imagine it would start with `pkg:`. What are we imagining the use case would be here?

I'm thinking we could improve the heuristics here so that we can match container image artifacts imported from SLSA and packages imported from sboms by

1. normalizing names for images that are on docker Hub (e.g. ianmlewis/demo vs. docker.io/ianmlewis/demo vs. index.docker.io/ianmlewis/demo)
2. Linking package and artifact by name and digest for the top-level SBOM package.

PR #396, opened 2023-02-02
Handle subpath in package trie.

mihaimaruseac:
Query to test

```gql
fragment allPkgTree on Package {
  type
  namespaces {
    namespace
    names {
      name
      versions {
        version
        qualifiers {
          key
          value
        }
        subpath
      }
    }
  }
}

query QB {
  packages(pkgSpec: {subpath: "subpath"}) {
    ...allPkgTree
  }
}
```

This concludes the package trie, we can now move to the rest of the ontology.

### DIFF
--- a/pkg/assembler/graphql/model/nodes.go
+++ b/pkg/assembler/graphql/model/nodes.go
@@ -84,9 +84,18 @@ type PackageQualifierInput struct {
 // For this level of GUAC, these are just opaque strings.
 //
 // This node can be referred to by other parts of GUAC.
+//
+// Subpath and qualifiers are optional. Lack of qualifiers is represented by an
+// empty list and lack of subpath by empty string (to be consistent with
+// optionality of namespace and version). Two nodes that have different qualifiers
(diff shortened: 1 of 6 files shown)


pxp928 (approved):
lgtm

pkg/assembler/graphql/model/nodes.go
```diff
@@ -103,4 +112,5 @@ type PkgSpec struct {
 	Name       *string                  `json:"name"`
 	Version    *string                  `json:"version"`
 	Qualifiers []*PackageQualifierInput `json:"qualifiers"`
```
lumjjb, line 114: just occurred to me that if we want to look for "no qualifiers" we would not be able to differentiate from all qualifiers vs no qualifiers.

lumjjb (approved):

pkg/assembler/graphql/model/nodes.go
mihaimaruseac: Right. I have a TODO to think about this but likely it needs a different PR.

PR #398, opened 2023-02-02
Add source trie GraphQL schema

pxp928:
Add source trie with type, namespace, name, and qualifier that can be specified as either a tag or commit, or nothing. Testing done via the testing backend. Neo4j resolver implementation in another PR.

```
fragment allSrcTree on Source {
  type
  namespaces {
    namespace
    names {
      name
      tag
      commit
    }
  }
}

query Q1 {
  sources(sourceSpec: {name: "github.com/guacsec/guac"}) {
    ...allSrcTree
  }
}

query Q2 {
  sources(sourceSpec: {qualifier: {tag: "v0.0.1"}}) {
    ...allSrcTree
  }
}

query Q3 {
  sources(
    sourceSpec: {qualifier: {commit: "fcba958b73e27cad8b5c8655d46439984d27853b"}}
  ) {
    ...allSrcTree
  }
}

query Q4 {
  sources(sourceSpec: {qualifier: {}}) {
    ...allSrcTree
  }
}

query Q5 {
  sources(sourceSpec: {type: "svn"}) {
    ...allSrcTree
  }
}

query Q6 {
  sources(sourceSpec: {namespace: "gitlab"}) {
    ...allSrcTree
  }
}

query Q7 {
  sources(sourceSpec: {}) {
    ...allSrcTree
  }
}

query Q8 {
  sources(sourceSpec: {qualifier: {tag: "asd", commit: "sad"}}) {
    ...allSrcTree
  }
}
```


pkg/assembler/graphql/schema/source.graphql
```diff
@@ -0,0 +1,104 @@
...
+"""
+SourceQualifier containers the commit or tag.
+
+Either a tag or commit needs to be specified.
+
+This node can be referred to by other parts of GUAC.
+"""
+type SourceQualifier {
+  tag: String!
+  commit: String!
```
mihaimaruseac, line 73: This would mean that both of them need to be specified / will be present in the output.

I'm actually thinking of making them `String` and pushing them directly into the `SourceName`. Like we do with `subpath` in the package trie

mihaimaruseac (approved):
LGTM, just a suggestion to simplify query and output

pkg/assembler/graphql/schema/source.graphql
```diff
@@ -0,0 +1,104 @@
...
+}
+
+"""
+SourceQualifierInput is the same as SourceQualifier, but usable as query
+input.
+"""
+input SourceQualifierInput {
+  tag: String
+  commit: String
+}
```
mihaimaruseac, line 99: I'm thinking to have the spec contain both the `tag` and `commit` fields, making it an error to have both. This way, the query is shorter and more readable: `source(sourceSpec: {tag:"v1.12"})` instead of `source(sourceSpec: {qualifier: {tag:"v1.12"}})` (we still need to error on qualifier if both `tag` and `commit` are given).

pkg/assembler/graphql/schema/source.graphql
pxp928: So if there was another release based on another commit, we would create a new `sourceName` node?

pkg/assembler/graphql/schema/source.graphql
pxp928: Now that I think about it more, does it matter if its a tag or commit? Can just be a string stored along with the `sourceName` node.

pkg/assembler/graphql/schema/source.graphql
mihaimaruseac: Yes, we do the same for packages: one node for PackageVersion with no subpath, another for a subpath, another for another subpath and so on, including new nodes for each classifier.

pkg/assembler/graphql/schema/source.graphql
mihaimaruseac: It depends if we want to push processing to upstream or to backend. If caller needs to interpret them, then it can be just a string and the query will have to be `source(sourceSpec: {comittish:"tag=v1.12"")`.

pxp928: I was thinking more like `source(sourceSpec: {comittish:"v1.12"")` or source(sourceSpec: {comittish:"fcba958b73e27cad8b5c8655d46439984d27853b"")

mihaimaruseac: I think it is possible to create a tag that looks like a commit, so there could be a source of confusion.

Even GitHub has https://github.com/guacsec/guac/commit/1f083269b1deec24d0a355f9b5c9b2f3e591906c vs https://github.com/guacsec/guac/tree/v0.0.1 (`commit/<commit>` vs `tree/<tag>`)

pxp928: ah good point.

mihaimaruseac: I'd still say we can merge these into the spec to simplify query and resolver a little bit more.

mihaimaruseac (approved):

pkg/assembler/graphql/schema/source.graphql
mihaimaruseac: Discussed offline, this change query semantics so we won't do it.

lumjjb (approved):
LGTM, I will open up a ticket for helping to flesh out docs for graphQL, think it will be helpful in the future.

Issue #244, opened 2022-11-22
Node expansion collector (parser request for more data with collector subscription)


mihaimaruseac:
In GUAC might have ingested nodes which contain references to artifact/package nodes/identifiers that the current populated GUAC database lacks metadata on - for example ingesting an SBOM with references to a PURL or github repo in which the collectors have not been configured to collect.

In this case, we want GUAC to be able to be intentional with seeking out information from configured sources/collectors to download this information. e.g., a private guac instance querying about the public one, or private GUAC instance upon seeing a PURL, to query deps.dev to ingest additional metadata.

Requirement

We need a mechanism to be able to fetch additional metadata to ingest into GUAC.
Augmenting the query protocol to handle user interactions when answering these types of queries when there is no/partial information. (e.g. provide feedback to the user on lack of data, and potentially the status of retrieving additional metadata)

lumjjb:
@mihaimaruseac 

Rewrote this to expand a bit more, if you think this reflect the intent, shall we replace the opening issue text?

In GUAC might have ingested nodes which contain references to artifact/package nodes/identifiers that the current populated GUAC database lacks metadata on - for example ingesting an SBOM with references to a PURL or github repo in which the collectors have not been configured to collect. 

In this case, we want GUAC to be able to be intentional with seeking out information from configured sources/collectors to download this information. e.g., a private guac instance querying about the public one, or private GUAC instance upon seeing a PURL, to query deps.dev to ingest additional metadata.

**Requirement**

1. We need a mechanism to be able to fetch additional metadata to ingest into GUAC. 
2. Augmenting the query protocol to handle user interactions when answering these types of queries when there is no/partial information. (e.g. provide feedback to the user on lack of data, and potentially the status of retrieving additional metadata)

mihaimaruseac:
This sounds great!

I wasn't thinking of user-in-the-loop, but actually that's also a nicer addition

lumjjb:
Wrote up a draft design for discussion

JonZeolla:
Left some feedback in the design doc

lumjjb:
Partially addressed by #306

pxp928:
Collector subscriber work completed for the beta. Ability to collect more information from git repos, OCI registry and deps.dev

PR #403, opened 2023-02-04
Add a collectsub client type Datasource for collectors

lumjjb:
Add a collectsub client type Datasource for collectors to use as part of #244 

Depends on https://github.com/guacsec/guac/pull/402

### DIFF
--- a/pkg/collectsub/datasource/csubsource/csubsource.go
+++ b/pkg/collectsub/datasource/csubsource/csubsource.go
@@ -27,7 +27,7 @@ import (
 )
 
 type csubDataSources struct {
-	c            *client.Client
+	c            client.Client
 	lastEntries  *datasource.DataSources
 	pollDuration time.Duration
 }
@@ -36,7 +36,7 @@ type csubDataSources struct {
 // from a configuration file. This configuration file is in YAML and
 // follows the structure outlined in the FileFormat struct. An example
 // is as follows:
-func NewCsubDatasource(c *client.Client, pollDuration time.Duration) (datasource.CollectSource, error) {
+func NewCsubDatasource(c client.Client, pollDuration time.Duration) (datasource.CollectSource, error) {
 	return &csubDataSources{
 		c:            c,
 		pollDuration: pollDuration,
@@ -46,7 +46,10 @@ func NewCsubDatasource(c *client.Client, pollDuration time.Duration) (datasource
 // GetDataSources returns a data source containing targets for the
 // collector to collect
 func (d *csubDataSources) GetDataSources(ctx context.Context) (*datasource.DataSources, error) {
-	entries, err := d.c.GetCollectEntries(ctx, []*pb.CollectEntryFilter{})
+	entries, err := d.c.GetCollectEntries(ctx, []*pb.CollectEntryFilter{
+		{Type: pb.CollectDataType_DATATYPE_OCI, Glob: "*"},
+		{Type: pb.CollectDataType_DATATYPE_GIT, Glob: "*"},
+	})
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/collectsub/datasource/csubsource/csubsource.go
+++ b/pkg/collectsub/datasource/csubsource/csubsource.go
@@ -1,0 +1,110 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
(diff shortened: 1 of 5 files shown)


pkg/collectsub/client/client.go
```diff
@@ -25,30 +25,36 @@ import (
...
+	Close()
+}
+
+type client struct {
 	client pb.ColectSubscriberServiceClient
 	conn   *grpc.ClientConn
 }
 
-func NewClient(addr string) (*Client, error) {
+func NewClient(addr string) (Client, error) {
```
naveensrinivasan, line 39: Is it possible to return a pointer to the `*Client` instead of the `Client` itself?

pkg/collectsub/datasource/csubsource/csubsource.go
```diff
@@ -0,0 +1,113 @@
...
+// DataSourcesUpdate will return a channel which will get an element
+// if the CollectSource has new data. Upon update, nil is inserted
+// into the channel and non-nil if the channel no longer is able to
+// serve updates.
+func (d *csubDataSources) DataSourcesUpdate(ctx context.Context) (<-chan error, error) {
+	updateChan := make(chan error)
+	go func() {
+		for {
+			select {
+			case <-time.After(d.pollDuration):
```
naveensrinivasan, line 71: I would recommend using `ticker := time.NewTicker(d.pollDuration)`
The use of a ticker instead of time.After can improve the reliability of the code in several ways:

1. A ticker is a periodic timer that repeatedly sends a value on a channel, while time.After sends a value only once after a specified duration. This means that a ticker can be used to run a task at regular intervals, while time.After would only trigger once after the specified duration.

1. The ticker's channel can be stopped or reset, providing more control over the timing of the periodic task. With time.After, once the specified duration has passed, there is no way to stop or reset the timer.

1. Tickers are more efficient than using time.After in a loop, as they use a single underlying timer. Each time time.After is used, a new timer is created, leading to increased resource usage.

In short, tickers provide a more flexible and efficient way to perform periodic tasks compared to using time.After.

pkg/collectsub/datasource/csubsource/csubsource.go
```diff
@@ -0,0 +1,113 @@
...
+			d.GitDataSources = append(d.GitDataSources, datasource.Source{
+				Value: e.Value,
+			})
+		case pb.CollectDataType_DATATYPE_OCI:
+			d.OciDataSources = append(d.OciDataSources, datasource.Source{
+				Value: e.Value,
+			})
+
+		default:
+			// unhandled datatype, skip
```
naveensrinivasan, line 109: A log of unhandled would be helpful to debug.

pkg/collectsub/datasource/filesource/filesource.go
```diff
@@ -115,6 +116,12 @@ func (d *fileDataSources) DataSourcesUpdate() (<-chan error, error) {
 				updateChan <- err
 				fmt.Printf("got err while watching: %+v\n", err)
 				return
+
+			case <-ctx.Done():
+				err := fmt.Errorf("file watcher ending from context closure")
+				updateChan <- err
+				fmt.Printf("ctx is closed: %+v\n", err)
```
naveensrinivasan, line 123: Recommend to use log instead of fmt.

pkg/collectsub/datasource/csubsource/csubsource.go
lumjjb: hmm yea that seems suitable! I'm guessing probably a ticker + do a reset after each select is completed so that the ticker doesn't get backlogged if the logic takes too long. Learn something new everyday. Thanks!

pkg/collectsub/datasource/filesource/filesource.go
lumjjb: +1 , will remove these statements since the errors are propagated and would be handled elsewhere

naveensrinivasan: 👍

pkg/collectsub/datasource/csubsource/csubsource.go
naveensrinivasan: 👍

pkg/collectsub/client/client.go
lumjjb: it is an interface now so it is already a pointer

pkg/collectsub/datasource/csubsource/csubsource_test.go
```diff
@@ -0,0 +1,172 @@
...
+		t.Errorf("unable to create FileDataSources: %v", err)
+		return
+	}
+
+	upChan, err := cds.DataSourcesUpdate(ctx)
+	if err != nil {
+		t.Errorf("unable to get DataSourcesUpdate: %v", err)
+	}
+
+	ds, err := cds.GetDataSources(ctx)
```
pxp928, line 106: is there a reason why `GetDataSources` gets called twice and compared in this test?

pxp928 (approved):
LGTM! Question about one of the tests.

pkg/collectsub/datasource/csubsource/csubsource_test.go
lumjjb: yup - it gets updated and it checks to see if it got the updated content.

mihaimaruseac (approved):

PR #404, opened 2023-02-04
Remove stack overflow error

nathannaveen:
I realized that there could be a stack overflow error in `getCompHelper()` from the file `pkg/components/root_package/root_package.go`.  The reason for this being that there is a recusive call to `getCompHelper()`. Because of this I edited `getCompHelper()` and wrote a test. After figuring out that there was a stack overflow error I created a `vistited` map to solve the issue.

I edited `getCompHelper()` to make it easier for me to test whether there would be a stack overflow error.

``` go
func getCompHelper(ctx context.Context, client graphdb.Client, parentPurl string, dependencies []interface{}, visited map[int64]bool) ([]*PackageComponent, error) {
	//dependencies, err := graphdb.ReadQuery(client, "MATCH (p:Package) WHERE p.purl = $rootPurl WITH p MATCH (p)-[:DependsOn]->(p2:Package) return p2",
	//	map[string]any{"rootPurl": parentPurl})
	//if err != nil {
	//	return nil, err
	//}
	depPackages := []*PackageComponent{}
	for _, dep := range dependencies {
		foundDep, ok := dep.(dbtype.Node)
		if !ok {
			return nil, errors.New("failed to cast to node type")
		}
		if _, ok := visited[foundDep.Id]; ok {
			continue
		}
		visited[foundDep.Id] = true
		foundDepPack := assembler.PackageNode{}
		foundDepPack.Purl, ok = foundDep.Props["purl"].(string)
		if !ok {
			return nil, errors.New("failed to cast purl property to string type")
		}
		deps, err := getCompHelper(ctx, client, foundDepPack.Purl, dependencies, visited)
		if err != nil {
			return nil, err
		}
		depPackages = append(depPackages, &PackageComponent{
			Package:     foundDepPack,
			DepPackages: deps,
		})
	}
	return depPackages, nil
}
```

``` go
func TestGetCompHelperForStackOverflow(t *testing.T) {
	ctx := context.Background()
	ctrl := gomock.NewController(t)
	driver := mocks.NewMockDriver(ctrl)
	client := driver
	parentPurl := "testPurl"
	dependencies := []interface{}{}
	// create a circular dependency to cause stack overflow
	for i := 0; i < 10000; i++ {
		dependencies = append(dependencies, dbtype.Node{
			Id:     int64(i),
			Labels: []string{"Package"},
			Props:  map[string]interface{}{"purl": parentPurl},
		})
	}

	_, err := getCompHelper(ctx, client, parentPurl, dependencies, map[int]bool{})
	if err != nil {
		t.Error(err)
	}
}
```

### DIFF
--- a/pkg/certifier/components/root_package/root_package.go
+++ b/pkg/certifier/components/root_package/root_package.go
@@ -63,7 +63,7 @@ func (q *packageQuery) GetComponents(ctx context.Context, compChan chan<- interf
 		if !ok {
 			return errors.New("failed to cast purl property to string type")
 		}
-		deps, err := getCompHelper(ctx, q.client, rootPackage.Purl)
+		deps, err := getCompHelper(ctx, q.client, rootPackage.Purl, map[int64]bool{})
 		if err != nil {
 			return err
 		}
@@ -76,7 +76,7 @@ func (q *packageQuery) GetComponents(ctx context.Context, compChan chan<- interf
 	return nil
 }
 
-func getCompHelper(ctx context.Context, client graphdb.Client, parentPurl string) ([]*PackageComponent, error) {
+func getCompHelper(ctx context.Context, client graphdb.Client, parentPurl string, visited map[int64]bool) ([]*PackageComponent, error) {
 	dependencies, err := graphdb.ReadQuery(client, "MATCH (p:Package) WHERE p.purl = $rootPurl WITH p MATCH (p)-[:DependsOn]->(p2:Package) return p2",
 		map[string]any{"rootPurl": parentPurl})
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/certifier/components/root_package/root_package.go
```diff
@@ -88,12 +88,17 @@ func getCompHelper(ctx context.Context, client graphdb.Client, parentPurl string
 		if !ok {
 			return nil, errors.New("failed to cast to node type")
 		}
+		if visited[foundDep.Id] {
+			continue
+		}
+		visited[foundDep.Id] = true
```
lumjjb, line 94: There is a very small optimization where the root is not bootstrapped in the visited map, this can be moved to the top of the function (with the parentPurl) in order to remediate that, else lgtm

knrc:
Note this PR also addresses significant performance issues, for example it reduced the certifier run for a single bom (158 components) from what had been over 15 mins down to 26 seconds.

nathannaveen:
> Note this PR also addresses significant performance issues, for example it reduced the certifier run for a single bom (158 components) from what had been over 15 mins down to 26 seconds.
> 
> Note the query count reduced from 162094 to 158 with this change

Thank You :) That is cool!

pkg/certifier/components/root_package/root_package.go
nathannaveen: @lumjjb A friendly ping

nathannaveen: Thanks, done!

pxp928 (approved):
LGTM

lumjjb (approved):
lgtm

PR #409, opened 2023-02-06
Update nats_emitter.go constants and logging

naveensrinivasan:
- Change the SubjectNameDocParsed constant to be unused
- Update BackOffTimer constant to be of type time.Duration
- Change logger.Errorf and logger.Infof calls to use %v instead of %w

[pkg/emitter/nats_emitter.go]
- Change the `SubjectNameDocParsed` constant to be unused
- Change the `BackOffTimer` constant to be of type `time.Duration`
- Change the `logger.Errorf` call to use `%v` instead of `%w`


lumjjb:
closing from stale.

PR #412, opened 2023-02-06
implement PURL parse helper

lumjjb:
implement PURL parse helper that will take in a PURL and convert it to a software tree/trie graphql model

### DIFF
--- a/pkg/assembler/helpers/purl.go
+++ b/pkg/assembler/helpers/purl.go
@@ -1,0 +1,141 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package helpers
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	purl "github.com/package-url/packageurl-go"
+
+	"github.com/guacsec/guac/pkg/assembler/graphql/model"
+)
+
+const PurlTypeGuac = "guac"
+
+// PurlToPkg converts a purl URI string into a graphql package node
+func PurlToPkg(purlUri string) (*model.Package, error) {
+	p, err := purl.FromString(purlUri)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse purl: %v", err)
+	}
+
+	return purlConvert(p)
+}
+
+func purlConvert(p purl.PackageURL) (*model.Package, error) {
+	switch p.Type {
+
+	// Enumeration of https://github.com/package-url/purl-spec#known-purl-types
+	// TODO(lumjjb): each PURL definition usually comes with a default repository
+	// we should consider addition of default repository to the prefix of the namespace
+	// so that they can be referenced with higher specificity in GUAC
+	//
+	// PURL types not defined in purl library handled generically
+	case "alpm", "apk", "huggingface", "mlflow", "qpkg", "pub", "swid":
+		fallthrough
+	// PURL types defined in purl library handled generically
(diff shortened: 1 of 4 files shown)


pkg/assembler/helpers/purl.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	purl "github.com/package-url/packageurl-go"
+
+	"github.com/guacsec/guac/pkg/assembler/graphql/model"
+)
+
+const PurlTypeGuac = "guac"
```
pxp928, line 28: should we have another helper function that takes in a string and return a generic `pkg:guac` purl. This "guac purl" can be passed in purlConvert to define the `model.Package`

pkg/assembler/helpers/purl_test.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+
+	for _, tt := range testCases {
+		t.Run(fmt.Sprintf("processing %v", tt.purlUri), func(t *testing.T) {
+			got, err := PurlToPkg(tt.purlUri)
+			if err != nil {
+				t.Errorf("unable to parse purl %v: %v", tt.purlUri, err)
+				return
+			}
+			//if !reflect.DeepEqual(got, tt.expected) {
```
pxp928, line 202: nit: remove as not needed.

pxp928 (approved):
LGTM. We should add in the `pkg:guac` usecase here as well.

mihaimaruseac (approved):

pkg/assembler/helpers/purl.go
```diff
@@ -0,0 +1,141 @@
...
+func PurlToPkg(purlUri string) (*model.Package, error) {
+	p, err := purl.FromString(purlUri)
+	if err != nil {
+		return nil, fmt.Errorf("unable to parse purl: %v", err)
+	}
+
+	return purlConvert(p)
+}
+
+func purlConvert(p purl.PackageURL) (*model.Package, error) {
```
naveensrinivasan, line 40: How about something like this?
``` go
func purlConvert(p purl.PackageURL) (*model.Package, error) {
	// Create a map of all known types and their corresponding code
	typeMap := map[string]func(purl.PackageURL) (*model.Package, error){
		purl.TypeBitbucket:   genericPkg,
		purl.TypeCocoapods:   genericPkg,
		purl.TypeCargo:       genericPkg,
		purl.TypeComposer:    genericPkg,
		purl.TypeConan:       genericPkg,
		purl.TypeConda:       genericPkg,
		purl.TypeCran:        genericPkg,
		purl.TypeDebian:      genericPkg,
		purl.TypeGem:         genericPkg,
		purl.TypeGithub:      genericPkg,
		purl.TypeGolang:      genericPkg,
		purl.TypeHackage:     genericPkg,
		purl.TypeHex:         genericPkg,
		purl.TypeMaven:       genericPkg,
		purl.TypeNPM:         genericPkg,
		purl.TypeNuget:       genericPkg,
		purl.TypePyPi:        genericPkg,
		purl.TypeRPM:         genericPkg,
		purl.TypeSwift:       genericPkg,
		purl.TypeGeneric:     genericPkg,
		"alpm":               genericPkg,
		"apk":                genericPkg,
		"huggingface":        genericPkg,
		"mlflow":             genericPkg,
		"qpkg":               genericPkg,
		"pub":                genericPkg,
		"swid":               genericPkg,
		purl.TypeOCI:         ociPkg,
		purl.TypeDocker:      dockerPkg,
	}

	// Call the appropriate function for the package type
	fn, ok := typeMap[p.Type]
	if !ok {
		return nil, fmt.Errorf("unsupported package type: %s", p.Type)
	}
	return fn(p)
}
```

The purlConvert method stores the relevant function for each type in a map of known types. If the type is not supported, it returns an error. This makes it easier to add support for new types in the future. Instead of adding a new case statement to the switch statement, you can just add an entry to the typeMap map and implement the corresponding function.

pkg/assembler/helpers/purl.go
lumjjb: oo yea this would be a good one, but i think it lands inside the parser instead of assembler

pxp928: Should we add a definition here to parse `pkg:guac`?

pkg/assembler/helpers/purl.go
lumjjb: an opportunity for a PR :)

PR #415, opened 2023-02-07
Update error context and timeout for NATS calls

naveensrinivasan:
- Remove `ctx` parameter from `GetDataFromNats` calls
- Add `fmt` and `log` imports
- Update `NewPubSub` and `GetDataFromNats` functions to return an error with more context and add a timeout for data or error on the channels

[pkg/ingestor/parser/parser.go]
- Remove the `ctx` parameter from `psub.GetDataFromNats` call [pkg/emitter/nats_data.go]
- Add `fmt` and `log` imports
- Update `NewPubSub` function to return an error with more context
- Update `GetDataFromNats` function to return an error with more context and add a timeout for data or error on the channels [pkg/certifier/certify/certify_test.go]
- Remove the `ctx` parameter from `GetDataFromNats` [pkg/handler/collector/collector_test.go]
- Remove `ctx` parameter from `psub.GetDataFromNats` function call [pkg/handler/processor/process/process.go]
- Remove the `ctx` parameter from the `GetDataFromNats` call [pkg/emitter/nats_emitter_test.go]
- Remove `ctx` from the `GetDataFromNats` call

### DIFF
--- a/pkg/certifier/certify/certify_test.go
+++ b/pkg/certifier/certify/certify_test.go
@@ -245,7 +245,7 @@ func testSubscribe(ctx context.Context, transportFunc func(processor.DocumentTre
 		return nil
 	}
 
-	err = psub.GetDataFromNats(ctx, processFunc)
+	err = psub.GetDataFromNats(processFunc)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


lumjjb:
It is valuable to have context passed into NATS as it still interfaces with an external service and the context is valuable to it - though not yet implemented.

lumjjb:
closing for now, reopen if needed

PR #417, opened 2023-02-08
Add Neo4j Backend queries

pxp928:

Adds backend resolver queries for neo4j. 

Note: Additional queries based on the fields specified will be added in another PR.


pkg/assembler/backends/neo4j/artifact.go
```diff
@@ -45,3 +50,74 @@ func (an *artifactNode) IdentifiablePropertyNames() []string {
...
+
+	result, err := session.ReadTransaction(
+		func(tx neo4j.Transaction) (interface{}, error) {
+
+			var sb strings.Builder
+			var firstMatch bool = true
+			queryValues := map[string]any{}
+
+			sb.WriteString("MATCH (n:Artifact)")
+			if artifactSpec.Algorithm != nil {
```
lumjjb, line 66: nit: i think for artifact should we mandate both be specified or not at all?

pkg/assembler/backends/neo4j/builder.go
```diff
@@ -38,3 +47,58 @@ func (bn builderNode) PropertyNames() []string {
...
+
+			sb.WriteString("MATCH (n:Builder)")
+
+			if builderSpec.URI != nil {
+
+				if firstMatch {
+					err := matchWhere(&sb, "n", "uri", "$builderUri")
+					if err != nil {
+						return nil, fmt.Errorf("string builder failed with err: %w", err)
+					}
```
lumjjb, line 70: missing setting `firstMatch=false`, i see what you're doing with having the blocks be modular, but for most cases, it looks like it isn't quite used. maybe remove it if its deterministic and then use it only for the for loops?

pkg/assembler/backends/neo4j/cve.go
```diff
@@ -137,3 +143,140 @@ func (e *cveYearToCveID) PropertyNames() []string {
...
+				queryValues["cveYear"] = cveSpec.Year
+			}
+
+			if cveSpec.CveID != nil {
+
+				if firstMatch {
+					err := matchWhere(&sb, "cveID", "id", "$cveID")
+					if err != nil {
+						return nil, fmt.Errorf("string builder failed with err: %w", err)
+					}
```
lumjjb, line 183: set firstMatch

pkg/assembler/backends/neo4j/cve.go
```diff
@@ -137,3 +143,140 @@ func (e *cveYearToCveID) PropertyNames() []string {
...
+
+			sb.WriteString("MATCH (n:Cve)-[:CveIsYear]->(cveYear:CveYear)")
+
+			if cveSpec.Year != nil {
+
+				if firstMatch {
+					err := matchWhere(&sb, "cveYear", "year", "$cveYear")
+					if err != nil {
+						return nil, fmt.Errorf("string builder failed with err: %w", err)
+					}
```
lumjjb, line 247: set firstMatch

pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -246,3 +253,210 @@ func (e *nameToVersion) PropertyNames() []string {
...
+			}
+
+			if pkgSpec.MatchOnlyEmptyQualifiers != nil && !*pkgSpec.MatchOnlyEmptyQualifiers {
+
+				if len(pkgSpec.Qualifiers) > 0 {
+
+					for _, qualifier := range pkgSpec.Qualifiers {
+						// neo4j does not accept "." in its properties. If the qualifier contains a "." that must
+						// be replaced by an "-"
+						qualifierKey := strings.ReplaceAll(qualifier.Key, ".", "_")
```
lumjjb, line 358: nit: make helper function in case we need to change this in the future.

pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -246,3 +253,210 @@ func (e *nameToVersion) PropertyNames() []string {
...
+						} else {
+							err := matchAnd(&sb, "version", qualifierKey, "$"+qualifierKey)
+							if err != nil {
+								return nil, fmt.Errorf("string builder failed with err: %w", err)
+							}
+						}
+						queryValues[qualifierKey] = qualifier.Value
+					}
+				}
+			}
```
lumjjb, line 374: should there be an else here where we want to match only on empty qualifiers?

pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -246,3 +253,210 @@ func (e *nameToVersion) PropertyNames() []string {
...
+			pkgTypes := map[string]map[string]map[string][]*model.PackageVersion{}
+			pkgNamespaces := map[string]map[string][]*model.PackageVersion{}
+			pkgNames := map[string][]*model.PackageVersion{}
+			for result.Next() {
+
+				versionNode := result.Record().Values[3].(dbtype.Node)
+
+				pkgQualifiers := []*model.PackageQualifier{}
+				if pkgSpec.MatchOnlyEmptyQualifiers != nil && !*pkgSpec.MatchOnlyEmptyQualifiers {
+					if len(pkgSpec.Qualifiers) > 0 {
```
lumjjb, line 392: I think either way these 2 paths are equivalent, perhaps we can just keep the else block?

pkg/assembler/backends/neo4j/builder.go
pxp928: so the linter errored on that saying it will never be used. If you see the execution it's true that even if firstMatch is set it won't be used after that point. `firstMatch` is used to determine if a `WHERE` is needed or if the `WHERE` exists, than use `AND` to append another condition.

pxp928: Depending on what is specified in the query. So depending on where the query start, a `WHERE` is needed, and in other cases `AND`

pxp928: ah yes you are right. Removed where its not needed.

pkg/assembler/backends/neo4j/cve.go
pxp928: lint error if added as after this point it does not matter if its set to true or false as its last time its used.

pkg/assembler/backends/neo4j/cve.go
pxp928: removed

pkg/assembler/backends/neo4j/artifact.go
mihaimaruseac: +1, we need both

pkg/assembler/backends/neo4j/pkg.go
mihaimaruseac: I think it should be covered by `MatchOnlyEmptyQualifiers`

mihaimaruseac (approved):

pkg/assembler/backends/neo4j/artifact.go
pxp928: add new error case:
```go
	if artifactSpec.Algorithm == nil && artifactSpec.Digest == nil {
		return nil, gqlerror.Errorf("must specify both algorithm and digest for artifact")
	}
```

pkg/assembler/backends/neo4j/pkg.go
lumjjb: i think the code here just omits the qualifiers, but doesn't look for nodes with no qualifiers (filtering on no qualifiers), don't think there's a good way to query this right now. Unless i'm understanding the flag incorrectly.

pxp928: @mihaimaruseac and I discussed this. We have a path forward to remedy this.

pxp928: the method we discussed might not work. Looking at alternatives.

mihaimaruseac (approved):

pxp928:
Wait. The logic for pkg and src is wrong when returning the results. I am looking into it.

pkg/assembler/backends/neo4j/pkg.go
pxp928: added helper.

pxp928:
Fixed!

pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -246,3 +274,201 @@ func (e *nameToVersion) PropertyNames() []string {
...
+				namespaceString := result.Record().Values[1].(string)
+				typeString := result.Record().Values[0].(string)
+
+				pkgVersion := &model.PackageVersion{
+					Version:    versionString,
+					Subpath:    subPathString,
+					Qualifiers: pkgQualifiers,
+				}
+
+				if _, ok := pkgTypes[typeString]; ok {
```
lumjjb, line 414: minor nit, could take the map value out the if block declaration as well

lumjjb (approved):
lgtm!

PR #421, opened 2023-02-09
update oci collectors to use datasources

lumjjb:
This PR updates the OCI collector to use a datasource, as per #244 (with an in memory datasource, not yet to a collectsub datasource).

- create inmemDataSource and change oci convention
- ocicollector uses datasources
- update oci collector consumers to use datasources


mihaimaruseac (changes requested):
You have some binary temp files that we need to remove: `pkg/collectsub/datasource/filesource/.filesource.go.swp`

pxp928 (approved):
LGTM!

mihaimaruseac (approved):

PR #422, opened 2023-02-09
Add resolvers based on selected fields

pxp928:

Add remainder resolvers based on the fields selected.


pkg/assembler/backends/testing/backend.go
```diff
@@ -69,6 +69,42 @@ func (c *demoClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*m
 	return packages, nil
 }
 
+func (c *demoClient) PackagesType(ctx context.Context, pkgSpec *model.PkgSpec) ([]*model.Package, error) {
```
mihaimaruseac, line 72: Oh, I was thinking these additional resolvers would be needed only by the neo4j backend so they would be internal there. We won't expose them to the other backends as they are just optimizations in how the neo4j receives the data

pxp928: The `queryResolver` goes to the interface so it would have to be implemented by all the backends

mihaimaruseac: Agree, but I was thinking we can make the public resolver then call private functions in the module. Kind of like what we did on ingesting:

https://github.com/guacsec/guac/blob/421309e7419590db05b2a31e5363f88f36183e29/pkg/assembler/backends/testing/ingest_pkg.go#L74-L87

Here `registerPackage` is part of the interface but `registerNamespace` not anymore

pxp928: got it making the change.

pxp928: how about this? Combining via string builder would be messy. This would be easier to modify each individual query as we need.

mihaimaruseac (approved):
Awesome

pkg/assembler/backends/neo4j/backend.go
```diff
@@ -108,3 +110,28 @@ func matchNotEdge(sb *strings.Builder, firstMatch bool, firstNodeLabel string, e
 	sb.WriteString(secondNodeLabel)
 	sb.WriteString(")")
 }
+
+func getPreloads(ctx context.Context) []string {
```
lumjjb, line 114: nit: would be helpful to add some context of this function

pkg/assembler/backends/neo4j/src.go
```diff
@@ -197,6 +197,26 @@ func (e *srcNamespaceToName) IdentifiablePropertyNames() []string {
...
+
+	// fields: [type namespaces namespaces.namespace namespaces.names namespaces.names.name namespaces.names.tag namespaces.names.commit]
+	fields := getPreloads(ctx)
+	nameRequired := false
+	namespaceRequired := false
+	for _, f := range fields {
+		if f == "namespaces" {
+			namespaceRequired = true
+		}
+		if f == "namespaces.names" {
```
lumjjb, line 209: nit: may be cleaner to make this a const that gets built, to show the structure, e.g. const ns = "...", names = ns + "." + "names", etc.

lumjjb (approved):
lgtm minor nits

PR #423, opened 2023-02-10
Included Wrapcheck linter

nathannaveen:
- Included wrapcheck linter
- Updated settings to ignore previous issues
- Fixes https://github.com/guacsec/guac/issues/371

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>

PR #424, opened 2023-02-10
Fixed error statments

nathannaveen:
- Changed the errors from returning `return err` to `return fmt.Errorf()`.
- Only changed some of the errors because it would be too big of the PR otherwise and would take forever.
- Helps with https://github.com/guacsec/guac/pull/423#discussion_r1102924221

### DIFF
--- a/cmd/guacone/cmd/files.go
+++ b/cmd/guacone/cmd/files.go
@@ -207,14 +207,14 @@ func validateFlags(user string, pass string, dbAddr string, realm string, keyPat
 
 func getProcessor(ctx context.Context) (func(*processor.Document) (processor.DocumentTree, error), error) {
 	return func(d *processor.Document) (processor.DocumentTree, error) {
-		return process.Process(ctx, d)
+		return process.Process(ctx, d) // nolint:wrapcheck
 	}, nil
 }
 func getIngestor(ctx context.Context) (func(processor.DocumentTree) ([]assembler.Graph, error), error) {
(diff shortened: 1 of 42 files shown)


cmd/guacone/cmd/files.go
```diff
@@ -207,14 +207,14 @@ func validateFlags(user string, pass string, dbAddr string, realm string, keyPat
...
-		return process.Process(ctx, d)
+		return process.Process(ctx, d) // nolint:wrapcheck
 	}, nil
 }
 func getIngestor(ctx context.Context) (func(processor.DocumentTree) ([]assembler.Graph, error), error) {
 	return func(doc processor.DocumentTree) ([]assembler.Graph, error) {
 		inputs, err := parser.ParseDocumentTree(ctx, doc)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("unable to parse document tree: %v", err)
```
lumjjb, line 217: probably should be %w for wrapping error

lumjjb (changes requested):
there seem to be a couple where the format string uses %v or %s instead of %w. can you change that please. Thanks!

cmd/guacone/cmd/files.go
nathannaveen: Good idea, I forgot about that. I have changed all the `%v`’s that I have added to `%w`’s.

nathannaveen:
Closing this PR in favor of doing smaller versions of this because I am unable to debug where the integration tests are failing.

PR #449, opened 2023-02-14
Add simple SLSA identifier strings and provide integration with ingestor

lumjjb:
Implementing the parser side of the collectsub service where found string tokens that resemble potential metadata targets get registered in the collectsub service and the collectors are able to listen for. As part of #244 

### DIFF
--- a/cmd/guacone/cmd/files.go
+++ b/cmd/guacone/cmd/files.go
@@ -212,13 +212,11 @@ func getProcessor(ctx context.Context) (func(*processor.Document) (processor.Doc
 }
 func getIngestor(ctx context.Context) (func(processor.DocumentTree) ([]assembler.Graph, error), error) {
 	return func(doc processor.DocumentTree) ([]assembler.Graph, error) {
-		inputs, idStrings, err := parser.ParseDocumentTree(ctx, doc)
+		// for guacone collectors, we do not integrate with the collectsub service
+		inputs, _, err := parser.ParseDocumentTree(ctx, doc)
 		if err != nil {
 			return nil, err
 		}
-
-		// for guacone collectors, we do not integrate with the collectsub service
-		_ = idStrings
 
 		return inputs, nil
 	}, nil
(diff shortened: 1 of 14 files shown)


pkg/ingestor/parser/parser.go
```diff
@@ -113,18 +113,30 @@ func Subscribe(ctx context.Context, transportFunc func([]assembler.Graph) error)
 }
 
 // ParseDocumentTree takes the DocumentTree and create graph inputs (nodes and edges) per document node.
-func ParseDocumentTree(ctx context.Context, docTree processor.DocumentTree) ([]assembler.Graph, error) {
+func ParseDocumentTree(ctx context.Context, docTree processor.DocumentTree) ([]assembler.Graph, []*common.IdentifierStrings, error) {
```
naveensrinivasan, line 116: I firmly believe that having more than a tuple in return for a function makes it harder to maintain. I suggest we change this into a struct.

lumjjb: Thanks for the review. I believe 2 values besides an error is reasonable. Will get input from comments from other maintainers as well.

mlieberman85 (approved):
LGTM

cmd/guacone/cmd/files.go
```diff
@@ -212,10 +212,14 @@ func getProcessor(ctx context.Context) (func(*processor.Document) (processor.Doc
 }
 func getIngestor(ctx context.Context) (func(processor.DocumentTree) ([]assembler.Graph, error), error) {
 	return func(doc processor.DocumentTree) ([]assembler.Graph, error) {
-		inputs, err := parser.ParseDocumentTree(ctx, doc)
+		inputs, idStrings, err := parser.ParseDocumentTree(ctx, doc)
```
pxp928, line 215: is this needed? If not used can remove `idStrings` and move the below comment here.

pxp928 (approved):
LGTM. Small nits

PR #453, opened 2023-02-15
Fix issue when intermediate dependencies also contain vulnerabilities

knrc:
This PR addresses the scenario when the intermediate artifact contains vulnerabilities.  The current codebase ignores these vulnerabilities when running `certifier`, for example the vertx-web vulnerability in the test case.

### DIFF
--- a/internal/testing/testdata/testdata.go
+++ b/internal/testing/testdata/testdata.go
@@ -802,6 +802,185 @@ var (
 		Package:     rootPackage,
 		DepPackages: []*root_package.PackageComponent{secondLevel, log4j},
 	}
+
+	VertxWebCommonAttestation = `{
+		"_type": "https://in-toto.io/Statement/v0.1",
+		"predicateType": "https://in-toto.io/attestation/vuln/v0.1",
+		"subject": [
+			{
+				"name": "pkg:maven/io.vertx/vertx-web-common@4.3.7?type=jar",
+				"digest": null
+			}
+		],
+		"predicate": {
+			"invocation": {
+				"uri": "guac",
+				"producer_id": "guacsec/guac"
+			},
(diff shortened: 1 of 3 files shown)


pxp928:
Thanks, @knrc that's a great catch! In our testing, we were working with custom images that would not have vulnerabilities associated so this was missed in the test case.

knrc:
@pxp928 I'm seeing test failures for a different test, TestCertifyHelperStackOverflow, which is failing with "Invalid Package URL".  I've been running tests locally but cannot reproduce, is there something in the backend environment I should know about?

pxp928:
> @pxp928 This PR is generating CI test failures for a different test, TestCertifyHelperStackOverflow, which is failing with "Invalid Package URL". I've been running tests locally but cannot reproduce, is there something in the backend environment I should know about?

So if you replace it with the following test values that will fix that issue. It was not formatted properly and did not hit the OSV endpoint so it never error before.:
```go
	A = &root_package.PackageComponent{
		Package: assembler.PackageNode{
			Purl: "pkg:example.com/A",
		},
	}

	B = &root_package.PackageComponent{
		Package: assembler.PackageNode{
			Purl: "pkg:example.com/B",
		},
	}
	C = &root_package.PackageComponent{
		Package: assembler.PackageNode{
			Purl: "pkg:example.com/C",
		},
	}
```

knrc:
Weird, now it is failing for me :)

pxp928:
> Weird, now it is failing for me :)

Right, it should fail for you locally as well. :)

pxp928 (approved):
LGTM!

mlieberman85 (approved):
LGTM

knrc:
Thanks for your help @pxp928

PR #464, opened 2023-02-17
Fixed parsing when bom contains no dependent components

knrc:


pxp928:
@lumjjb @mlieberman85 Should this merge into [heuristic-dev](https://github.com/guacsec/guac/tree/heuristic-dev)? Also, FYI @kevinconner there is a [heuristic-dev](https://github.com/guacsec/guac/tree/heuristic-dev) where we are trying to capture all the changes we need to make for SBOM parsing.

lumjjb:
Thanks @knrc for catching this!

I think this would work for the branch but also should be good with merging into main since it is not changing any behavior that is used in SETUP.md, so its not intrusive on the current work we're doing in that sense.

pxp928 (approved):
LGTM! Thanks @knrc

knrc:
@pxp928 Good to know about [heuristic-dev](https://github.com/guacsec/guac/tree/heuristic-dev), I'll take a look later today and see what is going on in that branch.

lumjjb (approved):
Thanks @knrc !

PR #473, opened 2023-02-22
Add graphql query examples

dejanb:
I thought it would be useful to collect all graph query examples from different PR into a single folder, so people can easier test them and further work on them. This is just a first step to see what others think of this idea. We can improve on them later on if this is useful. Maybe there is a better place to stored them?

### DIFF
--- a/.graphqlconfig.yaml
+++ b/.graphqlconfig.yaml
@@ -1,0 +1,4 @@
+schemaPath: ./pkg/assembler/graphql/schema/*.graphql
+extensions:
+  endpoints:
+    dev:  http://localhost:8080/query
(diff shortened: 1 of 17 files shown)


lumjjb (approved):
LGTM! Thanks @dejanb !

lumjjb:
Looks like the DCO check isn't passing, can you resolve that? https://github.com/guacsec/guac/pull/473/checks?check_run_id=11519901021

pxp928 (commented):
Thanks @dejanb. Might make sense to move this under pkg/assembler/backends? Also, there are a few missing, like source, artifact, builder, cve, osv, ghsa...etc.

dejanb:
Thanks for the response.

* Fixed DCO
* Added missing examples
* Moved them to `pkg/graphql/examples`

@pxp928 if you think `pkg/assembler/backends` is a better choice, just let me know. I just thought to see how this fit.
This could be useful later as an input for some testing. But it is also a user facing API, so it should be good to to make them easy to find. Maybe just reference from the future API docs should be good enough.

mihaimaruseac (approved):

PR #477, opened 2023-02-23
[neo4j] Move qualifiers to a list within version

mihaimaruseac:
In neo4j backend, we are storing qualfiers as a list within the version node. So we need to change test ingestion to handle this scenario.

The code to generate `qualifier_list` is duplicated, there will be future work to remove all this duplication.

### DIFF
--- a/pkg/assembler/backends/neo4j/pkg.go
+++ b/pkg/assembler/backends/neo4j/pkg.go
@@ -368,7 +368,7 @@ func (c *neo4jClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*
 			} else {
 				matchProperties(&sb, firstMatch, "version", "qualifier_list", "$qualifier")
 				firstMatch = false
-				queryValues["qualifier"] = []
+				queryValues["qualifier"] = []string{}
 			}
 
 			sb.WriteString(" RETURN type.type, namespace.namespace, name.name, version.version, version.subpath, version.qualifier_list")
--- a/pkg/assembler/backends/neo4j/pkg.go
+++ b/pkg/assembler/backends/neo4j/pkg.go
@@ -23,7 +23,6 @@ import (
 	"github.com/guacsec/guac/pkg/assembler"
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
-	"github.com/neo4j/neo4j-go-driver/v4/neo4j/dbtype"
 )
 
 // pkgNode represents the top level pkg->Type->Namespace->Name->Version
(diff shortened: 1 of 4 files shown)


pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -339,63 +315,61 @@ func (c *neo4jClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*
...
+					for _, k := range keys {
+						qualifiers = append(qualifiers, k, qualifiersMap[k])
 					}
+					matchProperties(&sb, firstMatch, "version", "qualifier_list", "$qualifier")
+					firstMatch = false
+					queryValues["qualifier"] = qualifiers
 				}
 			} else {
-				matchNotEdge(&sb, firstMatch, "version", "PkgHasQualifier", "PkgQualifier")
+				matchProperties(&sb, firstMatch, "version", "qualifier_list", "[]")
```
pxp928, line 369: should this be an empty `[]string{}` or does it not matter?

pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -406,24 +380,15 @@ func (c *neo4jClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*
...
-								Value: qualifierNode.Props[qualifierKey].(string),
-							}
-							pkgQualifiers = append(pkgQualifiers, pkgQualifier)
-						}
-					} else {
-						for key, value := range qualifierNode.Props {
-							pkgQualifier := &model.PackageQualifier{
-								Key:   key,
-								Value: value.(string),
+				if result.Record().Values[5] != nil {
```
pxp928, line 383: In #476 I make this into a helper function as it's used in more places. I can rebase and get that added to that PR.

pkg/assembler/backends/neo4j/pkg.go
pxp928:
```
matchProperties(sb, firstMatch, "depVersion", "qualifier_list", "$depQualifierList")
firstMatch = false
queryValues["depQualifierList"] = []string{}
```

pxp928 (approved):
LGTM just some questions.

pkg/assembler/backends/neo4j/pkg.go
mihaimaruseac: Shouldn't matter, but updated for consistency

pkg/assembler/backends/neo4j/pkg.go
mihaimaruseac: Yeah, it's one of the things I'm (was) planning to refactor to utilities.

lumjjb (approved):
lgtm

Issue #472, opened 2023-02-21
Add make command to check if necessary build/test tools are installed


lumjjb:
Currently the makefile uses a suite of tools such as `olangci-lint`, `protoc`, `docker`, etc. 

We should have something in the makefile that checks if these tools are installed and should provide the right error messages if they are not. This can be one to check for all like `make check-tools`, but also individual checks for specific commands - like building a container with `make container` will call `make container-tool-check` will check for the docker command only if someone decides to use it.

naveensrinivasan:
I can work on this

lumjjb:
awesome! Thanks @naveensrinivasan , assigned to you!

PR #478, opened 2023-02-23
Check if tools are installed.

naveensrinivasan:
- Add checks for docker, protoc, and golangci-lint tool
- Fixes https://github.com/guacsec/guac/issues/472

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ generated_up_to_date: generate
 
 # Run all the linters
 .PHONY: lint
-lint:
+lint: check-golangci-lint-tool-check
 	golangci-lint run ./...
 
 # Run all the tests and code checks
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -103,3 +103,22 @@ start-service:
 .PHONY: stop-service
 stop-service:
 	docker compose down
+
+# Check that docker is installed.
+.PHONY: check-docker-tool-check
+check-docker-tool-check:
+	@docker --version > /dev/null
```
lumjjb, line 110: can we make it print a note that tool X needs to be installed if it fails?

lumjjb (commented):
Minor change requested to add a statement print about why it failed, otherwise LGTM!

Makefile
naveensrinivasan: Took care of it. Thanks

lumjjb (approved):
lgtm thanks!

pxp928 (approved):
LGTM

PR #480, opened 2023-02-24
Merge SourceSpec and SourceQualifierInput

mihaimaruseac:
This makes both backends respond correctly when we want sources with no qualifiers (tags/commits) vs when we want all sources.

This makes it easier to input sources, we don't need a separate node just for querying.


pxp928 (approved):
LGTM

lumjjb (approved):

PR #481, opened 2023-02-24
Add ingestion for sources, on both backends

mihaimaruseac:


lumjjb (approved):

pkg/assembler/backends/neo4j/src.go
```diff
@@ -442,3 +442,85 @@ func (c *neo4jClient) sourcesNamespace(ctx context.Context, sourceSpec *model.So
...
+				tagStr := record.Values[4].(string)
+				tag = &tagStr
+			}
+			commit := (*string)(nil)
+			if record.Values[3] != nil {
+				// make sure to take a copy
+				commitStr := record.Values[3].(string)
+				commit = &commitStr
+			}
+			nameStr := record.Values[2].(string)
```
pxp928, line 500: should we check if `record.Values[2] != nil`? I understand it should never but worth checking.

pkg/assembler/backends/neo4j/src.go
```diff
@@ -442,3 +442,85 @@ func (c *neo4jClient) sourcesNamespace(ctx context.Context, sourceSpec *model.So
...
+			name := &model.SourceName{
+				Name:   nameStr,
+				Tag:    tag,
+				Commit: commit,
+			}
+
+			namespaceStr := record.Values[1].(string)
```
pxp928, line 507: same as above

pkg/assembler/backends/neo4j/src.go
```diff
@@ -442,3 +442,85 @@ func (c *neo4jClient) sourcesNamespace(ctx context.Context, sourceSpec *model.So
...
+			namespace := &model.SourceNamespace{
+				Namespace: namespaceStr,
+				Names:     []*model.SourceName{name},
+			}
+
+			srcType := record.Values[0].(string)
```
pxp928, line 513: same as above

pxp928 (approved):
LGTM quick comment

PR #482, opened 2023-02-24
Refactored Redundencys in gcs

nathannaveen:
- In the function `NewGCSClient` the function `getBucketPath()` was called multiple times without any difference. Now the code has one bucket that is being reused.
- In the function `RetrieveArtifacts` there was a part that was being called a couple times without any difference, now it is put into a seperate function.
- In the function `getArtifacts` there were two if statments with the same content inside them, so now they are combined.

### DIFF
--- a/pkg/handler/collector/gcs/gcs_test.go
+++ b/pkg/handler/collector/gcs/gcs_test.go
@@ -78,6 +78,16 @@ func TestGCS_RetrieveArtifacts(t *testing.T) {
 		wantErr:  false,
 		wantDone: true,
 	}, {
+		name: "last download time the same",
+		fields: fields{
+			bucket:       getBucketPath(),
+			reader:       &reader{client: client, bucket: getBucketPath()},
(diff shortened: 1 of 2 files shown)


pkg/handler/collector/gcs/gcs_test.go
```diff
@@ -77,16 +77,6 @@ func TestGCS_RetrieveArtifacts(t *testing.T) {
 		want:     doc,
 		wantErr:  false,
 		wantDone: true,
-	}, {
-		name: "last download time the same",
```
pxp928, line 81: this test case passes before the changes. Just need to add in the check again that the payload has a length greater than zero.

nathannaveen: Thanks, somehow I didn't realize that these weren't failing before the changes. I have updated the code!

pxp928 (approved):
LGTM

lumjjb (approved):

PR #484, opened 2023-02-24
Ensure artifacts are treated the same on both backends:

mihaimaruseac:
- [x] force canonicalization to lower case
- [x] ensure we can query all artifacts instead of returning error

Also did minor refactoring to code:

- [x] move query formation outside of Neo4j transaction
- [x] separate matching algorithm/digest in in-memory backend to only have a single `append` at the end

### DIFF
--- a/pkg/assembler/backends/neo4j/artifact.go
+++ b/pkg/assembler/backends/neo4j/artifact.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
-	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
 // ArtifactNode is a node that represents an artifact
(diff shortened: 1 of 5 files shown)


pkg/assembler/backends/neo4j/artifact.go
```diff
@@ -55,30 +54,26 @@ func (c *neo4jClient) Artifacts(ctx context.Context, artifactSpec *model.Artifac
...
 
-			var sb strings.Builder
-			var firstMatch bool = true
-			queryValues := map[string]any{}
-
-			sb.WriteString("MATCH (n:Artifact)")
+	if artifactSpec.Algorithm != nil {
+		matchProperties(&sb, firstMatch, "n", "algorithm", "$artifactAlgo")
+		firstMatch = false
+		queryValues["artifactAlgo"] = strings.ToLower(*artifactSpec.Algorithm)
```
pxp928, line 66: Shouldn't the ingest be enforcing the lowercase? Reason why I didn't add it to the query

mihaimaruseac: Ingest does, but we also want the query side to do it too.

```gql
query ArtQ5 {
  artifacts(artifactSpec: {digest: "7A8F47318E4676DACB0142AFA0B83029CD7BEFD9"}) {
    ...allArtifactTree
  }
}
```

would return nothing otherwise. We need to handle it here otherwise in each place we would run the query we'd need to convert o lower case.

pxp928: oh wait disregard.

pxp928 (approved):

lumjjb (approved):

PR #486, opened 2023-02-24
Added checks for isDependency and certifyPkg

pxp928:
- Added check if version is specified only query at pkg version level and not at pkg name level
- Removed unneeded nil value checks
- Removed unneeded `pkg.MatchOnlyEmptyQualifiers != nil` as it will default to false
- Removed bad query tests for isDependency (as dependent package would not have version)

### DIFF
--- a/pkg/assembler/backends/neo4j/pkg.go
+++ b/pkg/assembler/backends/neo4j/pkg.go
@@ -315,7 +315,7 @@ func (c *neo4jClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*
 				queryValues["pkgSubpath"] = pkgSpec.Subpath
 			}
 
-			if pkgSpec.MatchOnlyEmptyQualifiers != nil && !*pkgSpec.MatchOnlyEmptyQualifiers {
+			if !*pkgSpec.MatchOnlyEmptyQualifiers {
 				if len(pkgSpec.Qualifiers) > 0 {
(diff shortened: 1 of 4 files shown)


pkg/assembler/backends/neo4j/pkg.go
```diff
@@ -315,7 +315,7 @@ func (c *neo4jClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]*
 				queryValues["pkgSubpath"] = pkgSpec.Subpath
 			}
 
-			if pkgSpec.MatchOnlyEmptyQualifiers != nil && !*pkgSpec.MatchOnlyEmptyQualifiers {
+			if !*pkgSpec.MatchOnlyEmptyQualifiers {
```
lumjjb, line 318: ah ok so we assume anything with a ! will definitely be non nil yea

lumjjb (approved):

pkg/assembler/backends/neo4j/pkg.go
pxp928: In this case the `MatchOnlyEmptyQualifiers` defaults to:
```
 matchOnlyEmptyQualifiers: Boolean = false
 ```
 at the schema level. So it always default to `false`

lumjjb: 👍

mihaimaruseac (approved):

PR #487, opened 2023-02-24
enforce lowercase for vulnerabilities on both backends

pxp928:


mihaimaruseac (approved):
Nice.

Was just getting ready to work on these.

You'll likely need to regenerate the schema though

pxp928:
good catch! Added

lumjjb (approved):

PR #495, opened 2023-02-27
Fix missing attestation node

pxp928:
- moved query out of transaction
- added missing attestation nodes and edges for existing queries
- fixed cve-ghsa-osv example data


pxp928:
NOTE:

CertifyPkg now has a union of at most 8 queries due to the bi-directional nature. Based on testing we might change this in the future.

mihaimaruseac (approved):

pxp928:
Is there a better optimization for `CertifyPkg`?

lumjjb (approved):

PR #498, opened 2023-02-28
Add Github Collector

mlieberman85:
This adds a Github collector along with helpers including a Github wrapper client that should make it easier to mock.


pkg/collectsub/datasource/datasource.go
```diff
@@ -36,7 +36,17 @@ type CollectSource interface {
...
+	// https://spdx.github.io/spdx-spec/v2.3/package-information/#771-description
+	// <vcs_tool>+<transport>://<host_name>[/<path_to_repository>][@<revision_tag_or_branch>][#<sub_path>]
+	// e.g. git+https://github.com/guacsec/guac@v1
 	GitDataSources []Source
+	// NOTE: It is expected that a GithubReleaseDataSource is of the form:
+	// https://github.com/<org>/<repo>/releases/<tag> or
+	// https://github.com/<org>/<repo>/releases/tag/<tag> or
+	// https://github.com/<org>/<repo>/releases
+	// Tag is optional and left off will assume latest.
+	GithubReleaseDataSources []Source
```
lumjjb, line 49: note: we need to update the csub interface and proto to support this. Can you add a TODO with a link to an issue and assign it to me?

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+type Tag struct {
+	Value string
+}
+
+type Latest struct{}
+
+func (t *Tag) IsTagOrLatest()    {}
+func (l *Latest) IsTagOrLatest() {}
```
lumjjb, line 51: hmm, i think prefered golang way maybe is something like:

```
type TagOrLatest struct{
    Latest bool
    Tag string
}
```

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,327 @@
...
+// RetrieveArtifacts get the artifacts from the collector source based on polling or one time
+func (g *githubCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	err := g.populateRepoToReleaseTags(ctx)
+	if err != nil {
+		return err
+	}
+	if g.poll {
+		for repo, tags := range g.repoToReleaseTags {
+			g.fetchAssets(ctx, repo.Owner, repo.Repo, tags, docChannel)
+			time.Sleep(g.interval)
```
lumjjb, line 143: should the poll be between fetches of releases? if not the poll time scales with the number of tags?

internal/client/githubclient/githubclient_test.go
```diff
@@ -0,0 +1,357 @@
...
+			wantErr:    true,
+			errMessage: errors.New("GET https://api.github.com/users: 401 Bad credentials []"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.Background()
+			_, err := NewGithubClient(ctx, tt.args.token)
+			if err != nil {
+				if !tt.wantErr {
```
lumjjb, line 94: this check should probably be outside, it should also error if it expects and error and doesn't get it

internal/client/githubclient/githubclient_test.go
```diff
@@ -0,0 +1,357 @@
...
+			want:       nil,
+			wantErr:    true,
+			errMessage: errors.New("GET https://api.github.com/repos/guacsec/guac-data/releases/latest: 404 Not Found []"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := gc.GetLatestRelease(tt.args.ctx, tt.args.owner, tt.args.repo)
+			if err != nil {
+				if !tt.wantErr {
```
lumjjb, line 156: same comment as above

internal/client/githubclient/githubclient_test.go
```diff
@@ -0,0 +1,357 @@
...
+					t.Errorf("githubClient.GetLatestRelease() error = %v, wantErr %v", err, tt.wantErr)
+					return
+				}
+				if !strings.Contains(err.Error(), tt.errMessage.Error()) {
+					t.Errorf("githubClient.GetLatestRelease() error = %v, wantErr %v", err, tt.errMessage)
+					return
+				}
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
lumjjb, line 165: should only check if err == nil

internal/client/repo_client.go
```diff
@@ -0,0 +1,46 @@
+//
```
lumjjb, line 1: i think this can be under `pkg/handler/collector/github` subpath

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,327 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	if u.Scheme != "https" {
+		return nil, nil, fmt.Errorf("invalid github url scheme: %v", u.Scheme)
+	}
+	if u.Host != "github.com" {
+		return nil, nil, fmt.Errorf("invalid github host: %v", u.Host)
+	}
+	// [1:] to ignore leading slash
```
lumjjb, line 260: it would be helpful to have a comment here to define what the different array indices refer to or the cases you think are valid returns here, it'll make it easier to compare to documentation if something decides to change upstream

pkg/collectsub/datasource/datasource.go
mlieberman85: Sure.

pkg/handler/collector/github/github.go
mlieberman85: Sure.

internal/client/repo_client.go
mlieberman85: Well, once we support Gitlab as well as some of the other things I don't think that makes sense. These are internal wrappers around Github's stuff.

pkg/handler/collector/github/github.go
mlieberman85: I am still changing it since it's "idiomatic"

internal/client/githubclient/githubclient_test.go
```diff
@@ -0,0 +1,383 @@
...
+				if !tt.wantErr {
+					t.Errorf("NewGithubClient() error = %v, wantErr %v", err, tt.wantErr)
+					return
+				}
+				if !strings.Contains(err.Error(), tt.errMessage.Error()) {
+					t.Errorf("NewGithubClient() error = %v, wantErr %v", err, tt.errMessage)
+					return
+				}
+			} else {
+				if tt.wantErr {
```
lumjjb, line 103: nit: at the start, we can probably do `if err != nil != tt.wantErr {`

lumjjb (approved):

pkg/handler/collector/github/github_test.go
```diff
@@ -0,0 +1,809 @@
...
+		repo       string
+		tags       []TagOrLatest
+		docChannel chan<- *processor.Document
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+	}{
+		// TODO: Add test cases.
```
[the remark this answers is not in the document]
mlieberman85: oops I thought I had actually removed this test, since all the test cases are being handled by RetrieveArtifacts

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,336 @@
...
+				logger.Warnf("unable to download asset: %w", err)
+				continue
+			}
+			doc := &processor.Document{
+				Blob:   content.Bytes,
+				Type:   processor.DocumentUnknown,
+				Format: processor.FormatUnknown,
+				SourceInformation: processor.SourceInformation{
+					Collector: GithubCollector,
+					Source:    content.Name,
```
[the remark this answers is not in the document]
mlieberman85: Yep

internal/client/githubclient/githubclient_test.go
```diff
@@ -0,0 +1,383 @@
...
+		panic(err)
+	}
+
+	return g
+}
+
+func TestNewGithubClient(t *testing.T) {
+	token, err := getGithubTokenFromEnv()
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
mlieberman85: I'm not actually sure what to do in this case as it's not part of the test case, it's an error at the test caller level, e.g. if a token isn't passed in it's not a failure of the test but of the test caller.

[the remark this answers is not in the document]
mihaimaruseac: I think this should panic

internal/client/githubclient/githubclient.go
```diff
@@ -0,0 +1,161 @@
...
+	"io"
+	"net/http"
+
+	"github.com/google/go-github/github"
+	"github.com/guacsec/guac/internal/client"
+	"golang.org/x/oauth2"
+)
+
+// Wrapper interface for Github
+type GithubClient interface {
```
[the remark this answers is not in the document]
mlieberman85: Yep. I was going to wait until we needed to for gitlab or similar.

internal/client/githubclient/githubclient.go
```diff
@@ -0,0 +1,161 @@
...
+	// GetReleaseAsset fetches the content of a release asset, e.g. artifacts, metadata documents, etc.
+	GetReleaseAsset(asset client.ReleaseAsset) (*client.ReleaseAssetContent, error)
+}
+
+type githubClient struct {
+	ghClient   *github.Client
+	httpClient *http.Client
+}
+
+var _ GithubClient = &githubClient{}
```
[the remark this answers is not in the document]
mlieberman85: This is just short hand to ensure that `githubClient` in fact implements the `GithubClient` interface as it will fail to compile if something changes.

pkg/handler/collector/github/github.go
```diff
@@ -0,0 +1,336 @@
...
+	g := &githubCollector{
+		poll:              false,
+		interval:          0,
+		client:            nil,
+		repoToReleaseTags: map[client.Repo][]TagOrLatest{},
+		assetSuffixes:     defaultAssetSuffixes(),
+		collectDataSource: nil,
+	}
+
+	for _, opt := range opts {
```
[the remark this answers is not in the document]
mlieberman85: Thank @sudo-bmitch for showing me how to use it.

internal/client/githubclient/githubclient.go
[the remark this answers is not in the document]
lumjjb: yea this is good but at the same time, i'd be good to just slap a TODO on since if gitlab functionality gets implemented, moving this out will probably be part of that whole refactor, as of now, its simpler to understand.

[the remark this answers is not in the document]
mlieberman85: Sure, I have the TODO in the repo_client where it would most likely live but yeah probably makes sense to have one here too

cmd/collector/cmd/github.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// else direct CLI call, no polling
+	opts.poll = false
+	if len(args) < 1 {
+		return opts, fmt.Errorf("expected positional argument for release_url")
+	}
+
+	sources := []datasource.Source{}
+	for _, arg := range args {
+		if _, _, err := github.ParseGithubReleaseDataSource(datasource.Source{}); err != nil {
```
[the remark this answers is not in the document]
mlieberman85: Oh, this is incorrect. Will fix.

PR #499, opened 2023-02-28
Use RFC 3339 time format in the API

dejanb:
I think it'd be good to define that user facing APIs return timestamps in RFC 3339 format. I stumbled upon this trying to parse payload in Rust which didn't like default Go "monotonic" time format. RFC 3339 is pretty much recommended for all API use cases.

The current commit is only to show how the change would look like. The default support in gqlgen is using RFC3339Nano format which works but might be a bit overkill. Anyway, it's a good first step for now and we can discuss if we should do something else.

If everyone is supportive of this change, I would go and continue to implement and document this for the whole API.

### DIFF
--- a/pkg/assembler/graphql/schema/certifyVuln.graphql
+++ b/pkg/assembler/graphql/schema/certifyVuln.graphql
@@ -33,7 +33,7 @@ collector (property) - the GUAC collector that collected the document that gener
 type CertifyVuln {
   package: Package!
   vulnerability: OsvCveGhsaObject!
-  timeScanned: String!
+  timeScanned: Time!
   dbUri: String!
   dbVersion: String!
   scannerUri: String!
@@ -53,7 +53,7 @@ input CertifyVulnSpec {
   osv: OSVSpec
   cve: CVESpec
   ghsa: GHSASpec
-  timeScanned: String
+  timeScanned: Time
   dbUri: String
   dbVersion: String
   scannerUri: String
@@ -71,3 +71,5 @@ extend type Query {
   "Returns all CertifyVuln"
   CertifyVuln(certifyVulnSpec: CertifyVulnSpec): [CertifyVuln!]!
 }
+
+scalar Time
--- a/pkg/assembler/backends/neo4j/certifyScorecard.go
+++ b/pkg/assembler/backends/neo4j/certifyScorecard.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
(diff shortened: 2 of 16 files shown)


lumjjb:
this sounds good to me, but will defer to @mihaimaruseac and @pxp928 who are currently actively working on the GraphQL portions.

pkg/assembler/graphql/schema/certifyVuln.graphql
```diff
@@ -71,3 +71,5 @@ extend type Query {
   "Returns all CertifyVuln"
   CertifyVuln(certifyVulnSpec: CertifyVulnSpec): [CertifyVuln!]!
 }
+
+scalar Time
```
pxp928, line 75: So there was discussion on using `scalar Time` but we backtracked the change for all the timestamps as we were not sure what the format should look like at this point. We were planning to refine it once we were done with the initial implementation.

mihaimaruseac: +1, we were not certain if we were ready to specify the format of this scalar (like `scalar Time @specifiedBy(....)`)

If I understand correctly, Go also uses RFC3339Nano, so at least on that side we won't have issues.

dejanb: Thanks for the feedback. I understand why delaying this decision for later made sense.

The only issue I have is that a default formatting is a bit non-standard and hard to parse outside Go ecosystem. For example:

```
Default: 2023-03-01 10:36:31.630137 +0100 CET m=+0.006376126 // This is how it looks now
RFC3339Nano: 2023-03-01T10:36:31.630137+01:00 // This is how it looks with Time scalar
RFC3339: 2023-03-01T10:36:31+01:00 // IMHO this would be the ideal format
```

I can work for now with my branch to make response parsable with RFC3339Nano. Whenever we decide on the final format, I can pick this up and finish it.

jchestershopify: Second-level resolution means a lot of events will be bucketed that it would be useful to distinguish. Given a choice I'd prefer the nano variant.

mihaimaruseac: Let's go with the Nano variant.

I didn't realize that the formatting gets bad when I was suggesting to use a string instead. It slipped my mind.

dejanb: Great. I agree with @jchestershopify, Nano will make sense in a high-volume system. I'll continue then to change the rest of the API.

dejanb: @mihaimaruseac @pxp928 The PR should now have a fix for the whole API. Let me know if you see anything missing.

mihaimaruseac (approved):
Thank you for the PR!

pxp928 (approved):
LGTM Thanks!

PR #503, opened 2023-03-01
[neo4j] Added neo4j implementation of IsVulnerability

pxp928:

### DIFF
--- a/pkg/assembler/backends/neo4j/isVulnerability.go
+++ b/pkg/assembler/backends/neo4j/isVulnerability.go
@@ -17,11 +17,187 @@ package neo4jBackend
 
 import (
 	"context"
-	"fmt"
+	"strings"
 
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
+	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
+	"github.com/neo4j/neo4j-go-driver/v4/neo4j/dbtype"
+	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/assembler/backends/neo4j/isVulnerability.go
```diff
@@ -17,11 +17,187 @@ package neo4jBackend
...
+
+	if queryAll || isVulnerabilitySpec.Cve != nil {
+		var sb strings.Builder
+		var firstMatch bool = true
+		queryValues := map[string]any{}
+
+		// query CVE
+		query := "MATCH (root:Osv)-[:OsvHasID]->(osvID:OsvID)" +
+			"-[:subject]-(isVulnerability:IsVulnerability)-[:alias]-(cveID:CveID)<-[:CveHasID]" +
+			"-(cveYear:CveYear)<-[:CveIsYear]-(rootCve:Cve)"
```
lumjjb, line 50: should edges to subject/objects be directed? else lgtm

lumjjb (approved):

pkg/assembler/backends/neo4j/isVulnerability.go
mihaimaruseac: On query it's ok to not be directed if we don't have yet decided which direction to go. But for ingestion they have to be directed.

mihaimaruseac (approved):

pkg/assembler/backends/neo4j/isVulnerability.go
pxp928: Yes, this captures both directions but we can make it directed if needed in the future.

PR #513, opened 2023-03-01
Ingest scorecard certifications in neo4j and in-memory backends

mihaimaruseac:
Have several TODOs left for later analysis when we're going to look at performance. Also a few TODOs to clean up code and be more DRY. But pushing this in since we want to have the new ingestion pipeline working for the rest of GUAC.


pkg/assembler/backends/neo4j/certifyScorecard.go
```diff
@@ -203,3 +204,135 @@ func setCertifyScorecardValues(sb *strings.Builder, certifyScorecardSpec *model.
...
+
+	if source.Commit != nil {
+		values["commit"] = *source.Commit
+	}
+
+	if source.Tag != nil {
+		values["tag"] = *source.Tag
+	}
+
+	values["timeScanned"] = scorecard.TimeScanned
```
pxp928, line 231: can use the constant defied above for these.

pxp928 (approved):
LGTM just minor comment

pkg/assembler/backends/neo4j/certifyScorecard.go
mihaimaruseac: Changed to use constants

lumjjb (approved):

PR #514, opened 2023-03-01
Fix matching on tag and commit for sources in in-memory backend

mihaimaruseac:
We need to make sure that nil matches with everything whereas empty string matches only with nil or empty string (and any other string matches exactly).

Tested this for source as that surfaced during scorecards certification, but we might have this issue in other places too. Probably will need to do some fuzzing at some point soon.


mihaimaruseac:
This is a bad rebase, will reopen

PR #520, opened 2023-03-02
Remove re-used code from neo4j backend

pxp928:

### DIFF
--- a/pkg/assembler/backends/neo4j/src.go
+++ b/pkg/assembler/backends/neo4j/src.go
@@ -452,37 +452,13 @@ RETURN type.type, ns.namespace, name.name, name.commit, name.tag`
 				return nil, err
 			}
 
-			// TODO(mihaimaruseac): Extract this to a utility since it is repeated
-			tag := (*string)(nil)
-			if record.Values[4] != nil {
-				// make sure to take a copy
-				tagStr := record.Values[4].(string)
-				tag = &tagStr
-			}
-			commit := (*string)(nil)
-			if record.Values[3] != nil {
-				// make sure to take a copy
-				commitStr := record.Values[3].(string)
-				commit = &commitStr
-			}
+			tag := record.Values[4]
+			commit := record.Values[3]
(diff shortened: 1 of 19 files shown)


pkg/assembler/backends/neo4j/src.go
```diff
@@ -549,3 +525,34 @@ func setSrcMatchValues(sb *strings.Builder, src *model.SourceSpec, objectSrc boo
 		}
 	}
 }
+
+func generateModelSource(srcType, namespaceStr, nameStr string, commitValue, tagValue interface{}) model.Source {
```
mihaimaruseac, line 529: Should we return pointers instead of copying the struct? Some of these are quite large, so we might get slowdown if we keep copying them on functions calls and returns.

pkg/assembler/backends/neo4j/src.go
```diff
@@ -549,3 +525,34 @@ func setSrcMatchValues(sb *strings.Builder, src *model.SourceSpec, objectSrc boo
...
+	if tagValue != nil {
+		// make sure to take a copy
+		tagStr := tagValue.(string)
+		tag = &tagStr
+	}
+	commit := (*string)(nil)
+	if commitValue != nil {
+		// make sure to take a copy
+		commitStr := commitValue.(string)
+		commit = &commitStr
```
mihaimaruseac, line 540: No longer need the copy now, it's copied in the method call

mihaimaruseac (approved):
LGTM, though there might be performance issues due to multiple copies

pkg/assembler/backends/neo4j/src.go
pxp928: Would still need to do this to pass it by reference

pkg/assembler/backends/neo4j/src.go
pxp928: Done.

lumjjb (approved):

PR #521, opened 2023-03-02
Add scorecard parser integration

lumjjb:
This PR implements scorecard with the new interface:
- Define fields of placeholder (another PR to rename Placeholder to another name right after for ease of review)
- Add parsing to predicates for scorecards
- Changing of DocumentParser interface to take `*Placeholder` instead of `[]assembler.Placeholder`
- Unfreeze test for scorecards parser

### DIFF
--- a/pkg/ingestor/parser/scorecard/parser_scorecard.go
+++ b/pkg/ingestor/parser/scorecard/parser_scorecard.go
@@ -22,23 +22,22 @@ import (
 	"strings"
 
 	"github.com/guacsec/guac/pkg/assembler"
+	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/guacsec/guac/pkg/handler/processor"
 	"github.com/guacsec/guac/pkg/ingestor/parser/common"
 	sc "github.com/ossf/scorecard/v4/pkg"
 )
 
 type scorecardParser struct {
-	scorecardNodes []assembler.MetadataNode
-	// artifactNode should have a 1:1 mapping to the index
-	// of scorecardNodes.
-	artifactNodes []assembler.ArtifactNode
+	scorecardPredicates []*model.ScorecardInputSpec
+	srcPredicates       []*model.SourceInputSpec
 }
 
 // NewSLSAParser initializes the slsaParser
 func NewScorecardParser() common.DocumentParser {
 	return &scorecardParser{
-		scorecardNodes: []assembler.MetadataNode{},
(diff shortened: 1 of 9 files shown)


lumjjb:
need to change this due to conflicting changes from rebasing https://github.com/guacsec/guac/pull/499, hold on review

pkg/ingestor/parser/scorecard/parser_scorecard.go
```diff
@@ -55,95 +54,68 @@ func (p *scorecardParser) Parse(ctx context.Context, doc *processor.Document) er
...
-	return mnNode
-}
-
-func getArtifactNode(s *sc.JSONScorecardResultV2) assembler.ArtifactNode {
-	return assembler.ArtifactNode{
-		Name:   sourceUri(s.Repo.Name),
-		Digest: hashToDigest(s.Repo.Commit),
+	srcInput := model.SourceInputSpec{
+		// assuming scorecards is only git
+		Type:      "git",
```
mihaimaruseac, line 99: Can we get it from `s.Repo.Name`?

mihaimaruseac (approved):
LGTM in principle, though I think we might need to change interfaces. But if this lands before my PR for the GQL client, I'll change the types accordingly.

lumjjb:
Ref: https://github.com/ossf/scorecard/issues/2711

pkg/ingestor/parser/scorecard/parser_scorecard.go
lumjjb: hmm you mean like parsing and saying "github.com" and "gitlab.com" is git and "www.mercurial-scm.org" is "hg"?

mihaimaruseac: OH, I thought `s.Repo` contains a full URL like `git@github.com:guacsec/guac`, but if it is not then nothing to do

pkg/ingestor/parser/scorecard/parser_scorecard.go
```diff
@@ -20,25 +20,25 @@ import (
 	"encoding/json"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/guacsec/guac/pkg/assembler"
+	"github.com/guacsec/guac/pkg/assembler/graphql/model"
```
mihaimaruseac, line 26: Will need to change import to be the same as in #522, but the structs would stay similar (just in different package)

pxp928 (approved):
LGTM

PR #523, opened 2023-03-02
add gql-server cmd to guacone

lumjjb:
Added `guacone gql-server` command with the following flags:
- gql-backend: backend used for graphql api server: [neo4j | inmem | testing]
- gql-port: port used for graphql api server")
- gql-debug: debug flag which enables the graphQL playground

example comand::
```
$ bin/guacone --gql-backend inmem  gql-server --gql-debug
{"level":"info","ts":1677788627.2500522,"caller":"cmd/root.go:106","msg":"Using config file: /Users/lumb/go/src/github.com/guacsec/guac/guac.yaml"}
{"level":"info","ts":1677788627.250664,"caller":"cmd/graphql_server.go:84","msg":"graphql server running with inmem backend at http://localhost:8080/query"}
{"level":"info","ts":1677788627.2508001,"caller":"cmd/graphql_server.go:87","msg":"connect to http://localhost:8080/ for GraphQL playground"}
```

With testing db
```
$ bin/guacone --gql-backend testing  gql-server --gql-debug
{"level":"info","ts":1677788667.970688,"caller":"cmd/root.go:106","msg":"Using config file: /Users/lumb/go/src/github.com/guacsec/guac/guac.yaml"}
{"level":"info","ts":1677788667.9719412,"caller":"cmd/graphql_server.go:84","msg":"graphql server running with testing backend at http://localhost:8080/query"}
{"level":"info","ts":1677788667.972001,"caller":"cmd/graphql_server.go:87","msg":"connect to http://localhost:8080/ for GraphQL playground"}
{"level":"info","ts":1677788667.972009,"caller":"cmd/graphql_server.go:90","msg":"using testing backend, test data will be bootstrapped in the API"}
```

### DIFF
--- a/cmd/guacone/cmd/graphql_server.go
+++ b/cmd/guacone/cmd/graphql_server.go
@@ -1,0 +1,169 @@
+//
+// Copyright 2022 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"os"
+
+	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/99designs/gqlgen/graphql/playground"
+	neo4j "github.com/guacsec/guac/pkg/assembler/backends/neo4j"
+	"github.com/guacsec/guac/pkg/assembler/backends/testing"
+	"github.com/guacsec/guac/pkg/assembler/graphql/generated"
+	"github.com/guacsec/guac/pkg/assembler/graphql/resolvers"
+	"github.com/guacsec/guac/pkg/logging"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


cmd/guacone/cmd/graphql_server.go
```diff
@@ -0,0 +1,169 @@
...
+		topResolver = resolvers.Resolver{Backend: backend}
+
+	case gqlBackendInmem:
+		args := testing.DemoCredentials{}
+		backend, err := testing.GetEmptyBackend(&args)
+		if err != nil {
+			return nil, fmt.Errorf("Error creating inmem backend: %w", err)
+		}
+
+		topResolver = resolvers.Resolver{Backend: backend}
```
mihaimaruseac, line 156: I'm thinking we should merge these and have a test flag that populates the data in either inmem/neo4j backends.

Or maybe we should punt this for later when all ingestion is pushed to be done via graphql instead of directly?

mihaimaruseac: But in that case, neo4j is not populated with testing data, we'd still need a flag for that.

mihaimaruseac (approved):
LGTM in general, we just need to decide if we want the test data present from guacone or not (and then how to pass it in)

cmd/guacone/cmd/graphql_server.go
lumjjb: yea - i feel like it wouldn't quite be needed. i'm more than happy to remove "testing" backend here. wdyt?

mihaimaruseac: Yeah, let's remove it, this way guacone is only production/real data

pxp928 (approved):
LGTM, agree that we dont want to pass in fake data into guacone.

cmd/guacone/cmd/graphql_server.go
lumjjb: removed

PR #526, opened 2023-03-03
added ingest isDependency for neo4j

pxp928:
- added `graphql_playground` to make file
- commented out neo4j testdata registration (will remove in another PR)
- generated code to ingest `isDependency` on client side
- backend implementation of `isDependency`
- testdata added to cmd
- tested query against `isDependency` for functionality check

### DIFF
--- a/pkg/assembler/backends/neo4j/backend.go
+++ b/pkg/assembler/backends/neo4j/backend.go
@@ -59,7 +59,7 @@ func GetBackend(args backends.BackendArgs) (backends.Backend, error) {
 		return nil, err
 	}
 	client := &neo4jClient{driver}
-	if config.TestData {
+	/* if config.TestData {
 		err = registerAllPackages(client)
 		if err != nil {
 			return nil, err
@@ -88,7 +88,7 @@ func GetBackend(args backends.BackendArgs) (backends.Backend, error) {
 		if err != nil {
 			return nil, err
 		}
-	}
+	} */
 	return client, nil
 }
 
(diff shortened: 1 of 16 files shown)


pkg/assembler/backends/neo4j/backend.go
```diff
@@ -59,7 +59,7 @@ func GetBackend(args backends.BackendArgs) (backends.Backend, error) {
 		return nil, err
 	}
 	client := &neo4jClient{driver}
-	if config.TestData {
+	/* if config.TestData {
```
lumjjb, line 62: @mihaimaruseac still using this? or we okay to comment it out to just what's been implemented?

lumjjb (approved):

pkg/assembler/backends/neo4j/backend.go
pxp928: I will make a PR to clean this all up, just wanted to move all the testdata over to `cmd/graphql_playground/cmd/ingest.go` before we did.

lumjjb: >This conflicts with the cmd/graphql_playground/cmd/ingest.go where more testdata is being added via the proper ingestion method. The commented-out code uses the old storeGraph that will all be removed.

I added a `GetEmptyBackend` function that should work? maybe?

pxp928: Add another `GetEmptyBackend` for neo4j? It would never be used. We are not losing anything by commenting out the above. It can be removed now as the data being ingested is not in the proper format via `storeGraph `

mihaimaruseac: I'd say we still need this for a while since we haven't ingested packages with the client queries.

However, if you get an error on ingesting from this one, then let's remove

mihaimaruseac (approved):
LGTM

pkg/assembler/backends/neo4j/backend.go
pxp928: Keep it commented out for now or remove it? Its not causing an error, its just bad data being added on top off the good test data.

mihaimaruseac: Let's keep it commented, I might need the old ingested data for testing

PR #536, opened 2023-03-04
Added ingest Vulnerability for both backends and cmd test data

pxp928:
- fix wrong variable name for `IsOccurrenceSpecInputSpec` to `IsOccurrenceInputSpec`
- fix wrong edge name on `certifyVuln`
- added ingest Vulnerability for testing and neo4j backends
- added ingestion via cmd


cmd/graphql_playground/cmd/ingest.go
```diff
@@ -176,15 +177,90 @@ func ingestOccurrence(ctx context.Context, client graphql.Client) {
...
+	version := "3.0.3"
+	pkg := model.PkgInputSpec{
+		Type:       "conan",
+		Namespace:  &ns,
+		Name:       "openssl",
+		Version:    &version,
+		Qualifiers: []model.PackageQualifierInputSpec{{Key: "user", Value: "bincrafters"}, {Key: "channel", Value: "stable"}},
+	}
+	certifyVuln := model.CertifyVulnInputSpec{
+		TimeScanned:    time.Now(),
```
pxp928, line 207: getting the following error while ingesting into neo4j:
```
{"level":"error","ts":1677966566.181216,"caller":"cmd/ingest.go:263","msg":"Error in ingesting: input: ingestVulnerability Neo4jError: Neo.ClientError.Statement.TypeError (Unable to construct ZonedDateTime value: `Unknown time-zone ID: Local`)\n\n","stacktrace":"github.com/guacsec/guac/cmd/graphql_playground/cmd.IngestVulnerability\n\t/Users/parth/Documents/pxp928/artifact-ff/cmd/graphql_playground/cmd/ingest.go:263\ngithub.com/guacsec/guac/cmd/graphql_playground/cmd.ingestData\n\t/Users/parth/Documents/pxp928/artifact-ff/cmd/graphql_playground/cmd/ingest.go:45"}
```
What was the solution to this?
Keeping this comment to block merge!

lumjjb: this wasn't the same error as the one i encountered - that was more for the missing values, this seems like maybe a formatting issue, maybe on the neo4j ingestion side setting the output of the value encoded to RFC3339 format maybe?

mihaimaruseac: I have encountered the same error, it's because the backend expects a different format for the time field.

I'm working on that

mihaimaruseac (approved):

cmd/graphql_playground/cmd/ingest.go
pxp928: Keeping this open to make sure all the ingestion and query occur properly before merging.

lumjjb (approved):

pkg/assembler/backends/neo4j/certifyVuln.go
```diff
@@ -397,3 +379,276 @@ func generateModelCertifyVuln(pkg *model.Package, vuln model.OsvCveGhsaObject, t
...
+	}
+
+	session := c.driver.NewSession(neo4j.SessionConfig{AccessMode: neo4j.AccessModeWrite})
+	defer session.Close()
+
+	var sb strings.Builder
+	var firstMatch bool = true
+	queryValues := map[string]any{}
+
+	queryValues[timeScanned] = certifyVuln.TimeScanned
```
mihaimaruseac, line 399:
```suggestion
	queryValues[timeScanned] = certifyVuln.TimeScanned.UTC()
```

Based on #538. We need to make sure times are in UTC to not confuse neo4j.

pkg/assembler/graphql/schema/certifyVuln.graphql
```diff
@@ -42,6 +42,8 @@ type CertifyVuln {
   collector: String!
 }
 
+scalar Time
+
```
mihaimaruseac, line 46: Let's remove this, I'm adding it in #538 with comments

cmd/graphql_playground/cmd/ingest.go
mihaimaruseac: Fixing in #538 , the issue was actually in a mismatch between Go's default timezone in `time.Now()` and neo4j's custom type for local timezones. Switching to UTC solves this and also the fact that our times need to be localized with some timezone.

mihaimaruseac (changes requested):

cmd/graphql_playground/cmd/ingest.go
pxp928: fixed.

mihaimaruseac (approved):

PR #538, opened 2023-03-06
Fix issue with Time in Neo4j backend.

mihaimaruseac:
The issue is neither on neo4j nor in ingestion. It is in the way go treats the time field. That is, `time.Now()` returns time in the current local timezone but that contains a confusing `Offset()`. Neo4j wants to store local times as `LocalTime`, mapped to `neo4j.LocalTime` in Go, which is not what we want.

Instead, and to solve issues with timezones, let's just enforce that all GUAC times are in UTC.  Added comments to this end to the schema.

While here, I also changed the example ingestion parameters as the old ones would not parse under Go's `time.Time` rules.


pxp928 (approved):
LGTM!

lumjjb (approved):

PR #544, opened 2023-03-07
change top level heuristic PURL to guac purl

lumjjb:
Changing the top level heuristic from spdx document name to guac purl

### DIFF
--- a/pkg/ingestor/parser/spdx/parse_spdx.go
+++ b/pkg/ingestor/parser/spdx/parse_spdx.go
@@ -76,9 +76,9 @@ func (s *spdxParser) getTopLevelPackage() error {
 	splitImage := strings.Split(s.spdxDoc.DocumentName, "/")
 	var purl string
 	if len(splitImage) == 3 {
-		purl = "pkg:oci/" + splitImage[2] + "?repository_url=" + splitImage[0] + "/" + splitImage[1]
+		purl = "pkg:guac/oci/" + s.spdxDoc.DocumentName
 	} else if len(splitImage) == 2 {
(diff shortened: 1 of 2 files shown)


pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -76,9 +76,9 @@ func (s *spdxParser) getTopLevelPackage() error {
 	splitImage := strings.Split(s.spdxDoc.DocumentName, "/")
 	var purl string
 	if len(splitImage) == 3 {
```
pxp928, line 78: can remove this check and `split` above as not needed.

pxp928 (approved):

pkg/ingestor/parser/spdx/parse_spdx.go
mihaimaruseac: +1

mihaimaruseac (approved):
LGTM, modulo Parth's question

lumjjb:
We still need the check since it may not be an oci image

pxp928:
Does it need to be type `oci`? Seems very error prone again splitting strings.

pkg/ingestor/parser/spdx/parse_spdx.go
lumjjb: apparently email replies to the entire issue, bringing the comments back to this thread 

>@lumjjb: We still need the check since it may not be an oci image
>@pxp928: Does it need to be type oci? Seems very error prone again splitting strings.

lumjjb: well i think its a guac heuristic for OCI images, so maybe we can make it `guac/heuristic/oci/path`?

PR #548, opened 2023-03-07
Try to ingest SLSA

mihaimaruseac:
This is already quite large so I'll definitely split it further.


pkg/assembler/graphql/schema/hasSLSA.graphql
```diff
@@ -15,85 +15,114 @@
...
   key: String!
   value: String!
 }
 
-"""
-HasSLSASpec allows filtering the list of HasSLSA to return.
-"""
+"HasSLSASpec allows filtering the list of HasSLSA to return."
 input HasSLSASpec {
   package: PkgSpec
```
pxp928, line 122: can this all be replaced by `PackageSourceOrArtifactSpec`

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -488,23 +489,20 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
...
-		invalidSubject = true
+	if hasSLSASpec.Artifact != nil {
+		subjectsDefined = subjectsDefined + 1
 	}
-	if hasSLSASpec.Source != nil && hasSLSASpec.Artifact != nil {
-		invalidSubject = true
-	}
-	if invalidSubject {
-		return gqlerror.Errorf("cannot specify more than one subject for CertifyBad query")
+	if subjectsDefined > 1 {
```
lumjjb, line 504: 👍

lumjjb (approved):
lgtm

pkg/assembler/graphql/schema/hasSLSA.graphql
mihaimaruseac: Right, but that would require more changes on the query side, so I was punting on it. However, I already changed a lot, so will do

PR #550, opened 2023-03-08
Prepare for SLSA ingestion

mihaimaruseac:
* Rename `PkgSrcArtObject` to `PackageSourceOrArtifact` for better readability
* Add comments to GraphQL structures
* Reformat the example GraphQL queries
* Split the `HasSLSA` node to contain the subject and the attestation as a separate node
* Allow mixed materials, propagate this to query filter too
* Simplify checking for valid query input

### DIFF
--- a/pkg/assembler/backends/neo4j/hasSLSA.go
+++ b/pkg/assembler/backends/neo4j/hasSLSA.go
@@ -36,6 +36,9 @@ const (
 )
 
 func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpec) ([]*model.HasSlsa, error) {
+	session := c.driver.NewSession(neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
+	defer session.Close()
+
 	err := checkHasSLSAInputs(hasSLSASpec)
 	if err != nil {
 		return nil, err
@@ -45,9 +48,6 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 	if hasSLSASpec.Package == nil && hasSLSASpec.Source == nil && hasSLSASpec.Artifact == nil {
 		queryAll = true
 	}
-
-	session := c.driver.NewSession(neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
-	defer session.Close()
 
 	aggregateHasSLSA := []*model.HasSlsa{}
 
@@ -207,7 +207,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 				collectedHasSLSA := []*model.HasSlsa{}
 				for subject, hasSLSA := range resultHasSlsaMap {
 					if builtFrom, ok := resultBuiltFromMap[subject]; ok {
-						hasSLSA.BuiltFrom = builtFrom
+						hasSLSA.Slsa.BuiltFrom = builtFrom
 					}
 					collectedHasSLSA = append(collectedHasSLSA, hasSLSA)
 				}
@@ -341,7 +341,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 				collectedHasSLSA := []*model.HasSlsa{}
 				for subject, hasSLSA := range resultHasSlsaMap {
 					if builtFrom, ok := resultBuiltFromMap[subject]; ok {
-						hasSLSA.BuiltFrom = builtFrom
+						hasSLSA.Slsa.BuiltFrom = builtFrom
 					}
 					collectedHasSLSA = append(collectedHasSLSA, hasSLSA)
 				}
@@ -471,7 +471,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 				collectedHasSLSA := []*model.HasSlsa{}
 				for subject, hasSLSA := range resultHasSlsaMap {
 					if builtFrom, ok := resultBuiltFromMap[subject]; ok {
-						hasSLSA.BuiltFrom = builtFrom
+						hasSLSA.Slsa.BuiltFrom = builtFrom
 					}
 					collectedHasSLSA = append(collectedHasSLSA, hasSLSA)
 				}
@@ -578,10 +578,11 @@ func setHasSLSAValues(sb *strings.Builder, hasSLSASpec *model.HasSLSASpec, first
 	}
 }
 
-func generateModelHasSLSA(subject model.PackageSourceOrArtifact, builder *model.Builder, slsaPredicate []interface{}, buildType,
-	slsaVersion, origin, collector string, startedOn, finishedOn time.Time) *model.HasSlsa {
-	hasSLSA := model.HasSlsa{
-		Subject:       subject,
+func generateModelHasSLSA(subject model.PackageSourceOrArtifact,
+	builder *model.Builder, slsaPredicate []interface{}, buildType,
+	slsaVersion, origin, collector string,
+	startedOn, finishedOn time.Time) *model.HasSlsa {
+	slsa := &model.Slsa{
 		BuiltBy:       builder,
 		BuildType:     buildType,
 		SlsaPredicate: getCollectedPredicate(slsaPredicate),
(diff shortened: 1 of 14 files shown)


pxp928 (approved):

pkg/assembler/backends/testing/hasSLSA.go
```diff
@@ -98,65 +146,56 @@ func (c *demoClient) registerHasSLSA(selectedPackage *model.Package, selectedSou
...
-	if hasSLSASpec.Package != nil && hasSLSASpec.Artifact != nil {
-		return nil, gqlerror.Errorf("cannot specify package and artifact together for hasSLSASpec")
+	if hasSLSASpec.Subject.Artifact != nil {
+		subjectsDefined = subjectsDefined + 1
 	}
-	if hasSLSASpec.Source != nil && hasSLSASpec.Artifact != nil {
-		return nil, gqlerror.Errorf("cannot specify source and artifact together for hasSLSASpec")
+	if subjectsDefined > 1 {
+		return nil, gqlerror.Errorf("Must specify at most one subject (package, source, or artifact)")
 	}
```
lumjjb, line 168: hmm, does the 0 case need to be handled as well?

lumjjb (approved):

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -36,22 +36,22 @@ const (
...
+	defer session.Close()
+
 	err := checkHasSLSAInputs(hasSLSASpec)
 	if err != nil {
 		return nil, err
 	}
 
 	queryAll := false
-	if hasSLSASpec.Package == nil && hasSLSASpec.Source == nil && hasSLSASpec.Artifact == nil {
+	if hasSLSASpec.Subject.Package == nil && hasSLSASpec.Subject.Source == nil && hasSLSASpec.Subject.Artifact == nil {
```
pxp928, line 48: this can be changed to just `hasSLSASpec.Subject == nil` to query all

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -36,22 +36,22 @@ const (
...
 		queryAll = true
 	}
 
-	session := c.driver.NewSession(neo4j.SessionConfig{AccessMode: neo4j.AccessModeRead})
-	defer session.Close()
-
 	aggregateHasSLSA := []*model.HasSlsa{}
 
-	if hasSLSASpec.Package != nil || queryAll {
+	if hasSLSASpec.Subject.Package != nil || queryAll {
```
pxp928, line 54: will have to add a check here at `hasSLSASpec.Subject =! nil` else it will error with queryAll

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -114,7 +114,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 
 			sb.WriteString(query)
 			firstMatch = true
-			setPkgMatchValues(&sb, hasSLSASpec.Package, false, &firstMatch, queryValues)
+			setPkgMatchValues(&sb, hasSLSASpec.Subject.Package, false, &firstMatch, queryValues)
```
pxp928, line 117: same here will have to put a check in that `subject` is not nil for query all case

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -220,7 +220,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 		aggregateHasSLSA = append(aggregateHasSLSA, result.([]*model.HasSlsa)...)
 	}
 
-	if hasSLSASpec.Source != nil || queryAll {
+	if hasSLSASpec.Subject.Source != nil || queryAll {
```
pxp928, line 223: same as above

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -252,7 +252,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 			"\nWITH *, hasSLSA, null AS objPkgVersion"
 
 		sb.WriteString(query)
-		setSrcMatchValues(&sb, hasSLSASpec.Source, false, &firstMatch, queryValues)
+		setSrcMatchValues(&sb, hasSLSASpec.Subject.Source, false, &firstMatch, queryValues)
```
pxp928, line 255: same as above

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -354,7 +354,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 		aggregateHasSLSA = append(aggregateHasSLSA, result.([]*model.HasSlsa)...)
 	}
 
-	if hasSLSASpec.Artifact != nil || queryAll {
+	if hasSLSASpec.Subject.Artifact != nil || queryAll {
```
pxp928, line 357: same as above

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -384,7 +384,7 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
 			"\nWITH *, hasSLSA, null AS objPkgVersion"
 
 		sb.WriteString(query)
-		setArtifactMatchValues(&sb, hasSLSASpec.Artifact, false, &firstMatch, queryValues)
+		setArtifactMatchValues(&sb, hasSLSASpec.Subject.Artifact, false, &firstMatch, queryValues)
```
pxp928, line 387: same as above

pkg/assembler/backends/neo4j/hasSLSA.go
```diff
@@ -488,23 +488,20 @@ func (c *neo4jClient) HasSlsa(ctx context.Context, hasSLSASpec *model.HasSLSASpe
...
+// TODO(pxp928): combine with testing backend in shared utility
 func checkHasSLSAInputs(hasSLSASpec *model.HasSLSASpec) error {
-	invalidSubject := false
-	if hasSLSASpec.Package != nil && hasSLSASpec.Source != nil && hasSLSASpec.Artifact != nil {
-		invalidSubject = true
-	}
-	if hasSLSASpec.Package != nil && hasSLSASpec.Source != nil {
-		invalidSubject = true
+	subjectsDefined := 0
+	if hasSLSASpec.Subject.Package != nil {
```
pxp928, line 494: need a check here that `hasSLSASpec.Subject != nil` as that is a valid case

pxp928 (commented):
Find some issues. As I was running through my code changes

pkg/assembler/backends/neo4j/hasSLSA.go
mihaimaruseac: I think I need to handle that case separately. I still have to do e2e testing after all ingestion is done. I'm planning to test/change that there.

pkg/assembler/backends/testing/hasSLSA.go
pxp928: For the query if the subject is not defined, you query all

mihaimaruseac: Right, but will uncover these cases on the e2e tests and can fix there.

pkg/assembler/backends/neo4j/hasSLSA.go
mihaimaruseac: This is now handled by where this gets called from

pkg/assembler/backends/testing/hasSLSA.go
mihaimaruseac: I don't think there's an issue here: 0 means match everything and it should match (testing e2e pending to uncover other bugs), 1 means match exactly one of them

pkg/assembler/backends/neo4j/hasSLSA.go
mihaimaruseac: Handled for now, but with e2e testing this might change in next PR(s)

pkg/assembler/backends/neo4j/hasSLSA.go
mihaimaruseac: Handled for now, but with e2e testing this might change in next PR(s)

pxp928 (approved):

PR #552, opened 2023-03-08
[Fix] Remove unneeded query and remove re-used code between backends

pxp928:
- consolidate pkg and source to subject in `isOccurrence`
- move all converts to helper package to be shared by both backends
- remove unneeded query at pkgName level for existing verbs

PR #553, opened 2023-03-08
Add ingestion mutation for SLSA

mihaimaruseac:
Need to make sure we use `XXXSpec` for read-only queries and `XXXInput` for mutation, since they have different nullability rules and semantics. Took a while to found bugs caused by using the wrong one (after merge conflict resolution).

While here also fix a typo in a file name and remove leftover debug print

Note: `pkg/assembler/backends/testing/hasSLSA.go` has been rewritten a lot, so github might hide it. That actually contains most of this PR


pkg/assembler/backends/neo4j/certifyVuln.go
```diff
@@ -401,7 +400,6 @@ func (c *neo4jClient) IngestVulnerability(ctx context.Context, pkg model.PkgInpu
 		sb.WriteString(merge)
 		sb.WriteString(returnValue)
 
-		fmt.Println(sb.String())
```
pxp928, line 404: you can remove this from your PR if you want. it's part of PR #552

pkg/assembler/backends/testing/backend.go
```diff
@@ -114,10 +114,6 @@ func GetBackend(args backends.BackendArgs) (backends.Backend, error) {
 	if err != nil {
 		return nil, err
 	}
-	err = registerAllHasSLSA(client)
```
pxp928, line 117: do we no longer want to add data to the testing backend?

pxp928 (approved):
LGTM

pkg/assembler/backends/testing/backend.go
pxp928: Does this need to be replaced by another? Or ingestion via the client only?

mihaimaruseac: Via client/playground is ok here.

Don't remove the others yet as we might use them for testing/perf

pkg/assembler/backends/neo4j/certifyVuln.go
mihaimaruseac: Dropped that specific commit from this branch

lumjjb (approved):

PR #555, opened 2023-03-09
Split materials and builder from SLSA input

mihaimaruseac:
We expect source trees to already exist when performing the ingestion of the evidence tree.

The fact that we only need this when trying to ingest SLSA for the Neo4J backend signals that we have a significant bug in the in-memory backend.

### DIFF
--- a/pkg/assembler/clients/operations/hasSLSA.graphql
+++ b/pkg/assembler/clients/operations/hasSLSA.graphql
@@ -17,29 +17,74 @@
 
 # Defines the GraphQL operations to ingest SLSA attestations into GUAC
 
-mutation SLSAForPackage($pkg: PkgInputSpec!, $slsa: SLSAInputSpec!) {
+mutation SLSAForPackage($pkg: PkgInputSpec!, $materials: [PackageSourceOrArtifactInput!]!, $builder: BuilderInputSpec!, $slsa: SLSAInputSpec!) {
   ingestPackage(pkg: $pkg) {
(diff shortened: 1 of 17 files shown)


pkg/assembler/clients/operations/hasSLSA.graphql
```diff
@@ -17,29 +17,74 @@
...
   }
 }
 
-mutation SLSAForSource($source: SourceInputSpec!, $slsa: SLSAInputSpec!) {
+mutation SLSAForSource($source: SourceInputSpec!, $materials: [PackageSourceOrArtifactInput!]!, $builder: BuilderInputSpec!, $slsa: SLSAInputSpec!) {
   ingestSource(source: $source) {
     ...allSourceTree
   }
-  ingestSLSA(subject: {source: $source}, slsa: $slsa) {
+  ingestMaterials(materials: $materials) {
```
pxp928, line 48: Looks good. Just a question on if there is a way we can call the `ingestPackage`, `ingestSource`...etc here instead of it having to be done in the `ingestMaterials`?

pxp928 (approved):
LGTM just a question

pkg/assembler/clients/operations/hasSLSA.graphql
mihaimaruseac: We cannot since we cannot look at the insides of the JSON.

lumjjb (approved):

PR #556, opened 2023-03-09
added ingest certifyBad for in-memory with tests

pxp928:

### DIFF
--- a/pkg/assembler/backends/neo4j/certifyBad.go
+++ b/pkg/assembler/backends/neo4j/certifyBad.go
@@ -17,31 +17,30 @@ package neo4jBackend
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
+	"github.com/guacsec/guac/pkg/assembler/backends/helper"
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/neo4j/neo4j-go-driver/v4/neo4j"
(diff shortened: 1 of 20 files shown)


lumjjb (approved):

pxp928:
Will have to look into the pkg and source duplication issue in another PR, but CertifyBad for in-memory works.

pkg/assembler/backends/neo4j/certifyBad.go
```diff
@@ -117,7 +122,7 @@ func (c *neo4jClient) CertifyBad(ctx context.Context, certifyBadSpec *model.Cert
 		aggregateCertifyBad = append(aggregateCertifyBad, result.([]*model.CertifyBad)...)
 	}
 
-	if certifyBadSpec.Source != nil || queryAll {
+	if queryAll || certifyBadSpec.Subject != nil && certifyBadSpec.Subject.Source != nil {
```
lumjjb, line 125: nit: i like to have brackets whenever there's a mix of ands and ors since left association is a bit weird to think about - or maybe it's just me

lumjjb (approved):

mihaimaruseac (approved):

pkg/assembler/backends/neo4j/certifyBad.go
pxp928: sure I can add the brackets

PR #557, opened 2023-03-09
Add Vcs URI parsing

lumjjb:
Add Vcs URI parsing

### DIFF
--- a/pkg/assembler/helpers/vcs.go
+++ b/pkg/assembler/helpers/vcs.go
@@ -1,0 +1,88 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):

pkg/assembler/helpers/vcs.go
```diff
@@ -0,0 +1,88 @@
+//
```
mlieberman85, line 1: Should this live in a common helpers directory? Looking at this, this function is more or less the same one we use in the github collector to parse a URI that comes in as well.

lumjjb: yea, maybe a helper function to call out, and then leave the conversion to the graphql model in assembler helpers.

lumjjb: i think this could be put into a good first issue and handled as another PR?

lumjjb: https://github.com/guacsec/guac/issues/572

pxp928 (approved):
LGTM

arorasoham9:
Is it alright if the functions if vcs.go and vcs_test.go are moved to misc?

pxp928:
> Is it alright if vcs.go and vcs_test.go are moved to misc?

We can either move to misc or internal helper directory.

arorasoham9:
Both the files lie in "pkg/assembler/helpers/" dir. Should I move them to misc or are they fine where they are?

PR #560, opened 2023-03-10
Don't ingest packages twice.

mihaimaruseac:
Because we were creating the version node over and over again, we were in a situation where the same package would be ingested multiple times.

This was blocking progress on SLSA ingestion, so it is now fixed.


pxp928 (approved):
Nice!

lumjjb (approved):
👍 👍 👍 👍 👍 👍

Issue #571, opened 2023-03-13
Update Certifier to use GraphQL interface rather than cypher


lumjjb:
We've updated the assembler to be interfaced via GraphQL. Cypher should no longer be the way in which the metadata is queried. Certifiers use [component queriers](https://github.com/guacsec/guac/tree/main/pkg/certifier/components), which currently use Cypher to query components it is looking for. It should be updated with calls to the [graphql API](https://github.com/guacsec/guac/tree/main/pkg/assembler/graphql/schema) instead.

mihaimaruseac:
First step is to define the query in clients/operations, then autogenerate the Go binding, then replace with a call to that, similar to how we do ingestion

pxp928:
completed with https://github.com/guacsec/guac/pull/634

PR #586, opened 2023-03-15
update scorecard certifier to use graphQL query

### LINKED ISSUE #573: Scorecard runner unable to use sha due to prefix in digest
## Description:
The current implementation of the scorecard runner in the scorecard.go file takes a digest with a "sha1:" prefix at the beginning. However, this prefix causes the scorecard runner not to be able to use the sha.

I have identified the issue and proposed a code fix for it. The fix involves removing the `"sha1:"` prefix from the digest before passing it to the scorecard runner.

## Solution:

Here's the proposed code fix:

``` go
digest := strings.TrimPrefix(digest, "sha1:")
```

I have tested this, and it will allow the scorecard runner to properly run.

pxp928:
Currently in draft until the ID/back edge work has been completed on the in-memory backend. 

Addresses issue #571 to update `source_artifact` in components  and fixes #573

### DIFF
--- a/cmd/guacone/cmd/scorecard.go
+++ b/cmd/guacone/cmd/scorecard.go
@@ -23,7 +23,7 @@ import (
 	"time"
 
 	"github.com/Khan/genqlient/graphql"
-	sc "github.com/guacsec/guac/pkg/certifier/components/source_artifact"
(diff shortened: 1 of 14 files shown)


pxp928:
Closing this PR as it is out of date and changing the implementation.

Issue #583, opened 2023-03-14
Update Vuln parser to generate GraphQL nodes


jeffmendoza:
The vuln parser `pkg/ingestor/parser/vuln` code is still on old GuacNode/GuacEdge generation. The tests are fully commented out.

It should be generating package/osv/cve/ghsa nodes and certifyVuln/isVulnerability connectors. IngestPredicates need to be added to `pkg/assembler/assembler.go` for vuln ingestion.

jeffmendoza:
I'll start working on this.

PR #588, opened 2023-03-15
Update Vuln parser to new ontology

jeffmendoza:
Adds IsOccurence for any digests found under packages named in the
subject. Adds CertifyVuln between all packages and vulns found, using OSV. Adds
IsVuln for each vuln found mapping from OSV to either CVE or GHSA.

Adds a helper for OSV to CVE/GHSA translaton. Also some testing helpers for
pointers to literals.

### DIFF
--- a/pkg/assembler/helpers/vuln_test.go
+++ b/pkg/assembler/helpers/vuln_test.go
@@ -1,0 +1,85 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package helpers_test
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"github.com/guacsec/guac/pkg/assembler/helpers"
+)
+
+func TestOSVToGHSACVE(t *testing.T) {
+	tests := []struct {
+		Name    string
+		Input   string
+		ExpCVE  *generated.CVEInputSpec
+		ExpGHSA *generated.GHSAInputSpec
+		ExpErr  bool
+	}{
+		{
+			Name:  "Good CVE",
+			Input: "CVE-1999-1234",
+			ExpCVE: &generated.CVEInputSpec{
+				CveId: "CVE-1999-1234",
+				Year:  "1999",
+			},
+			ExpGHSA: nil,
+			ExpErr:  false,
+		},
+		{
+			Name:   "Good GHSA",
+			Input:  "GHSA-1234-asdf-qwer",
+			ExpCVE: nil,
+			ExpGHSA: &generated.GHSAInputSpec{
+				GhsaId: "GHSA-1234-asdf-qwer",
+			},
+			ExpErr: false,
+		},
+		{
+			Name:    "Bad CVE",
+			Input:   "CVE-1999",
+			ExpCVE:  nil,
+			ExpGHSA: nil,
+			ExpErr:  true,
+		},
+		{
+			Name:    "Bad",
+			Input:   "asdf",
+			ExpCVE:  nil,
+			ExpGHSA: nil,
+			ExpErr:  true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			cve, ghsa, err := helpers.OSVToGHSACVE(test.Input)
+			if (err != nil) != test.ExpErr {
+				t.Errorf("Expected error: %v got: %v", test.ExpErr, err)
+			}
+			if err != nil {
+				return
+			}
+			if diff := cmp.Diff(test.ExpCVE, cve); diff != "" {
+				t.Errorf("Unexpected results. (-want +got):\n%s", diff)
+			}
+			if diff := cmp.Diff(test.ExpGHSA, ghsa); diff != "" {
+				t.Errorf("Unexpected results. (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
--- a/pkg/ingestor/parser/vuln/vuln.go
+++ b/pkg/ingestor/parser/vuln/vuln.go
@@ -13,115 +13,74 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// The Vulnerability attestation parser parses the attestation defined by
-// by the certifier using the predicate type"https://in-toto.io/attestation/vuln/v0.1"
-// Based on the information contained, a package node is generated with just purl and
-// digest (if found). This package node is merged into the graph database as it already exists.
+// Package vuln attestation parser parses the attestation defined by by
+// the certifier using the predicate type
+// "https://in-toto.io/attestation/vuln/v0.1" Three different types of ingest
+// predicates are created.
 //
-// An attestation node is generated that contains the information stored in the attestation document.
-// The attestation node is linked to the package node via a "attestation" edge type.
+// - IsOccurences are created mapping between any package
+// purls found in the subject, and any digests found under those.
 //
-// Depending on the number of vulnerabilities found, vulnerability nodes are generated that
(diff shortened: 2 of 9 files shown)


pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -0,0 +1,167 @@
...
+func parseSubject(s *attestation_vuln.VulnerabilityStatement) ([]*generated.PkgInputSpec,
+	error) {
+	var ps []*generated.PkgInputSpec
+	for _, sub := range s.StatementHeader.Subject {
+		p, err := helpers.PurlToPkg(sub.Name)
+		if err != nil {
+			return nil, fmt.Errorf("bad purl in statement header: %w", err)
+		}
+		ps = append(ps, p)
+		// FIXME Do we do anything with Digest here?
```
pxp928, line 97: if there is a digest, you can create an artifact that you can link to the created package above (via the purl) using `IsOccurrence`

pxp928 (commented):
Looks good! just a small addition

pkg/assembler/helpers/vuln_test.go
```diff
@@ -0,0 +1,82 @@
...
+			ExpErr:  true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.Name, func(t *testing.T) {
+			cve, ghsa, err := helpers.OSVToGHSACVE(test.Input)
+			if (err != nil) != test.ExpErr {
+				t.Errorf("Expected error: %v got: %v", test.ExpErr, err)
+			}
+			if diff := cmp.Diff(test.ExpCVE, cve); diff != "" {
```
lumjjb, line 74: continue checks only if err == nil

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -0,0 +1,167 @@
...
+func parseMetadata(s *attestation_vuln.VulnerabilityStatement,
+	doc *processor.Document) *generated.VulnerabilityMetaDataInput {
+	return &generated.VulnerabilityMetaDataInput{
+		TimeScanned:    *s.Predicate.Metadata.ScannedOn,
+		DbUri:          s.Predicate.Scanner.Database.Uri,
+		DbVersion:      s.Predicate.Scanner.Database.Version,
+		ScannerUri:     s.Predicate.Scanner.Uri,
+		ScannerVersion: s.Predicate.Scanner.Version,
+		Origin:         doc.SourceInformation.Source,
+		Collector:      doc.SourceInformation.Collector,
```
lumjjb, line 111: the origin and collector stuff will be handled by the [graph builder](https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/common/graph_builder.go), not required to set inside the parser, also will need to add something [here](https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/common/graph_builder.go#L61).

(it looks like i may have forgotten to add plumbing for Slsa predicate there, if you want to help with a drive-by :) )

lumjjb (approved):
Nice! couple comments else LGTM

pkg/ingestor/parser/vuln/vuln.go
jeffmendoza: Got it, I'll add this and slsa there.

pkg/ingestor/parser/vuln/vuln.go
jeffmendoza: SG

jeffmendoza:
Fixes #583

pkg/assembler/helpers/vuln_test.go
jeffmendoza: Done

pkg/ingestor/parser/vuln/vuln.go
jeffmendoza: Done.

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -0,0 +1,179 @@
...
+		ps = append(ps, p)
+		for a, d := range sub.Digest {
+			io := assembler.IsOccurenceIngest{
+				Pkg: p,
+				Artifact: &generated.ArtifactInputSpec{
+					Algorithm: a,
+					Digest:    d,
+				},
+				IsOccurence: &generated.IsOccurrenceInputSpec{
+					Justification: "Reported in vulnerability certifier", // FIXME is this ok?
```
jeffmendoza, line 109: What should we put here?

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -0,0 +1,179 @@
...
+		cve, ghsa, err := helpers.OSVToGHSACVE(id.VulnerabilityId)
+		if err != nil {
+			return nil, nil, fmt.Errorf("could not parse vuln id in attestation: %w", err)
+		}
+		iv := assembler.IsVulnIngest{
+			OSV:  v,
+			CVE:  cve,
+			GHSA: ghsa,
+			IsVuln: &generated.IsVulnerabilityInputSpec{
+				Justification: "Decoded OSV ID", // FIXME is this ok?
```
jeffmendoza, line 146: And here.

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -0,0 +1,179 @@
...
+// exists.
+//
+// An attestation node is generated that contains the information stored in the
+// attestation document.  The attestation node is linked to the package node
+// via a "attestation" edge type.
+//
+// Depending on the number of vulnerabilities found, vulnerability nodes are
+// generated that contain just the vulnerability ID that can be used to query
+// for more information as needed.  The vulnerability node is linked to the
+// attestation node via a "vulnerable" edge type.
```
jeffmendoza, line 30: The package doc here is probably out of date.

pkg/ingestor/parser/vuln/vuln.go
lumjjb: probably something [similar to the SLSA ](https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/slsa/parser_slsa.go#L161)one will do, more so how it was derived so folks can determine if they trust the fidelity of the field (on top of the trust info of the document)

pkg/ingestor/parser/vuln/vuln.go
lumjjb: Yea that's fine, maybe "data" instead of "ID"

internal/testing/val/val.go
```diff
@@ -0,0 +1,28 @@
...
+func Bool(v bool) *bool { return &v }
+
+// Int helps get int pointers for test literals
+func Int(v int) *int { return &v }
+
+// Int64 helps get int64 pointers for test literals
+func Int64(v int64) *int64 { return &v }
+
+// String helps get string pointers for test literals
+func String(v string) *string { return &v }
```
mihaimaruseac, line 28: Should we rename these to contain `Ptr` or something in them?

Readability-wise, `PtrFromString("foo")` is better than `String("foo")`

mihaimaruseac: Actually, we have an issue here. Because the arguments are copies, each time the function gets called we get a different argument. See https://go.dev/play/p/qXs-hnrChcD

mihaimaruseac (commented):
LG from the pov of integrating with the backend. Leaving it to Parth for second approval

internal/testing/val/val.go
jeffmendoza: keeping these short helps them read nicely, ex:
```
	Pkg: &generated.PkgInputSpec{
		Type:      "maven",
		Namespace: val.String("org.apache.logging.log4j"),
		Name:      "log4j-core",
		Version:   val.String("2.8.1"),
		Subpath:   val.String(""),
	},
```
vs
```
	Pkg: &generated.PkgInputSpec{
		Type:      "maven",
		Namespace: val.PtrFromString("org.apache.logging.log4j"),
		Name:      "log4j-core",
		Version:   val.PtrFromString("2.8.1"),
		Subpath:   val.PtrFromString(""),
	},
```
On another project, I used a lot of go-github, which does this:
https://github.com/google/go-github/blob/v50.1.0/github/github.go#L1499-L1513
ex: https://github.com/google/go-github/blob/v50.1.0/github/actions_artifacts_test.go#L204-L210

Maybe it makes sense to change the package name from `val` to `ptrfrom` ?

mihaimaruseac: I think since this is for testing only it is ok to use like this. I was more worried about the pointer change when you call it twice, but that has no impact if this is used only in testing code.

`ptrform` as package name looks good to me, thank you

pxp928 (approved):
LGTM!

internal/testing/val/val.go
jeffmendoza: Ah, yep. Added a package comment as well to clarify, and updated the name.

PR #598, opened 2023-03-16
fix duplication issue with src, package and hasSourceAt. Make hasSourceAt back link a list

pxp928:


lumjjb (approved):

lumjjb (approved):

pkg/assembler/backends/testing/backend.go
```diff
@@ -62,19 +60,17 @@ type demoClient struct {
...
 	certifyPkg          []*model.CertifyPkg
 	certifyVuln         []*model.CertifyVuln
-	hasSourceAt         []*model.HasSourceAt
 	certifyScorecard    []*model.CertifyScorecard
 	certifyBad          []*model.CertifyBad
 	isVulnerability     []*model.IsVulnerability
 	certifyVEXStatement []*model.CertifyVEXStatement
 	hasSLSA             []*model.HasSlsa
 	id                  uint32
+	index               indexType
```
mihaimaruseac, line 69: We have a global `index` and one in the client. We likely need to use only one of them, either the global (if we assume the client is singleton) or the one in the client struct (better for testing as we can then mock its contents).

mihaimaruseac (commented):
LGTM, modulo the issue of globals vs variables in the client struct

pkg/assembler/backends/testing/backend.go
pxp928: makes more sense to do it at the client struct. I will remove the duplicate.

pkg/assembler/backends/testing/hasSourceAt.go
```diff
@@ -17,15 +17,17 @@ package testing
...
+	"strconv"
 	"time"
 
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
 // Internal data: link between sources and packages (HasSourceAt)
-type srcMaps []*srcMapLink
+type hasSrcMaps map[uint32]*srcMapLink
```
mihaimaruseac, line 30: Hmm, shouldn't the map between id and `srcMapLink` element be done from the index? Meaning we don't need one additional map here? (lists are more concise than maps in memory, we don't need to traverse the list in most cases anyway)

mihaimaruseac (approved):

pkg/assembler/backends/testing/hasSourceAt.go
pxp928: yeah we are making a change here also. The key will not be a sorted concatenation of sourceID, packageID, justification, origin, collector, and knownSince. If those are provided we can pull out the value without looping through it.

mihaimaruseac: But you'll have to loop each time one of those values is missing. Plus, it's not too feasible to get an int key by concatenating other infos (it's kind of like hashing).

If we want to go from IDs to srcMapLink then we need separate indices but the `hasSrcMaps` should still be a list:
- sourceID/packageID -> get it from global index, get the hasSrcMaps from it
- justification/origin/collector/knownsince -> separate local indices

Now whenever you have a filter, say sourceID and knownSince, you get the positions in the list from these 2 and then look at the common elements. But this would be like implementing index merging from databases.

pxp928: yes, that is true if we are missing values. Yeah, it might be too complicated. We can drop it and revert back to a list?

mihaimaruseac: Yes, I think this should be always a list and we can separate indexing if we discover that this is a common bottleneck.

pxp928: so do we want to do index merging (like you are saying above?) or hold off for the time being until it does become a bottleneck

mihaimaruseac (approved):

PR #599, opened 2023-03-17
[memory] Add IDs for vulnerability (cve, osv, ghsa)

pxp928:


pkg/assembler/backends/testing/osv.go
```diff
@@ -17,77 +17,181 @@ package testing
...
+		return nil, gqlerror.Errorf("ID does not match existing node")
+	}
+
+	osvIDList := []*model.OSVId{}
+	if osvIDNode, ok := node.(*osvIDNode); ok {
+		if filter != nil && noMatchLowerCase(filter.OsvID, osvIDNode.osvID) {
+			return nil, nil
 		}
+		osvIDList = append(osvIDList, &model.OSVId{
+			ID:    fmt.Sprintf("%d", osvIDNode.id),
```
lumjjb, line 166: nit: may be useful to make this a helper function in case we decide to change it and also in case someone puts a slightly different format string that may lead to something different

lumjjb (approved):
LGTM, minor nit

mihaimaruseac (approved):

pkg/assembler/backends/testing/backend.go
```diff
@@ -128,6 +133,13 @@ func noMatch(filter *string, value string) bool {
 	return false
 }
 
+func noMatchLowerCase(filter *string, value string) bool {
+	if filter != nil {
+		return value != strings.ToLower(*filter)
+	}
+	return false
+}
+
```
mihaimaruseac, line 142: Should we instead canonicalize before trying to ingest? This way, we have functions to convert to lower case, remove `-`, etc, and then the match functions can stay as they are

pkg/assembler/backends/testing/cve.go
```diff
@@ -17,69 +17,199 @@ package testing
...
-	newCve := &model.Cve{Year: year}
-	newCve = registerCveID(newCve, idLower)
-	c.cve = append(c.cve, newCve)
-
-	return newCve
+// Internal data: osv
+type cveMap map[string]*cveNode
+type cveNode struct {
+	id     uint32
+	year   string
```
mihaimaruseac, line 60: Should the year be an int?

pkg/assembler/backends/testing/osv.go
```diff
@@ -17,77 +17,181 @@ package testing
...
-	c.osv = append(c.osv, newOsv)
+const osv string = "osv"
 
-	return newOsv
+// Internal data: osv
+type osvMap map[string]*osvNode
+type osvNode struct {
+	id     uint32
+	osvIDs osvIDMap
+}
```
mihaimaruseac, line 59: This is missing the string key.

pxp928: Which string key?

mihaimaruseac: The key that is used in the map above. If you have `map[string]*Smth` then in `type Smth struct` we need a string field to point back to the key, so we can go back up (given a pointer, find the string that is used as a key).

If we don't need to index on the string, then we don't need the map either, it's just a list.

Compare https://github.com/mihaimaruseac/guac/blob/9d9c9be4c18a1f2645353867386d81ca26e20970/server/backends/memory/trees.go#LL33-L38C2 (where we have map and string key) with https://github.com/mihaimaruseac/guac/blob/9d9c9be4c18a1f2645353867386d81ca26e20970/server/backends/memory/trees.go#L102-L108 where we don't have map, just a list

pxp928: Makes more sense to make it a list as it will be a single key. I will change that.

pxp928: On second thought it might be better to add the type as the key. In case we decide to add more types in the future. I will update.

mihaimaruseac (approved):

pkg/assembler/backends/testing/cve.go
pxp928: Changing it to Int for the year in the schema.

pxp928: I will make another PR to make this change as this one is already large.

PR #600, opened 2023-03-18
Update artifacts, hashEquals, and isOccurrance with internal structs and ids

jeffmendoza:
This is working, but is pretty rough.

registerAlls are not all converted, some commented out.
Some code duplication, as we are trying to work on the files in parallel, and want to avoid bad merge conflicts.
The query optimizations are missing on verbs for when an ID is not specified, but one of the nouns is. In this case we could only search the backedges instead of searching the whole verb list.
Code consistency is not great between these and the others, again working in parallel.

### DIFF
--- a/pkg/assembler/backends/testing/artifact.go
+++ b/pkg/assembler/backends/testing/artifact.go
@@ -39,6 +39,12 @@ type artStruct struct {
 
 func (n *artStruct) getID() uint32 { return n.id }
 
+func (n *artStruct) getHashEquals() []uint32 { return n.hashEquals }
+func (n *artStruct) setHashEquals(id uint32) { n.hashEquals = append(n.hashEquals, id) }
+
+func (n *artStruct) getOccurrences() []uint32 { return n.occurrences }
+func (n *artStruct) setOccurrences(id uint32) { n.occurrences = append(n.occurrences, id) }
+
 // TODO convert to unit tests
 // func registerAllArtifacts(c *demoClient) {
 // 	c.IngestArtifact(context.Background(), &model.ArtifactInputSpec{
--- a/pkg/assembler/backends/testing/artifact.go
+++ b/pkg/assembler/backends/testing/artifact.go
@@ -17,67 +17,154 @@ package testing
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"strconv"
 	"strings"
+
+	"github.com/vektah/gqlparser/v2/gqlerror"
 
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 )
 
-func registerAllArtifacts(client *demoClient) {
-	// strings.ToLower(string(checksum.Algorithm)) + ":" + checksum.Value
-	client.registerArtifact("sha256", "6bbb0da1891646e58eb3e6a63af3a6fc3c8eb5a0d44824cba581d2e14a0450cf")
-	client.registerArtifact("sha1", "7A8F47318E4676DACB0142AFA0B83029CD7BEFD9")
-	client.registerArtifact("sha512", "374AB8F711235830769AA5F0B31CE9B72C5670074B34CB302CDAFE3B606233EE92EE01E298E5701F15CC7087714CD9ABD7DDB838A6E1206B3642DE16D9FC9DD7")
+// Internal data: Artifacts
+type artMap map[string]*artStruct
+type artStruct struct {
+	id          uint32
+	algorithm   string
+	digest      string
+	hashEquals  []uint32
+	occurrences []uint32
+}
+
+func (n *artStruct) getID() uint32 { return n.id }
+
+func registerAllArtifacts(c *demoClient) {
+	c.IngestArtifact(context.Background(), &model.ArtifactInputSpec{
+		Algorithm: "sha256",
+		Digest:    "6bbb0da1891646e58eb3e6a63af3a6fc3c8eb5a0d44824cba581d2e14a0450cf",
+	})
+	c.IngestArtifact(context.Background(), &model.ArtifactInputSpec{
+		Algorithm: "sha1",
+		Digest:    "7A8F47318E4676DACB0142AFA0B83029CD7BEFD9",
+	})
+	c.IngestArtifact(context.Background(), &model.ArtifactInputSpec{
+		Algorithm: "sha512",
+		Digest:    "374AB8F711235830769AA5F0B31CE9B72C5670074B34CB302CDAFE3B606233EE92EE01E298E5701F15CC7087714CD9ABD7DDB838A6E1206B3642DE16D9FC9DD7",
+	})
 }
 
 // Ingest Artifacts
 
-func (c *demoClient) registerArtifact(algorithm, digest string) *model.Artifact {
-	// enforce lowercase for both the algorithm and digest when ingesting
-	lowerCaseDigest := strings.ToLower(digest)
-	lowerCaseAlgorithm := strings.ToLower(algorithm)
+func (c *demoClient) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (*model.Artifact, error) {
+	algorithm := strings.ToLower(artifact.Algorithm)
+	digest := strings.ToLower(artifact.Digest)
+	a, err := c.artifactByKey(algorithm, digest)
 
-	for _, a := range c.artifacts {
-		if a.Digest == lowerCaseDigest && a.Algorithm == lowerCaseAlgorithm {
-			return a
+	if err != nil {
+		a = &artStruct{
+			id:        c.getNextID(),
+			algorithm: algorithm,
+			digest:    digest,
+		}
+		c.index[a.id] = a
+		c.artifacts[strings.Join([]string{algorithm, digest}, ":")] = a
+	}
+
+	return convArtifact(a), nil
+}
+
+func (c *demoClient) artifactByID(id uint32) (*artStruct, error) {
+	o, ok := c.index[id]
+	if !ok {
+		return nil, errors.New("could not find artifact")
+	}
+	a, ok := o.(*artStruct)
+	if !ok {
+		return nil, errors.New("not an artifact")
+	}
+	return a, nil
+}
+
+func (c *demoClient) artifactByKey(alg, dig string) (*artStruct, error) {
+	algorithm := strings.ToLower(alg)
+	digest := strings.ToLower(dig)
+	if a, ok := c.artifacts[strings.Join([]string{algorithm, digest}, ":")]; ok {
+		return a, nil
+	}
+	return nil, errors.New("artifact not found")
+}
+
+func (c *demoClient) artifactExact(artifactSpec *model.ArtifactSpec) (*artStruct, error) {
+	algorithm := strings.ToLower(nilToEmpty(artifactSpec.Algorithm))
+	digest := strings.ToLower(nilToEmpty(artifactSpec.Digest))
+
+	// If ID is provided, try to look up, then check if algo and digest match.
+	if artifactSpec.ID != nil {
+		id64, err := strconv.ParseUint(*artifactSpec.ID, 10, 32)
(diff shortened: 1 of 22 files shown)


pkg/assembler/backends/testing/hashEqual.go
```diff
@@ -17,113 +17,205 @@ package testing
...
+		return nil, gqlerror.Errorf("IngestHashEqual :: Artifact not found")
 	}
-	if len(collectedEqualArt) != 1 {
-		return nil, gqlerror.Errorf(
-			"IngestHashEqual :: multiple artifacts found")
+	artIDs := []uint32{aInt1.id, aInt2.id}
+	sort.Slice(artIDs, func(i, j int) bool { return artIDs[i] < artIDs[j] })
+
+	// Search backedges for existing.
+	searchHEs := slices.Clone(aInt1.hashEquals)
```
lumjjb, line 92: this is pretty cool! didn't know about this!

pkg/assembler/backends/testing/hashEqual.go
```diff
@@ -17,113 +17,205 @@ package testing
...
 }
 
-// Query HashEqual
+func (c *demoClient) matchArtifacts(filter []*model.ArtifactSpec, value []uint32) bool {
+	val := slices.Clone(value)
+	var matchID []uint32
+	var matchPartial []*model.ArtifactSpec
+	for _, aSpec := range filter {
+		a, _ := c.artifactExact(aSpec)
+		// drop error here if ID is bad
```
lumjjb, line 130: not sure what you mean by the comment - can you explain further why we skip error checks here? if it shouldn't happen should we have a check for and panic if it errors?

pkg/assembler/backends/testing/artifact.go
```diff
@@ -17,67 +17,154 @@ package testing
...
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 )
 
-func registerAllArtifacts(client *demoClient) {
-	// strings.ToLower(string(checksum.Algorithm)) + ":" + checksum.Value
-	client.registerArtifact("sha256", "6bbb0da1891646e58eb3e6a63af3a6fc3c8eb5a0d44824cba581d2e14a0450cf")
-	client.registerArtifact("sha1", "7A8F47318E4676DACB0142AFA0B83029CD7BEFD9")
-	client.registerArtifact("sha512", "374AB8F711235830769AA5F0B31CE9B72C5670074B34CB302CDAFE3B606233EE92EE01E298E5701F15CC7087714CD9ABD7DDB838A6E1206B3642DE16D9FC9DD7")
+// Internal data: Artifacts
+type artMap map[string]*artStruct
```
mihaimaruseac, line 31: Is the map keyed on the algorithm or on the digest?

lumjjb: i think it is keyed on both "alg:digest"?

pkg/assembler/backends/testing/hashEqual.go
jeffmendoza: Yes, overall I'm dropping errors here and there. I believe all are from looking up IDs in the index that are already stored on existing nodes. All these should never error. Not sure if we should catch those and explicitly panic, or find a way to bubble the error up and return a gql "internal error" or something similar.  Currently they will probably panic on a pointer dereference later on.

pkg/assembler/backends/testing/artifact.go
jeffmendoza: yep, both.

pkg/assembler/backends/testing/artifact.go
```diff
@@ -17,67 +17,154 @@ package testing
...
+		if err != nil {
+			return nil, fmt.Errorf("couldn't parse id %w", err)
+		}
+		id := uint32(id64)
+		a, err := c.artifactByID(id)
+		if err != nil {
+			// Not found
+			return nil, nil
+		}
+		if algorithm != "" && algorithm != a.algorithm {
```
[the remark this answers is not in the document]
pxp928: yeah I dont think we need to do more matching if the nodeID is provided.

pkg/assembler/backends/testing/hashEqual.go
```diff
@@ -17,113 +17,205 @@ package testing
...
+	justification string
+	origin        string
+	collector     string
 }
 
-// Ingest HashEqual
+func (n *hashEqualStruct) getID() uint32 { return n.id }
 
-func (c *demoClient) registerHashEqual(artifacts []*model.Artifact, justification, origin, collector string) *model.HashEqual {
+func registerAllHashEqual(client *demoClient) {
```
pxp928, line 45: can remove this as we have test in cmd/graphql_playground/cmd/ingest. The only ones we dont have are the objects so like artifact would need it.

pkg/assembler/backends/testing/hashEqual.go
```diff
@@ -17,113 +17,205 @@ package testing
...
+	he := &hashEqualStruct{
+		id:            c.getNextID(),
+		artifacts:     artIDs,
+		justification: hashEqual.Justification,
+		origin:        hashEqual.Origin,
+		collector:     hashEqual.Collector,
+	}
+	c.index[he.id] = he
+	c.hashEquals = append(c.hashEquals, he)
+	aInt1.hashEquals = append(aInt1.hashEquals, he.id)
```
pxp928, line 118: can do a similar helper function like with package and source to append and get.

pkg/assembler/backends/testing/isOccurrence.go
```diff
@@ -17,209 +17,343 @@ package testing
...
-	selectedSourceSpec := &model.SourceSpec{Type: &selectedSourceType, Namespace: &selectedSourceNameSpace, Name: &selectedSourceName, Tag: &selectedTag}
-	selectedSource, err := client.Sources(context.TODO(), selectedSourceSpec)
-	if err != nil {
-		return err
-	}
-	_, err = client.registerIsOccurrence(nil, selectedSource[0], client.artifacts[0], "this artifact is an occurrence of this source", "testing backend", "testing backend")
-	if err != nil {
-		return err
-	}
+	// TODO convert these
```
pxp928, line 50: remove (see comment on `hashEqual`)

pkg/assembler/backends/testing/isOccurrence.go
```diff
@@ -17,209 +17,343 @@ package testing
...
 	}
 
-	for _, occurrence := range c.isOccurrence {
-		if reflect.DeepEqual(occurrence.Artifact, artifact) && occurrence.Justification == justification {
-			if val, ok := occurrence.Subject.(model.Package); ok {
-				if reflect.DeepEqual(val, *selectedPackage) {
-					return occurrence, nil
+	// FIXME these sections are copied from ingestPackage/Source or hasSourceAt
+	packageID := maxUint32
+	if subject.Package != nil {
```
pxp928, line 101: there is a helper function for this now so there is no longer a need to repeat the code.

pkg/assembler/backends/testing/isOccurrence.go
```diff
@@ -17,209 +17,343 @@ package testing
...
 	}
 
-	newIsOccurrence := &model.IsOccurrence{
-		Justification: justification,
-		Artifact:      artifact,
-		Origin:        origin,
-		Collector:     collector,
+	sourceID := maxUint32
+	if subject.Source != nil {
+		source := subject.Source
```
pxp928, line 143: same here

pkg/assembler/backends/testing/artifact.go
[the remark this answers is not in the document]
mihaimaruseac: hmm, my comment didn't save :(

I think it depends on what we want to do in this corner case, but most databases will continue to match and return 0 rows/results on queries such as

```
select x
where x.id == 42 and x.stuff = foo ...
```

(assuming `.stuff` does not match).

There likely is not a performance degradation to continue matching, since we only have exactly one result.

pkg/assembler/backends/testing/isOccurrence.go
```diff
@@ -17,209 +17,343 @@ package testing
...
+		justification: occurrence.Justification,
+		origin:        occurrence.Origin,
+		collector:     occurrence.Collector,
+	}
+	c.index[o.id] = o
+	a.occurrences = append(a.occurrences, o.id)
+	if packageID != maxUint32 {
+		pv, pn, _ := c.pkgByID(packageID)
+		if pv != nil {
+			pv.occurrences = append(pv.occurrences, o.id)
```
pxp928, line 200: can add this to the existing `pkgNameOrVersion` interface and append the back edge

pkg/assembler/backends/testing/isOccurrence.go
```diff
@@ -17,209 +17,343 @@ package testing
...
+		} else {
+			pn.occurrences = append(pn.occurrences, o.id)
+		}
 	} else {
-		newIsOccurrence.Subject = selectedSource
+		s, _ := c.sourceByID(sourceID)
+		s.occurrences = append(s.occurrences, o.id)
```
pxp928, line 206: helper function like the others

pkg/assembler/backends/testing/hashEqual.go
pxp928: yeah that is fine. We can transfer it over to the unit tests.

pkg/assembler/backends/testing/artifact.go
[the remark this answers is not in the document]
jeffmendoza: Updated to not match other fields if ID is provided.

pkg/assembler/backends/testing/hashEqual.go
jeffmendoza: Added #606 606

pkg/assembler/backends/testing/isOccurrence.go
jeffmendoza: done

pkg/assembler/backends/testing/isOccurrence.go
jeffmendoza: done

pkg/assembler/backends/testing/isOccurrence.go
jeffmendoza: done

pkg/assembler/backends/testing/isOccurrence.go
jeffmendoza: The getters seem unnecessary, as the struct members are being freely accessed in general. In fact all the struct members should be capital. If the members were truly private, the getter should just be the capital of the internal member (https://go.dev/doc/effective_go#Getters). The setter helpers should probably be renamed AddFoo.

I'll update to match for constancy here, but we should revisit these.

jeffmendoza: done

jeffmendoza:
updated, PTAL

pkg/assembler/backends/testing/hashEqual.go
```diff
@@ -17,113 +17,201 @@ package testing
...
-	for _, value := range queryArtifacts {
-		queryArt := model.Artifact{
-			Algorithm: strings.ToLower(*value.Algorithm),
-			Digest:    strings.ToLower(*value.Digest),
-		}
-		if _, ok := exists[queryArt]; ok {
-			return true
-		}
+	return &model.HashEqual{
+		ID:            fmt.Sprint(h.id),
```
pxp928, line 211: nit: can use `nodeID` for consistency where assigning IDs (for artifact, isOccurrence and isdependency)

pxp928 (approved):
LGTM small nit

pkg/assembler/backends/testing/hashEqual.go
pxp928: https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/testing/backend.go#L133-L135

PR #601, opened 2023-03-19
added ID and back edge for isDependency

pxp928:
- Added ID fields to `isDependency` and back edges connecting to package/dependent package
- consolidate common code in  `hasSourceAt` and `isDependency`
- remove reused code in `hasSourceAt`
- add helper functions `getSourceIDFromInput` and `getPackageIDFromInput` to remove reused code between verbs
- updated examples

### DIFF
--- a/pkg/assembler/backends/testing/pkg.go
+++ b/pkg/assembler/backends/testing/pkg.go
@@ -94,20 +94,22 @@ type pkgNameStruct struct {
 }
 type pkgNameMap map[string]*pkgVersionStruct
 type pkgVersionStruct struct {
-	id         uint32
-	parent     uint32
-	name       string
-	versions   pkgVersionList
-	srcMapLink []uint32
+	id               uint32
+	parent           uint32
+	name             string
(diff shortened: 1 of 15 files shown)


pkg/assembler/backends/testing/pkg.go
```diff
@@ -409,6 +427,47 @@ func (c *demoClient) buildPackageResponse(id uint32, filter *model.PkgSpec) (*mo
 	return &p, nil
 }
 
+func getPackageIDFromInput(c *demoClient, input model.PkgInputSpec, pkgMatchType model.MatchFlags) (*uint32, error) {
```
mihaimaruseac, line 430: Does the return have to be pointer to ID?

mihaimaruseac (approved):

lumjjb (approved):

pkg/assembler/backends/testing/pkg.go
pxp928: in order to return a nil value returning a pointer.

PR #620, opened 2023-03-20
Add single happy path test for isOccurrence

jeffmendoza:


pkg/assembler/backends/testing/isOccurrence_test.go
```diff
@@ -0,0 +1,163 @@
...
+			},
+			ExpOcc: []*model.IsOccurrence{
+				&model.IsOccurrence{
+					Subject: &model.Package{
+						Type: "pypi",
+						Namespaces: []*model.PackageNamespace{{
+							Names: []*model.PackageName{{
+								Name: "tensorflow",
+								Versions: []*model.PackageVersion{{
+									Qualifiers: []*model.PackageQualifier{},
```
jeffmendoza, line 106: It's weird that with empty versions and qualifiers, we still get back empty slices here instead of nil, is this correct?

jeffmendoza: ex: should this line just be creating a nil slice: https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/testing/pkg.go#L471

mihaimaruseac: Should be correct. We can force the json intermediarry to omit empty fields and then we'd get nil, but then we won't be able to distinguish between `nil` in a filter (meaning select everything) and `""` (meaning select exactly those things where this field is empty).

mihaimaruseac: Oh, I think I was referring to something else. Let's try with empty slice here?

jeffmendoza: Ah, I see it is clarified here: https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/package.graphql#L88 so it all looks correct.

lumjjb (approved):

PR #629, opened 2023-03-22
Expose Path, Neighbors, Packages and Sources query

pxp928:
For `Path` and `Neighbors` queries `fragment allNodes` do not work as it causes an error in the generated code.


mihaimaruseac (approved):

lumjjb (approved):
lgtm

pkg/assembler/clients/operations/path.graphql
```diff
@@ -0,0 +1,150 @@
...
+    }
+    ... on Builder {
+      ...allBuilderTree
+    }
+  }
+}
+
+query Neighbors($node: ID!) {
+  neighbors(node: $node) {
+    __typename
```
lumjjb, line 88: qq: what is the typename here again?

lumjjb (approved):

pkg/assembler/clients/operations/path.graphql
pxp928: that's how graphQL knows how to parse: 	Typename *string `json:"__typename"`

mihaimaruseac: It's the node name, the name from GraphQL (so, Package, Artifact, etc.) This is how you know what type was actually used in the union

PR #634, opened 2023-03-26
Update OSV and Scorecard certifier to new ontology and add unit tests

pxp928:
- update root_package and source query to use graphql
- add unit tests for root_package and source query
- update unit test for OSV certifier
- add `PkgToPurl` helper function with unit tests
- Change certifier commands to use graphql client
- make client functions public
- how to handle OSV that is not CVE or GHSA?
```
cve, ghsa, err := helpers.OSVToGHSACVE(id.VulnerabilityId)
```
as this currently returns an error. Changed to be a log and not create the `isVulnerability`

### DIFF
--- a/pkg/assembler/clients/operations/certifyBad.graphql
+++ b/pkg/assembler/clients/operations/certifyBad.graphql
@@ -19,7 +19,7 @@
 
 mutation CertifyBadPkg($pkg: PkgInputSpec!, $pkgMatchType: MatchFlags, $certifyBad: CertifyBadInputSpec!) {
   ingestPackage(pkg: $pkg) {
-    ...allPkgTree
+    ...AllPkgTree
   }
   ingestCertifyBad(subject: {package: $pkg}, pkgMatchType: $pkgMatchType, certifyBad: $certifyBad) {
     ...allCertifyBad
@@ -37,7 +37,7 @@ mutation CertifyBadSrc($source: SourceInputSpec!, $certifyBad: CertifyBadInputSp
 
 mutation CertifyBadArtifact($artifact: ArtifactInputSpec!, $certifyBad: CertifyBadInputSpec!) {
   ingestArtifact(artifact: $artifact) {
-    ...allArtifactTree
+    ...AllArtifactTree
   }
   ingestCertifyBad(subject: {artifact: $artifact}, certifyBad: $certifyBad) {
     ...allCertifyBad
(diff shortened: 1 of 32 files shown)


pkg/assembler/clients/operations/certifyBad.graphql
```diff
@@ -19,7 +19,7 @@
 
 mutation CertifyBadPkg($pkg: PkgInputSpec!, $pkgMatchType: MatchFlags, $certifyBad: CertifyBadInputSpec!) {
   ingestPackage(pkg: $pkg) {
-    ...allPkgTree
+    ...AllPkgTree
```
lumjjb, line 22: hmm, any particular reason we’re changing this? probably need to change this on the examples side as well?

pkg/assembler/helpers/purl_test.go
```diff
@@ -207,6 +207,328 @@ func TestPurlConvert(t *testing.T) {
...
+			name:            "com.qnx.sdp",
+			version:         "7.0.0.SGA201702151847",
+			subpath:         "",
+			qualifiers:      []string{},
+		}, {
+			// Special OCI case
+			//TODO: similar issue to above.
+
+			//expectedPurlUri: "pkg:oci/debian@sha256%3A244fd47e07d10?repository_url=docker.io/library/debian&arch=amd64&tag=latest",
+			expectedPurlUri: "pkg:oci/docker.io/library/debian@sha256:244fd47e07d10?arch=amd64&tag=latest",
```
lumjjb, line 456: this one is actually pretty tricky, technically the PURL has the repository in the qualifiers (it’s kind of weird, but as long as OSV can recognize it i think its fine we don’t particularly need a 1:1 mapping

pkg/assembler/clients/operations/certifyBad.graphql
pxp928: to make them public so that I can use them in the unit tests. Before they were all private. Yeah I can change the examples also.

pkg/assembler/helpers/purl_test.go
pxp928: Yeah not sure how we want to handle this issue. Seems like the library does it differently from the examples. So which way is right?

pkg/assembler/clients/operations/certifyBad.graphql
pxp928: Seems like we should do it for the rest also as they should be public.

mihaimaruseac (approved):

lumjjb (approved):
lgtm - let's add a TODO with an issue for that conversion in the line around purl conversion.

pkg/assembler/clients/operations/certifyBad.graphql
mihaimaruseac: Sure. I didn't realize that these being package private was an issue. But there's no problem making them public for testing, except that when we decide to reduce the amount of data sent on the wire (e.g., maybe we realize we only need an ID for ingested stuff instead of everything), the test code will break.

pkg/assembler/helpers/purl_test.go
pxp928: I will link the issue.

pxp928: Opened issue #635

PR #636, opened 2023-03-27
Improve query performance for verbs / links.

### LINKED ISSUE #627: Update in-memory queries to search through backedges if subject or object is specified
Currently, we are searching through all possible relationships during a query (ingestion already handles this). If subject/object is specified in the filter spec, determine the back edges they possess and only search through them.

jeffmendoza:
If a subject or object is specified on a verb query that exactly maps to a single node (ex: ID or full details are specified), then only search the backedges of that node for any matching verbs. Code cleanup on these functions for consistancy, and ease of copying. Change byid functions to use a single one with generic types. Standardise on using ParseUint() instead of Atoi().

Fixes #627

### DIFF
--- a/pkg/assembler/backends/inmem/backend.go
+++ b/pkg/assembler/backends/inmem/backend.go
@@ -16,6 +16,7 @@
 package inmem
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"sync/atomic"
@@ -141,4 +142,17 @@ func toLower(filter *string) *string {
 		return &lower
 	}
 	return nil
+}
+
+func byID[E node](id uint32, c *demoClient) (E, error) {
+	var nl E
+	o, ok := c.index[id]
+	if !ok {
+		return nl, errors.New("could not find node")
+	}
+	s, ok := o.(E)
+	if !ok {
+		return nl, fmt.Errorf("not a %T", nl)
+	}
+	return s, nil
 }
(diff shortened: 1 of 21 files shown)


pkg/assembler/backends/inmem/hashEqual.go
```diff
@@ -169,47 +156,72 @@ func (c *demoClient) matchArtifacts(filter []*model.ArtifactSpec, value []uint32
...
+			link, err := byID[*hashEqualStruct](id, c)
+			if err != nil {
+				return nil, gqlerror.Errorf("%v :: %v", funcName, err)
+			}
+			out, err = c.addHEIfMatch(out, filter, link)
+			if err != nil {
+				return nil, gqlerror.Errorf("%v :: %v", funcName, err)
+			}
+		}
+	} else {
```
lumjjb, line 209: hmm if one side was not found, should this return not found?

lumjjb (approved):

pkg/assembler/backends/inmem/backend.go
```diff
@@ -142,3 +143,16 @@ func toLower(filter *string) *string {
 	}
 	return nil
 }
+
+func byID[E node](id uint32, c *demoClient) (E, error) {
```
mihaimaruseac, line 147: Nice!!

mihaimaruseac (approved):

pkg/assembler/backends/inmem/hashEqual.go
jeffmendoza: Yes, if there is an exact match on a subject/object then we don't need a fallback to search all links, only search the backedges and if none match, it will fall through and return an empty list.

PR #646, opened 2023-03-27
Update inmem backend to be concurrency-safe.

### LINKED ISSUE #645: Make inmem backend concurrency-safe
There are many maps used. Also on ingest, we search for an existing match, then add to the relevant maps/lists if not found.

jeffmendoza:
Add a global RW mutex. All reads lock a read mutex. Any ingest will first lock
read, then check if exists. If it does not exist it will release read, lock
write, then check if exists again and insert if it still does not.
Also fixed a bug with the HasSLSA comparison.

Fixes #645

### DIFF
--- a/pkg/assembler/backends/inmem/artifact.go
+++ b/pkg/assembler/backends/inmem/artifact.go
@@ -88,8 +88,8 @@ func (c *demoClient) ingestArtifact(ctx context.Context, artifact *model.Artifac
 	algorithm := strings.ToLower(artifact.Algorithm)
 	digest := strings.ToLower(artifact.Digest)
 
-	c.lock(readOnly)
-	defer c.unlock(readOnly)
+	lock(&c.m, readOnly)
+	defer unlock(&c.m, readOnly)
 
 	a, err := c.artifactByKey(algorithm, digest)
 	if err != nil {
--- a/pkg/assembler/backends/inmem/artifact.go
+++ b/pkg/assembler/backends/inmem/artifact.go
@@ -81,11 +81,24 @@ func (n *artStruct) setCertifyBadLinks(id uint32) { n.badLinks = append(n.badLin
 // Ingest Artifacts
 
 func (c *demoClient) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (*model.Artifact, error) {
+	return c.ingestArtifact(ctx, artifact, true)
+}
+
+func (c *demoClient) ingestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec, readOnly bool) (*model.Artifact, error) {
 	algorithm := strings.ToLower(artifact.Algorithm)
 	digest := strings.ToLower(artifact.Digest)
-	a, err := c.artifactByKey(algorithm, digest)
 
+	c.lock(readOnly)
+	defer c.unlock(readOnly)
+
+	a, err := c.artifactByKey(algorithm, digest)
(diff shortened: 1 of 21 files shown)


jeffmendoza:
Thoughts on the strategy here? Ingest needs to be:

Lock Read
Check if exists
If does: Unlock Read and return
If doesnt...
Unlock Read
Lock Write
Check if exists again
If doesnt, add it
Unlock Write
Return.

I think just making each ingest call itself with a "readonly" bool makes sense.

mihaimaruseac (approved):

lumjjb:
looks good to me!

mihaimaruseac (approved):

pkg/assembler/backends/inmem/artifact.go
```diff
@@ -81,11 +81,24 @@ func (n *artStruct) setCertifyBadLinks(id uint32) { n.badLinks = append(n.badLin
...
 
+	c.lock(readOnly)
+	defer c.unlock(readOnly)
+
+	a, err := c.artifactByKey(algorithm, digest)
 	if err != nil {
+		if readOnly {
+			c.m.RUnlock()
+			a, err := c.ingestArtifact(ctx, artifact, false)
+			c.m.RLock() // relock so that defer unlock does not panic
```
lumjjb, line 99: hmm i think we can do some other type of manipulation that is less costly here than acquiring the lock again, maybe something along the lines of https://stackoverflow.com/questions/50318473/is-there-a-way-to-manipulate-the-defer-stack

pkg/assembler/backends/inmem/artifact.go
```diff
@@ -81,11 +81,24 @@ func (n *artStruct) setCertifyBadLinks(id uint32) { n.badLinks = append(n.badLin
...
 	digest := strings.ToLower(artifact.Digest)
-	a, err := c.artifactByKey(algorithm, digest)
 
+	c.lock(readOnly)
+	defer c.unlock(readOnly)
+
+	a, err := c.artifactByKey(algorithm, digest)
 	if err != nil {
+		if readOnly {
+			c.m.RUnlock()
```
lumjjb, line 97: hmm should we always the same interfaces for lock and unlock? (i.e. c.unlock()) ?

lumjjb (approved):
Couple comments/suggestions then lgtm.

pkg/assembler/backends/inmem/artifact.go
jeffmendoza: I think explicitly calling the read unlock is clearer to a future reader to know what is going on (We know we are "readOnly" so release read lock and call-self to write) ?

pkg/assembler/backends/inmem/artifact.go
lumjjb: this sounds fine for now, but may need to revisit if we run into concurrency perf issues

pkg/assembler/backends/inmem/artifact.go
lumjjb: copying notes from our discussion: 

i'd prefer it to be a helper function that isn't a receiver if possible (like lock(c.m,readOnly)instead).

lumjjb (approved):

pkg/assembler/backends/inmem/artifact.go
jeffmendoza: Done

PR #639, opened 2023-03-27
Graphql changes for consistency

lumjjb:
Made HasSLSA have `Slsa` be compulsory
Made Vex be able to apply to OSV as well
Add misc missing fragments and renames

PR #647, opened 2023-03-28
add osv to CertifyVEXStatement

pxp928:
- add OSV to CertifyVexStatement
- update generated code, ingestion and query
- add OSV testdata for CertifyVexStatement
- add backlink from osv to vex
- update neo4j backend query to include osv 
- fix small bug in `hasSBOM` with source
- fix SLSA to be not null in graphQL Schema


pxp928:
Address PR #639

PR #652, opened 2023-03-29
Add CertifyGood as a copy of CertifyBad

jeffmendoza:


mihaimaruseac (approved):

pkg/assembler/graphql/examples/certify_good.gql
```diff
@@ -0,0 +1,130 @@
+fragment allArtifactTree on Artifact {
```
lumjjb, line 1: should reuse artifact tree fragment? (this will unfortunately cause my gql codegen for UI stuff to fail)

pkg/assembler/graphql/examples/certify_good.gql
```diff
@@ -0,0 +1,130 @@
...
+  id
+  algorithm
+  digest
+}
+
+fragment allCertifyGoodTree on CertifyGood {
```
lumjjb, line 7: do we need to add this to path/neighbors example (if exists)?

lumjjb (approved):
lgtm, minor nits

pkg/assembler/graphql/examples/certify_good.gql
jeffmendoza: Yeah, I don't know what our strategy for examples should be. I copied these in so that I could cut&paste the whole thing into the playground and be able to do an ingestion mutation.

lumjjb: yea i think mainly i'm using the examples for UI stuff, so as long as nothing breaks , doesn't have to be synchronous

PR #655, opened 2023-03-30
Make SPDX parser work properly with DEPENDENCY_OF and CONTAINED_BY 

dejanb:
I come across an sbom that doesn't use CONTAINS relationship, but CONTAINED_BY. This fix should support that use cases. I also added DEPENDENCY_OF for now. From the brief look at the spec, there could be more cases we should support in the future so I refactored code a bit to make it easier to change in the future.
Unfortunately, I can't share the sbom in question so I modified existing test data a bit. I hope that's OK. Let me know if you think there's a better way to handle it.

### DIFF
--- a/internal/testing/testdata/testdata.go
+++ b/internal/testing/testdata/testdata.go
@@ -316,6 +316,9 @@ var (
 	isDepJustifyContains = &model.IsDependencyInputSpec{
 		Justification: "Derived from SPDX CONTAINS relationship",
 	}
+	isDepJustifyContainedBy = &model.IsDependencyInputSpec{
+		Justification: "Derived from SPDX CONTAINED_BY relationship",
+	}
 	isDepJustifyDepends = &model.IsDependencyInputSpec{
 		Justification: "Derived from SPDX DEPENDS_ON relationship",
 	}
@@ -385,7 +388,7 @@ var (
 		{
 			Pkg:          keysPack,
 			DepPkg:       rsaPubFilePack,
-			IsDependency: isDepJustifyContains,
+			IsDependency: isDepJustifyContainedBy,
 		},
 	}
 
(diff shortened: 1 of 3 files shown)


internal/testing/testdata/testdata.go
```diff
@@ -316,6 +316,9 @@ var (
 	isDepJustifyContains = &model.IsDependencyInputSpec{
 		Justification: "Derived from SPDX CONTAINS relationship",
 	}
+	isDepJustifyContainedBy = &model.IsDependencyInputSpec{
```
lumjjb, line 319: do we need one for `isDepJustifyDepends` as well?

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -188,18 +187,23 @@ func (s *spdxParser) GetPredicates(ctx context.Context) *assembler.IngestPredica
...
-		foundPackNodes := s.getPackageElement(string(rel.RefA.ElementRefID))
-		foundFileNodes := s.getFileElement(string(rel.RefA.ElementRefID))
-		relatedPackNodes := s.getPackageElement(string(rel.RefB.ElementRefID))
-		relatedFileNodes := s.getFileElement(string(rel.RefB.ElementRefID))
+		foundPackNodes := s.getPackageElement(foundId)
+		foundFileNodes := s.getFileElement(foundId)
+		relatedPackNodes := s.getPackageElement(relatedId)
+		relatedFileNodes := s.getFileElement(relatedId)
 
 		justification := getJustification(rel)
```
lumjjb, line 208: seems like there should be a change here that i'm not seeing in the PR?

lumjjb (commented):
Awesome - LGTM, couple questions

lumjjb (approved):

internal/testing/testdata/testdata.go
lumjjb: hehe awesome ! Thanks. LGTM!

pxp928 (approved):
LGTM

Issue #592, opened 2023-03-15
Update parsers to emit Identifier strings


lumjjb:
All parsers should implement the emission of [Identifier strings](https://github.com/guacsec/guac/blob/main/pkg/collectsub/collectsub/input/identifier_strings.go) through the [GetIdentifiers](https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/common/types.go#L40) function.

An example of this is implemented in [SLSA](https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/slsa/parser_slsa.go#L272)

jeffmendoza:
I can take this one.

lumjjb:
Awesome ! assigned it to you!

pxp928:
Updated the above issue with the remaining tasks.

arorasoham9:
Is this issue fixed with #817 or needs more attention?

pxp928:
The "Remaining work" is still left to do @arorasoham9. Thanks

PR #658, opened 2023-03-30
add ability to parse identifier strings for PURLs

### LINKED ISSUE #590: Add ability to parse identifier strings for PURLs
With the inclusion of PURL and GitHubRelease additions to collectsub/datasources(#589), we now have the ability to create collect subscription requests for PURLs. [Identifier strings](https://github.com/guacsec/guac/blob/main/pkg/collectsub/collectsub/input/identifier_strings.go) struct should be expanded to include PURLs and guessers should be implemented for unclassified strings to find potential PURLs.

pxp928:
closes #590 

Will work on https://github.com/guacsec/guac/issues/592 next.

### DIFF
--- a/pkg/collectsub/collectsub/input/identifier_strings.go
+++ b/pkg/collectsub/collectsub/input/identifier_strings.go
@@ -50,6 +50,13 @@ func identifierStringsToCollectEntry(i *parser_common.IdentifierStrings) []*pb.C
 		})
 	}
 
+	for _, v := range i.PurlStrings {
+		entries = append(entries, &pb.CollectEntry{
+			Type:  pb.CollectDataType_DATATYPE_PURL,
+			Value: v,
+		})
+	}
+
 	for _, v := range i.UnclassifiedStrings {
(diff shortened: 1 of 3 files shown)


pkg/collectsub/collectsub/input/identifier_strings.go
```diff
@@ -72,5 +79,11 @@ func guessUnknownIdentifierString(s string) (*pb.CollectEntry, error) {
 		}, nil
 	}
 
+	if strings.HasPrefix(s, "pkg:") {
```
lumjjb, line 82: nit: also can use the helper to parse the PURL but i think this should be okay as well

lumjjb (approved):

pkg/collectsub/collectsub/input/identifier_strings.go
pxp928: Yup. I'll add that.

mihaimaruseac (approved):

Issue #653, opened 2023-03-29
Behavior of Neighbors() on various backend nodes


jeffmendoza:
I went through and took a look at what we are doing

| Type             | Root/parent Node              | Neighbors                                                   |
|------------------|-------------------------------|-------------------------------------------------------------|
| ghsa             | no | certVuln, equalVuln, Vex                              |
| cve              | no                  | certVuln, equalVuln, Vex                        |
| osv              | no | certVuln, equalVuln, Vex                              |
| builder          | no                            | hasSLSA links                                               |
| artifact         | no                            | all links                                                   |
| pkgVersion       | pkgName                       | parent(pkgName), all other links                            |
| pkgName          | pkgNamespace                  | parent(pkgNamespace), all version children, all other links |
| pkgNamespace     | pkgType                       | parent(pkgType), all name children                          |
| pkgType          | no                            | all namespace children in this type                         |
| srcName          | srcNamespace                  | parent(srcNamespace), all other links                       |
| srcNamespace     | srcType                       | parent(srcType), all srcName children                       |
| srcType          | no                            | all srcNamespace children                                   |
| **links**         |                             |                                                     |
| certifyVex       | no                            | pkg/art, osv/cve/ghsa                                       |
| certifyVuln      | no                            | pkg, osv/cve/ghsa                                           |
| hasSLSA          | no                            | artifacts, builder                                          |
| hasSourceAt      | no                            | pkg, source                                                 |
| hashEqual        | no                            | artifacts                                                   |
| isDep            | no                            | pkg, depPkg                                                 |
| isOccurrence     | no                            | pkg/source, artifact                                        |
| isVuln           | no                            | osv, cve/ghsa                                               |
| pkgEqual         | no                            | packages                                                    |
| **leaves**         |                             |                                                     |
| certifyBad       | no                            | pkg/src/art                                                 |
| certifyGood      | no                            | pkg/src/art                                                 |
| certifyScorecard | no                            | source                                                      |
| hasSBOM          | no                            | pkg/src                                                     |

Some discrepancies:

* Package and Source have "type" nodes that will form a path between all nodes of that type. The other "nouns" don't have any root connection.

Two unrelated python packages could have a shortest path like so:

```
pkgName -> pkgNamespace -> pkgType[pip] -> pkgNamespace -> pkgName
```

A search might miss a longer dependency chain that links these two.

Current recommendation to find this path would be to do a client-side search and filter the direction that package links are followed.

[...]

PR #663, opened 2023-04-03
Remove root nodes from vulnerabilities

jeffmendoza:
Looking at #653 we don't want "path" to find a path between two vulns through the root node (or year node for cve).


pxp928 (commented):
LGTM

mihaimaruseac (approved):

jeffmendoza:
Added #669 to cover neo4j

pxp928 (approved):
LGTM

PR #673, opened 2023-04-04
Add NoKnownVuln node and related operations.

mihaimaruseac:
This is to signal that when running a vuln certifier a specific package was found to not have any vulnerability.


pkg/assembler/backends/backends.go
```diff
@@ -26,57 +26,59 @@ import (
...
-	PkgEqual(ctx context.Context, pkgEqualSpec *model.PkgEqualSpec) ([]*model.PkgEqual, error)
-	HasSourceAt(ctx context.Context, hasSourceAtSpec *model.HasSourceAtSpec) ([]*model.HasSourceAt, error)
 	CertifyBad(ctx context.Context, certifyBadSpec *model.CertifyBadSpec) ([]*model.CertifyBad, error)
 	CertifyGood(ctx context.Context, certifyGoodSpec *model.CertifyGoodSpec) ([]*model.CertifyGood, error)
-	Scorecards(ctx context.Context, certifyScorecardSpec *model.CertifyScorecardSpec) ([]*model.CertifyScorecard, error)
-	CertifyVuln(ctx context.Context, certifyVulnSpec *model.CertifyVulnSpec) ([]*model.CertifyVuln, error)
-	IsVulnerability(ctx context.Context, isVulnerabilitySpec *model.IsVulnerabilitySpec) ([]*model.IsVulnerability, error)
+	CertifyNoKnownVuln(ctx context.Context, certifyNoKnownVulnSpec *model.CertifyNoKnownVulnSpec) ([]*model.NoKnownVuln, error)
 	CertifyVEXStatement(ctx context.Context, certifyVEXStatementSpec *model.CertifyVEXStatementSpec) ([]*model.CertifyVEXStatement, error)
+	CertifyVuln(ctx context.Context, certifyVulnSpec *model.CertifyVulnSpec) ([]*model.CertifyVuln, error)
```
lumjjb, line 42: hmm, my interpretation would be a "NoVuln" being a type of vuln node and then CertifyVuln with point to that node? It seems fine either way, but i feel like this may help with check for counterfactual

pxp928: Agree. I was thinking similar to having a "NoVuln" node.

mihaimaruseac: Oh, right. I can do those changes.

I was thinking that the client might want to issue separate calls, but this is cleaner

mihaimaruseac: Should be fixed now

pkg/assembler/graphql/schema/certifyVEXStatement.graphql
```diff
@@ -79,22 +93,16 @@ input VexStatementInputSpec {
...
 
 extend type Mutation {
-  "certify that an either a package or artifact has an associated VEX for a CVE, GHSA or OSV"
-  ingestVEXStatement(subject: PackageOrArtifactInput!, vulnerability: OsvCveOrGhsaInput!, vexStatement: VexStatementInputSpec!): CertifyVEXStatement!
+  """
+  Certify that a package or an artifact has an associated VEX for a vulnerability.
+
+  Note that setting `noVuln` in VulnerabilityInput is invalid for VEX statements!
+  """
+  ingestVEXStatement(subject: PackageOrArtifactInput!, vulnerability: VulnerabilityInput!, vexStatement: VexStatementInputSpec!): CertifyVEXStatement!
```
pxp928, line 107: why dont we keep this union type `OsvCveOrGhsaInput` so that it enforces that `noVuln` cannot be entered

pkg/assembler/graphql/schema/certifyVEXStatement.graphql
```diff
@@ -34,33 +32,49 @@ input PackageOrArtifactSpec {
...
+Only one subject type (package or artifact) and one vulnerability type (CVE,
+GHSA or OSV) may be specified.
+
+Note that setting `noVuln` in VulnerabilitySpec is invalid for VEX statements!
 """
 input CertifyVEXStatementSpec {
   id: ID
   subject: PackageOrArtifactSpec
-  vulnerability: OsvCveOrGhsaSpec
+  vulnerability: VulnerabilitySpec
```
pxp928, line 77: why not keep `OsvCveOrGhsaSpec` to avoid wrong inputs?

pxp928 (approved):
LGTM. Not sure we should remove the existing union type.

pkg/assembler/graphql/schema/certifyVEXStatement.graphql
mihaimaruseac: I renamed the struct first and then added the node. I can add back the other struct, but that would result in duplicated validation which is now hidden by the bool flag.

pxp928: Ah I see there is no client-side implementation to ingest via the `noVuln` node so this won't be an issue and you have a check on the query side to make sure it's not set for certifyVEX. Should be good then.

pkg/assembler/graphql/schema/certifyVEXStatement.graphql
pxp928: no need for this

lumjjb (approved):

PR #679, opened 2023-04-05
Add Node query

jeffmendoza:
Node query will take an ID and return the node union type. This can be used if a cliend has an ID, but doesn't know the type.

I want to use this in the `path.py` demo program to simplify the logic some.


mihaimaruseac (approved):

lumjjb (approved):
yes lol i was going to just ask about a feature like this!

PR #681, opened 2023-04-05
Allow filtering neighbors and paths on GUAC verbs

mihaimaruseac:
~~We can expand it if we want to also filter on the other neighbors, it ended up being quite generic.~~

It is now expanded to have all valid combinations of Node-Node.

~~Depends on https://github.com/guacsec/guac/pull/673.~~

### DIFF
--- a/pkg/assembler/backends/inmem/artifact.go
+++ b/pkg/assembler/backends/inmem/artifact.go
@@ -45,22 +45,22 @@ func (n *artStruct) ID() uint32 { return n.id }
 
 func (n *artStruct) Neighbors(allowedEdges edgeMap) []uint32 {
 	out := []uint32{}
-	if allowedEdges[model.EdgeHashEqual] {
+	if allowedEdges[model.EdgeArtifactHashEqual] {
 		out = append(out, n.hashEquals...)
 	}
-	if allowedEdges[model.EdgeIsOccurrence] {
+	if allowedEdges[model.EdgeArtifactIsOccurrence] {
 		out = append(out, n.occurrences...)
 	}
-	if allowedEdges[model.EdgeHasSlsa] {
+	if allowedEdges[model.EdgeArtifactHasSlsa] {
 		out = append(out, n.hasSLSAs...)
 	}
-	if allowedEdges[model.EdgeCertifyVexStatement] {
+	if allowedEdges[model.EdgeArtifactCertifyVexStatement] {
 		out = append(out, n.vexLinks...)
 	}
-	if allowedEdges[model.EdgeCertifyBad] {
+	if allowedEdges[model.EdgeArtifactCertifyBad] {
 		out = append(out, n.badLinks...)
 	}
-	if allowedEdges[model.EdgeCertifyGood] {
+	if allowedEdges[model.EdgeArtifactCertifyGood] {
 		out = append(out, n.goodLinks...)
 	}
 	return out
--- a/pkg/assembler/backends/inmem/artifact.go
+++ b/pkg/assembler/backends/inmem/artifact.go
@@ -44,30 +44,7 @@ type artStruct struct {
 func (n *artStruct) ID() uint32 { return n.id }
 
 func (n *artStruct) Neighbors(allowedEdges edgeMap) []uint32 {
-	maxLen := 0
-	if allowedEdges[model.EdgeHashEqual] {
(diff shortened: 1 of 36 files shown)


pkg/assembler/graphql/schema/path.graphql
```diff
@@ -46,21 +47,55 @@ union Node
...
   """
   neighbors returns all the direct neighbors of a node
 
   Similarly, the input is only specified by its ID.
+
+  Specifying any Edge value in `usingOnly` will make the neighbors list only
+  contain the corresponding GUAC evidence trees (GUAC verbs).
   """
-  neighbors(node: ID!): [Node!]!
+  neighbors(node: ID!, usingOnly: [Edge!]!): [Node!]!
```
mihaimaruseac, line 98: Blocking merge until #673

mihaimaruseac: Resolved

pkg/assembler/graphql/model/nodes.go
```diff
@@ -1230,6 +1230,79 @@ type VulnerabilitySpec struct {
...
+// Each member of the enum matches with one of the verbs. The naming scheme is
+// converting the CamelCase name to CAPITALS_WITH_UNDERSCORES.
+//
+// Note that each GUAC verb is at the same time both an Edge and a Node. We need
+// to return it as a Node to have it show up in the return type of the queries and
+// we need it to be an edge to allow filtering paths to only consider certain
+// predicates.
+type Edge string
+
+const (
```
lumjjb, line 1245: I think we need to be able to define one for "subject" vs "depends_on" for example, so that we can filter on the dependency vs the dependent (when exploring connected component). Wondering if this needs to expand to "HAS_SLSA:SUBJECT" and "HAS_SLSA:BUILDER" for example?

lumjjb: i guess - do we want to do it in this PR or want to merge in one then extend it?

pkg/assembler/backends/inmem/artifact.go
```diff
@@ -43,14 +43,49 @@ type artStruct struct {
...
+	}
+	if allowedEdges[model.EdgeCertifyBad] {
+		maxLen = maxLen + len(n.badLinks)
+	}
+	if allowedEdges[model.EdgeCertifyGood] {
+		maxLen = maxLen + len(n.goodLinks)
+	}
+	if maxLen == 0 {
+		return []uint32{}
+	}
```
jeffmendoza, line 68: Is the maintenance burden of the code to calculate the slice allocation size worth the perf improvement? It was neat when it was just 1 line.

pkg/assembler/graphql/model/nodes.go
mihaimaruseac: So, we don't have edge directions in GraphQL, we can expand on the current enum but we'll have combinatorial explosion.

Backedges in GraphQL SDL are a better solution, but require more work to get in.

pxp928: whats the path forward here?

mihaimaruseac: https://github.com/guacsec/guac/pull/681/commits/510c4371b898aeaf1b70018f8c0aa55efa1c0aa9 is how it will look if we define these to be in the form of `struct_link`.

In fact, this is only half of the work, only from noun to the verb. But we need the reverse too, from the verb to the noun.

If we agree to move forward with this patch, I'll define the others. If we agree that this leads to combinatorial explosion and makes UX worse, then I can revert said commit.

lumjjb: I think for our purposes, we only need the edge name, not so much the direction, so there shouldn't be too much combinatorial explosion? Our edge name definitions implicit declare the filter types.

mihaimaruseac: There is no edge names in GraphQL, we can only define them by linking nodes together. I'll continue on 510c437 then to add the other side (from noun to verb), then

mihaimaruseac: Reverse defined too

pxp928 (approved):
LGTM

lumjjb (approved):
lgtm, like the (node, node) solution

Issue #274, opened 2022-12-09
Regression tests for GUAC POC example


lumjjb:
In order to ensure the new users are able to properly run the demo and get expected results, we should have regression tests to ensure that the code matches up with the demo queries and outputs! 

This issue describes adding a regression test to carry out steps in the [POC](https://github.com/guacsec/guac/blob/main/SETUP.md) to be run as a blocking check for the CI.

This should include:
- Downloading the latest guac-data repository
- Performing the ingestion
- Performing the queries and ensuring that the output is matching

This is a great first issue for those interested!

mihaimaruseac:
These will be very useful for the GraphQL migration / query rewrite work.

lukehinds:
please assign to me @lumjjb , should have something up soon we can chew over.

lumjjb:
Awesome!!! Looking forward to it!!

naveensrinivasan:
@lukehinds, are you working on this? If not, I'd be happy to on this.

lukehinds:
I have something half cooked @naveensrinivasan , I can push it up as a draft PR if you like and you can take what you need?

naveensrinivasan:
> I have something half cooked @naveensrinivasan , I can push it up as a draft PR if you like and you can take what you need?

Great! Thanks.

lumjjb:
This issue is currently on hold due to changes in interfaces, check back soon!

pxp928:
completed via https://github.com/guacsec/guac/pull/685

PR #685, opened 2023-04-06
Add end to end test

jeffmendoza:
Adds test to postmerge GHA workflow. Test sets up graphql_playground, ingests all of guac-data, then runs the certifiers. Next, all the queries from demo/queries.gql are run and compared against expected output.


pxp928 (approved):

pxp928:
closes issue #274

PR #683, opened 2023-04-06
Update vuln parser to handle NoVuln node and update unit tests

pxp928:
Signed-off-by: pxp928 <parth.psu@gmail.com>

PR #687, opened 2023-04-06
Fix some issues discovered with lint.

jeffmendoza:
The linter CI seems to be not working correctly. I'm looking into it, but thought I would push these in.

### DIFF
--- a/pkg/assembler/backends/inmem/backend.go
+++ b/pkg/assembler/backends/inmem/backend.go
@@ -104,7 +104,7 @@ func (n *noKnownVuln) Neighbors() []uint32 {
 }
 
 func (n *noKnownVuln) BuildModelNode(c *demoClient) (model.Node, error) {
-	return &model.NoVuln{nodeID(n.id)}, nil
+	return &model.NoVuln{ID: nodeID(n.id)}, nil
 }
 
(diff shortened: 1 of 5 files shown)


pkg/assembler/backends/inmem/backend.go
```diff
@@ -104,7 +104,7 @@ func (n *noKnownVuln) Neighbors() []uint32 {
 }
 
 func (n *noKnownVuln) BuildModelNode(c *demoClient) (model.Node, error) {
-	return model.NoVuln{nodeID(n.id)}, nil
+	return model.NoVuln{ID: nodeID(n.id)}, nil
```
mihaimaruseac, line 107:
```suggestion
	return &model.NoVuln{ID: nodeID(n.id)}, nil
```

(Parth has the similar change in his PR, to fix type casting)

mihaimaruseac (approved):

pkg/assembler/backends/inmem/backend.go
jeffmendoza: This PR is minor, so lets merge after #683 and I'll rebase

jeffmendoza: rebased and resolved conflict.

pxp928 (approved):

PR #689, opened 2023-04-06
Update linter config. Add type assertion check.

jeffmendoza:
Move to it's own job, I think this was causing caching issues. Remove unecessary exclusion filters, these are unintuitive. Use setting on GHA 'only-new-issues' which is for the PR case.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,7 +90,15 @@ jobs:
         run: go install golang.org/x/tools/cmd/goimports@latest
       - name: Check format
         run: make fmt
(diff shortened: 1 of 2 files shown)


jeffmendoza:
Previous config was not working at all. Lint is reporting errors that have been submitted recently.

testfile
mihaimaruseac, line 1: I think this can be removed?

mihaimaruseac (approved):

testfile
jeffmendoza: got it, thanks! =)

pxp928 (approved):

PR #695, opened 2023-04-09
fix versionRange missing from isDependency

pxp928:
fixes #693

PR #697, opened 2023-04-10
Move E2E test to PR.

jeffmendoza:
E2E has been taking 3-4m. Any objections to putting it in for PRs?


pxp928:
Can you add the fix to this PR: https://github.com/guacsec/guac/pull/695#issuecomment-1501930515

mihaimaruseac (approved):

pxp928 (approved):

lumjjb:
I think until deps.dev api token isnt required the e2e will always fail? We can probably revisit end of week when they GA

pxp928:
> I think until deps.dev api token isnt required the e2e will always fail? We can probably revisit when they GA

deps.dev does not run on the e2e. That is s separate integration test

PR #707, opened 2023-04-11
Improve purl handling in cdx and spdx parsers

dejanb:
This PR fixes TODO left to use 'guac/cdx' prefix instead of the `oci` one for the CyclonDX parser. I also made some small changes to the spdx one. I tried to see if there could any further alignment between two, but failed to see it.

### DIFF
--- a/pkg/ingestor/parser/spdx/parse_spdx.go
+++ b/pkg/ingestor/parser/spdx/parse_spdx.go
@@ -74,6 +74,9 @@ func (s *spdxParser) getTopLevelPackage() error {
 	// TODO: Add CertifyPkg to make a connection from GUAC purl to OCI purl guessed
 	// oci purl: pkg:oci/debian@sha256%3A244fd47e07d10?repository_url=ghcr.io/debian&tag=bullseye
 
+	// TODO (dejanb): do a quick check for SPDX to see if there are any "DESCRIBES" relationships
+	// from the document and omit this top level heuristic package otherwise.
+
 	// Currently create TopLevel package as well in some cases where we guess that the SPDX document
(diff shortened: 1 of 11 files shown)


pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -73,16 +73,10 @@ func (s *spdxParser) Parse(ctx context.Context, doc *processor.Document) error {
...
 
 	// Currently create TopLevel package as well in some cases where we guess that the SPDX document
 	// may not encode it
-	var purl string
-	if len(splitImage) == 3 {
-		purl = "pkg:guac/spdx/" + s.spdxDoc.DocumentName
-	} else if len(splitImage) == 2 {
-		purl = "pkg:guac/spdx/" + s.spdxDoc.DocumentName
-	}
+	var purl string = "pkg:guac/spdx/" + s.spdxDoc.DocumentName
```
lumjjb, line 79: Think perhaps we can do a quick check for SPDX to see if there are any "DESCRIBES" relationships from the document and omit this top level heuristic package otherwise.. Fine with this being a separate issue/PR though

lumjjb (approved):
LGTM Thanks @dejanb . Just a minor comment that we can probably defer to later.

dejanb:
Thanks @lumjjb. I fixed some tests in the meantime, there's still some failure in python example in e2e tests that I'll investigate tomorrow.

I'll add spdx improvement as TODO

dejanb:
All tests are now passing

pxp928 (approved):
LGTM Thanks @dejanb!

PR #709, opened 2023-04-11
update deps.dev to new proto and remove token

pxp928:
removed post-merge workflow as it no longer need an token.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,7 +42,7 @@ require (
 	golang.org/x/text v0.9.0 // indirect
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 	google.golang.org/appengine v1.6.7 // indirect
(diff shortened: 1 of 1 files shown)


pkg/handler/collector/deps_dev/internal/api.proto
```diff
@@ -1,11 +1,24 @@
...
+
 syntax = "proto3";
 
 package deps_dev.v3alpha;
 
-import "pkg/handler/collector/deps_dev/internal/google/api/annotations.proto";
 import "google/protobuf/timestamp.proto";
 
-option go_package = "sos.googlesource.com/sos/proto/api/v3alpha";
+option go_package = "github.com/guacsec/guac/pkg/handler/collector/deps_dev/internal";
```
pxp928, line 21: Is there a way to remove the `proto` file and call it from the upstream repo?

lumjjb (approved):

mlieberman85 (approved):
LGTM

PR #718, opened 2023-04-12
fix nil value error in certifyVuln and deps.dev for parsing GO packages

pxp928:
As I was ingesting the minikube SBOM found some issues in `certifyVuln` ingestion of `noVuln` and deps.dev's handling of "golang" packages as just "go". So you have to switch between them else it does not find anything.


lumjjb (approved):

pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -161,7 +163,7 @@ func (d *depsCollector) fetchDependencies(ctx context.Context, purl string, docC
 	dependenciesReq := &pb.GetDependenciesRequest{
 		VersionKey: &pb.VersionKey{
 			System:  sys,
-			Name:    packageInput.Name,
+			Name:    *packageInput.Namespace + "/" + packageInput.Name,
```
pxp928, line 166: going to clean this up a bit

pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -229,15 +237,15 @@ func (d *depsCollector) fetchDependencies(ctx context.Context, purl string, docC
...
+func (d *depsCollector) collectAdditionalMetadata(ctx context.Context, system, namespace, name, version string, pkgComponent *PackageComponent) error {
 	sys, err := parseSystem(system)
 	if err != nil {
 		return err
 	}
 	versionReq := &pb.GetVersionRequest{
 		VersionKey: &pb.VersionKey{
 			System:  sys,
-			Name:    name,
+			Name:    namespace + "/" + name,
```
lumjjb, line 248: nit may be more robust with trim postfix in case there is already a "/" in namespace (shouldn't be though)

pxp928:
There seems to be an  issue with `GO` package query via the API. I have an issue open https://github.com/google/deps.dev/issues/2

mihaimaruseac (approved):

pkg/handler/collector/deps_dev/deps_dev_test.go
```diff
@@ -68,6 +68,22 @@ func Test_depsCollector_RetrieveArtifacts(t *testing.T) {
 		want:     []*processor.Document{},
 		poll:     false,
 		wantErr:  false,
+	}, {
+		name:     "github.com/PuerkitoBio/purell go package",
+		packages: []string{"pkg:golang/github.com/PuerkitoBio/purell@v1.1.1"},
+		want: []*processor.Document{
+			{
+				Blob:   []byte(testdata.CollectedGoLangMakeNowJust),
```
pxp928, line 76: This test is failing due to not being able to query the source properly

pxp928: fixed

PR #720, opened 2023-04-12
Tutorial for running GUAC with Docker Compose.

jeffmendoza:
Please take a look, especially for technical correctness.
Also, I need a short blurb for the collector-subscriber.


docs/Compose.md
```diff
@@ -0,0 +1,181 @@
+# Running GUAC with Docker Compose
+
+This tutorial introduces the full GUAC ingestion pipeline and how to deploy it
+with the included Docker Compose configuration.
+
+## Demo GUAC Ingestion
+
+The GUAC demonstrations use the full GUAC GraphQL server, but use a demo-level
```
lumjjb, line 8: i think let's skip this line since reading this without context, i got a bit confused which demo this is refering to ( i had assumed it was this demo, which is incorrect)

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+may ingest what you wish to here instead.
+
+```bash
+pushd ../guac-data/docs
+docker run --rm -v $PWD:/data --network guac_default local-organic-guac:latest /opt/guac/collector files /data --natsaddr nats:4222
+popd
+```
+
+Notice how the collectors run and complete quickly. If you have run the
+`guacone files` command in a previous demo, you know that that is a slower
```
lumjjb, line 120: nit: link other demo here

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Start Ingesting Data
+
+Now you can run any GUAC collector and push data to the running ingestion
+pipeline. For example we can ingest the sample `guac-data` data. However, you
+may ingest what you wish to here instead.
+
+```bash
+pushd ../guac-data/docs
+docker run --rm -v $PWD:/data --network guac_default local-organic-guac:latest /opt/guac/collector files /data --natsaddr nats:4222
```
lumjjb, line 115: let's add setup instructions for using `guac-data` repo, or point to an initial repo set up that they should ensure be done before hand. Also would be helpful here to explain parts of the docker command (like the mounting of volumes, etc..

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
+# Running GUAC with Docker Compose
+
+This tutorial introduces the full GUAC ingestion pipeline and how to deploy it
+with the included Docker Compose configuration.
```
lumjjb, line 4: perhaps have a separate repo set up with git clone in a separate md linking to it as pre-requisite

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+make build
+./bin/collector files ../guac-data/docs
+```
+
+or
+
+```bash
+go run ./cmd/collector files ../guac-data/docs
+```
+
```
lumjjb, line 147: would be helpful to add a step here to show that the graphql api is queryable with the data. have some UI screenshots of the GraphQL interface.. and then followed by some placeholder for when we add the UI exploration options (which we will probably link out to another document).

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+## What is running?
+
+Taking a look at the `docker-compose.yaml` we can see what is actually running:
+
+- **Nats**: Nats is used for communication between the ingestion components. It
+  is available on port `4222`.
+
+- **Collector-Subscriber**: This component (help here).
```
lumjjb, line 155: This server helps communicate to the collectors when additional information is needed. For example, when parsing an SBOM, we may encounter identifiers like Package URLs (PURLs). The collector subscriber service allows registration of identifiers of interest to collectors so that they can in turn further augment the dataset.

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+
+- **GraphQL Server**: Serving GUAC GraphQL queries and storing the data. As the
+  in-memory backend is used, no separate backend is needed behind the server.
+
+- **Ingestor**: The ingestor listens for things to ingest through Nats, then
+  pushes to the GraphQL Server. The ingestor also runs the assembler and parser
+  internally.
+
+- **Image Collector**: This collector can pull OCI images from registries for
```
lumjjb, line 164: this pulls OCI image metadata (SBOMs and attestations), not actually the images

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+  further inspection.
+
+- **Deps.dev Collector**: This collector gathers further information from
+  [Deps.dev](https://deps.dev/) for supported packages.
+
+- **OSV Certifier**: This certifier gathers OSV vulnerability information about
+  packages.
```
lumjjb, line 171: nit: should note from `osv.dev`

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+to GUAC will not affect the GraphQL interface that the server provides.
+
+### Ingestion
+
+| Name      | Short Description                                                                 |
+| --------- | --------------------------------------------------------------------------------- |
+| Ingestor  | Takes completed GUAC objects and performs GraphQL mutation queries to add to GUAC |
+| Assembler | Takes parsed document objects and assembles GUAC objects                          |
+| Parser    | Takes documents (ex: SBOMs) and parses the objects inside them                    |
+| Collector | Reads or watches locations for new documents and collects them when found         |
```
lumjjb, line 31: may be helpful to rearrange in order that they run, collect, parse, ingest, assemble .. (although i think we threat what's described here as Ingestor and Assembelr as Assembler, and the parser is the ingestor.

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Demo GUAC Ingestion
+
+The GUAC demonstrations use the full GUAC GraphQL server, but use a demo-level
+all-in-one ingestion command `guacone files`. This command runs the GUAC
+collector, parser, assembler, and ingestor components all at once. These
+components are designed to be run as separate asynchronous services that combine
+to form a robust and scalable ingestion pipeline.
+
+## GUAC Components
```
lumjjb, line 14: I think this section would be super helpful with that architecture diagram actually... let me send you the arch diagram, as well as some of the texts from previous docs

lumjjb (commented):
awesome! shaping up nicely!

docs/Compose.md
jeffmendoza: We do need a full architecture doc. The main thing to focus on here, is why you want to run an ingestion pipeline vs just run `guacone`.

docs/Compose.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+> Note: you may also run the `docker build` command directly without `make`.
+
+## Start GUAC
+
+In another terminal, start up GUAC.
+
+```bash
+docker-compose up
```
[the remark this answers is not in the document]
jeffmendoza: I'd rather let the user know what is going on behind the scenes. We want to give them the knowledge to take what is here, understand it, and run with it. Using make commands covers that up.

docs/Compose.md
jeffmendoza: I rewrote this paragraph, main point is to explain why someone would want to run compose, vs do what is in the demos.

docs/Compose.md
jeffmendoza: done.

docs/Compose.md
jeffmendoza: updated, but shortened. We want to tell people what they are running here, but not have a full architecture dive.

docs/Compose.md
jeffmendoza: updated.

docs/Compose.md
lumjjb: how about a HTTP query to the GQL endpoint just to show that it is working

docs/Compose.md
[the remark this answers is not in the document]
lumjjb: so there's an issue with some of the services where in docker compose it caches the container and this causes there to be issues with the setting up of the services (which may use bad data from previous runs), so at least including that to destroy and start the service

docs/Compose.md
lumjjb: 👍

lumjjb:
I helped to write some text here: https://github.com/jeffmendoza/guac/pull/2

docs/Compose.md
jeffmendoza: done

docs/Compose.md
jeffmendoza: done

jeffmendoza:
Updated PTAL
- Changed from `collector files` to `guacone files`, and removed text about ingestion being faster.
- Removed the name "Ingestion Pipeline" from a few different areas, but not all. Overall want to call this the "full GUAC deployment" and not focus on ingestion as we are using `guacone files` for ingestion due to the bug.
- Added curl to show GraphQL is up and running
- Added arch diagram

lumjjb (approved):
LGTM

docs/Compose.md
jeffmendoza: done.

pxp928 (approved):
LGTM

PR #721, opened 2023-04-12
Update path.py to use the node query for first node.

jeffmendoza:


demo/GraphQL.md
```diff
@@ -582,10 +582,11 @@ were hoping to find, for client-side searching, see the section below...
 The data we have ingested in GUAC is based on the SBOM files in the `guac-data`
 repo, but does not contain any vulnerability information. GUAC has built-in
 "certifiers" which can search the ingested data and attach vulnerability data to
-them. To run the certifiers, run:
+them. The OSV certifier will search for OSV vulnerability information. To run
+the OSV certifiers, run:
 
 ```bash
-go run ./cmd/guacone certifier
+go run ./cmd/guacone osv -p=false
```
lumjjb, line 589: let's encourage folks to use `make` and use bianries instead of `go run` i think there's a tendency sometimes for go run to recompile and users may have a false sense of execution time from running these commands

lumjjb (approved):

demo/GraphQL.md
jeffmendoza: Removed this change, as you have it in your pr.

pxp928 (approved):

pxp928:
@jeffmendoza if you rebase on main it will fix the CI issues.

PR #723, opened 2023-04-13
Inmem backend unit tests

jeffmendoza:
Not all unit tests done. I want to merge in these so far so that I get in the logic fix in `certifyVEXStatement.go` merged.

### DIFF
--- a/pkg/assembler/backends/inmem/cve.go
+++ b/pkg/assembler/backends/inmem/cve.go
@@ -138,11 +138,12 @@ func (c *demoClient) Cve(ctx context.Context, filter *model.CVESpec) ([]*model.C
 		if err != nil {
 			return nil, err
 		}
-		osv, err := c.buildCveResponse(uint32(id), filter)
+		cve, err := c.buildCveResponse(uint32(id), filter)
 		if err != nil {
-			return nil, err
+			// not found
+			return nil, nil
 		}
-		return []*model.Cve{osv}, nil
+		return []*model.Cve{cve}, nil
 	}
 
 	var out []*model.Cve
--- a/pkg/assembler/backends/inmem/cve.go
+++ b/pkg/assembler/backends/inmem/cve.go
@@ -17,6 +17,8 @@ package inmem
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -140,8 +142,11 @@ func (c *demoClient) Cve(ctx context.Context, filter *model.CVESpec) ([]*model.C
 		}
 		cve, err := c.buildCveResponse(uint32(id), filter)
 		if err != nil {
-			// not found
-			return nil, nil
+			if errors.Is(err, errNotFound) {
+				// not found
+				return nil, nil
+			}
+			return nil, err
 		}
 		return []*model.Cve{cve}, nil
 	}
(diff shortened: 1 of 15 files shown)


pkg/assembler/backends/inmem/isOccurrence_test.go
```diff
@@ -133,31 +484,120 @@ func TestOccurrence(t *testing.T) {
...
+					},
+				},
+			},
+			ExpNeighbors: map[string][]string{
+				"2": []string{"2"},
+				"3": []string{"2", "2"},
+				"4": []string{"2", "2"},
+				"5": []string{"2", "7"}, // pkg version
+				"6": []string{"7"},      // artifact
+				"7": []string{"2", "6"}, // isOccurence
```
jeffmendoza, line 558: this is awkward because it is the top id

pkg/assembler/backends/inmem/osv.go
```diff
@@ -135,7 +135,8 @@ func (c *demoClient) Osv(ctx context.Context, filter *model.OSVSpec) ([]*model.O
 		}
 		osv, err := c.buildOsvResponse(uint32(id), filter)
 		if err != nil {
-			return nil, err
+			// not found
+			return nil, nil
```
mihaimaruseac, line 139: This is weird, we are losing the error here.

mihaimaruseac (commented):
I was thinking that we'd test this via the GraphQL API. That way, to test a new backend, all we need to do is start the playground with that backend instead of inmem.

mihaimaruseac:
Not necessarily. We can still keep the same go test table structure, but just send client requests.

pkg/assembler/backends/inmem/osv.go
jeffmendoza: The error here is if the id is not found. The correct thing to do is return an empty response, not error.

jeffmendoza:
The expected IDs (neighbors, and query by ID tests) are implementation-aware. The rest could be acceptance tests as well.

pkg/assembler/backends/inmem/osv.go
mihaimaruseac: We are returning both `nil, err` and `nil, nil` from the function: https://github.com/guacsec/guac/blob/fc490e080b27234538d8bf2afa6b7471fc8322ef/pkg/assembler/backends/inmem/osv.go#L199-L205

I think we actually need to process that error and return nil, nil from there on not found and propagate any other error upstream

jeffmendoza: ByID() returns error if the id is not in the map or is not the type we are looking for. The parseuint error is checked further up (and we now have a unit test for that).

mihaimaruseac: I think we should still report an error if the id does not map to the expected type. Can we make `byID` return that error and `nil` if not found instead of erroring there?

jeffmendoza: > I think we should still report an error if the id does not map to the expected type

I'd expect 

```
{
  packages(pkgSpec: { ID: 10}) {
    ...
  }
}

```
To return an empty list if there is no package with ID 10. That would be the same behavior if you specified a name or other field that didn't match any package.

mihaimaruseac: Yes, but we still need to signal an error if the database is corrupted and we get an unexpected type out of ByID when passing in an ID from a backedge and getting something else.

We can make `ByID` not return an error and ignore the corruption scenario above, or we can add a flag to it to signal when the type mismatch should be treated as not found instead of a data corruption.

jeffmendoza: Right, most calls to ByID expect to find something (backedge), and will throw on error. In this case we know that an error from ByID means not found, and because the ID was part of the query we don't throw, but return an empty result.

mihaimaruseac: Yes, but I saying we should mark it explicit in the API that we want to `ByID` to not error here / that we want this specific error from `buildOSVResponse` to be ignored. Otherwise, future changes in `buildOSVResponse` might raise a new error which we are silently ignoring here when we shouldn't.

jeffmendoza: >  Otherwise, future changes in buildOSVResponse might raise a new error which we are silently ignoring here when we shouldn't.

Yes, I agree this is best practice. We have quite a few of this same pattern, I'll take a look

pkg/assembler/backends/inmem/cve.go
```diff
@@ -138,11 +140,15 @@ func (c *demoClient) Cve(ctx context.Context, filter *model.CVESpec) ([]*model.C
 		if err != nil {
 			return nil, err
 		}
-		osv, err := c.buildCveResponse(uint32(id), filter)
+		cve, err := c.buildCveResponse(uint32(id), filter)
 		if err != nil {
+			if errors.Is(err, errNotFound) {
```
pxp928, line 145: does this need to be done for builder and artifact?

pxp928 (approved):
LGTM, one comment

pkg/assembler/backends/inmem/cve.go
jeffmendoza: They look like this
```go
		b, err := byID[*builderStruct](id, c)
		if err != nil {
			return nil, nil
		}
		return []*model.Builder{c.convBuilder(b)}, nil
```
Calling byID() directly, so the error can be ignored. The worry is where the BuildXResposne is used to search for the node by ID. BuildXResposne could return "not found" error from byID(), or error some other way.

PR #727, opened 2023-04-14
query purl to find all transitive vulnerabilities

pxp928:
Update vulnerability CLI to return all vulnerabilities path (transitive) for a given purl

### DIFF
--- a/cmd/guacone/cmd/query_vulnerability.go
+++ b/cmd/guacone/cmd/query_vulnerability.go
@@ -113,7 +113,7 @@ var queryVulnCmd = &cobra.Command{
 
 			osvResponse, err := model.OSVs(ctx, gqlclient, &model.OSVSpec{OsvId: &opts.vulnerabilityID})
 			if err != nil {
-				logger.Debugf("error querying for osvs: %v", err)
+				logger.Fatalf("error querying for osvs: %v", err)
 			}
 
 			if len(osvResponse.Osv) == 0 {
@@ -123,7 +123,7 @@ var queryVulnCmd = &cobra.Command{
 						logger.Fatalf("error querying for cves: %v", err)
 					}
 					if len(cveResponse.Cve) == 0 {
-						logger.Fatalf("error failed to find CVE or OSV matching vulnerability ID")
+						logger.Debugf("error failed to find CVE or OSV matching vulnerability ID")
 					}
 				} else if strings.HasPrefix(opts.vulnerabilityID, "ghsa") {
 					ghsaResponse, err = model.GHSAs(ctx, gqlclient, &model.GHSASpec{GhsaId: &opts.vulnerabilityID})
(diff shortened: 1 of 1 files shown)


cmd/guacone/cmd/query_vulnerability.go
```diff
@@ -105,72 +106,198 @@ var queryVulnCmd = &cobra.Command{
...
-			if strings.HasPrefix(opts.vulnerabilityID, "cve") {
-				cveResponse, err = model.CVEs(ctx, gqlclient, &model.CVESpec{CveId: &opts.vulnerabilityID})
-				if err != nil {
-					logger.Fatalf("error querying for cves: %v", err)
+			osvResponse, err := model.OSVs(ctx, gqlclient, &model.OSVSpec{OsvId: &opts.vulnerabilityID})
+			if err != nil {
+				logger.Debugf("error querying for osvs: %v", err)
+			}
+
+			if len(osvResponse.Osv) == 0 {
```
pxp928, line 119: need to test this error

pxp928:
> minor nit also, if there are no vuln paths available it still returns a path on just the pkg, but fairly minor.

fixed

pxp928:
> i think we may eventually want to check that every package has a CertifyVuln
Not sure if we need this...that is the job of the certifiers. To go an keep adding `CertifyVuln` to each package.

lumjjb:
yea i was thinking more on the CLI side, where we check to see if the certifier has run (knowing if there is lack of data) so that the user doesn't get false negatives, but otherwise agreed.

cmd/guacone/cmd/query_vulnerability.go
```diff
@@ -105,72 +106,207 @@ var queryVulnCmd = &cobra.Command{
...
 		}
-		logger.Infof("found path %v", strings.Join(path, `,`))
-		logger.Infof("Visualizer url: http://localhost:3000/visualize?path=[%v]", strings.Join(path, `,`))
 	},
 }
 
-func queryCertifyVulnsNeighbors(ctx context.Context, gqlclient graphql.Client, topPkgResponse *model.PackagesResponse, vulnerabilityNodeID string, edgeType model.Edge, depth int) ([]string, error) {
+func queryVulnsViaPackageNeighbors(ctx context.Context, gqlclient graphql.Client, pkgVersionID string, edgeType model.Edge) ([]string, error) {
+	path := []string{}
+	certifyVulns, err := model.Neighbors(ctx, gqlclient, pkgVersionID, []model.Edge{edgeType})
```
pxp928, line 195: > yea i was thinking more on the CLI side, where we check to see if the certifier has run (knowing if there is lack of data) so that the user doesn't get false negatives, but otherwise agreed.

Adding check for this.

pxp928: Added this check

PR #730, opened 2023-04-17
graphql doc review

lumjjb:
Ran through the demos! looks good! Some cleanup and added some notes on difficulties i've had. 

CC: @jeffmendoza

### DIFF
--- a/demo/GraphQL.md
+++ b/demo/GraphQL.md
@@ -30,20 +30,30 @@ The rest of the demo will assume you are in the GUAC directory
 cd guac
 ```
 
+## Building the GUAC binaries
+
+Build the GUAC binaries using the `make` command.
+
+```bash
+make
+```
+
 ## Running the GUAC Server
 
 The GUAC server can be run in different ways. For this demo, we will use the
-`graphql_playground` command, which sets up a GraphQL endpoint and playground,
-and runs an in-memory backend to store the GUAC graph.
+`guacone gql-server` command with the `--debug` flag command, which sets up a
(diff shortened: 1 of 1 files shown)


jeffmendoza (approved):
Looks good. Personally I don't like make for the demo. It's opaque to the user what is going on, and one would assume it is something more complex than what it is. We are just running `go build`. In this case, instead of `go run` we could do `go install ./cmd/...` and then have folks run the commands.

I agree with a `Makefile` for the purposes of having all the devs using the same workflow, and for generating releases with the appropriate flags, but not for the demo. Same thing for the `docker build` and `compose up` in the compose tutorial.

lumjjb:
i lean towards using `make` since it allows the docs to keep a bit more in sync with development, and allows us to correct under the hood without making running the command overly complicated (e.g. if we need extra compilation flags). Since this is a preference, maybe let's get some fuzzy consensus

@mlieberman85 @pxp928 @SantiagoTorres @mihaimaruseac

SantiagoTorres:
+1 on make,  we can perhaps add a target (e.g., test) to spin up the try guacone w/ in-memory backend bits as described in the PR

mihaimaruseac:
I'm also +1 on solution that makes docs stay up to date regardless of how we update the steps. This seems to be `make` now.

pxp928 (approved):

pxp928:
@lumjjb this can be rebased to fix the CI issues

SantiagoTorres (approved):
LGTM!

Issue #719, opened 2023-04-12
Improve CLI UX


jeffmendoza:
How to run the various GUAC pieces is a bit haphazard eg which binary and which subcommand to use.

I propose a rename/shuffle of the commands. Overall the theory is to divide the commands into two categories:

- User-facing: Made to be run by a user / tester to accomplish something. Should be consistent and intuitive CLI UX. One powerful command makes sense for the user to learn.
- Service: Things to be run as a service, ex: in a container, as a cron, as a systemd service, etc. Seperate single-purpose binaries make sense.

Also, the main server is the same between demo and "real" deployments, so just have one way to run that.

**server**:
- name: **guacgql**
- what it does: runs a GraphQL server
- options:
  - backend: inmem or neo4j or future DB
  - backend-specific options: neo4j connection options
  - playground / debug: also start playground
 
**all-in-one (user facing):**
- name: **guacone**
- what it does: performs user-facing operations that require setting up a processor / ingestor / assembler in one binary, only talks to GQL, no nats/csub
- options:
  - gql endpoint
- commands:
  - certify \<type\> - runs the \<type\> certifier once (once by default, optional poll)
  - collector \<type\> - runs the \<type\> collector once, includes "files" (once by default, optional poll)
  - query \<name\> -  runs the canned \<name\>  query.

**services:**

- name: **guaccsub**
- what it does: runs the collector-subscriber service
- options:
  - listening port

- name: **guacingest**
- what it does: runs the ingestor connected to nats and GraphQL
- options:
  - nats addr
  - gql endpoint

- name: **guaccollect**
- what it does: runs the named collector or certifier connected to GraphQL and nats
- options:
  - nats addr
  - gql addr
  - colsub addr
  - collector/certifier name
  - polling options
 
**Other notes:**

Prefix all the binaries with `guac`. Binaries could eventually be installed in `/usr/bin`. Avoid collisions by namespacing, `ingestor` or `collector` are too generic.

**Todo**

From where we are now, I think this is all that is needed.

- [x] Merge `graphql_playground` and `guacone gql-server` to a single `guacgql` command. (I have this done already, will open PR).
- [x] Rename `ingest` to `guacingest`.
- [x] Rename `collector` to `guaccollect`
- [x] Move `guacone csub-server` to `guaccsub`
- [x] Rework the collectors / certifiers in guacone to be subcommands, ex: `certifier <name>` and change to default to once-only

[...]

PR #733, opened 2023-04-17
Remove guacone graphql. Rename playground to guac.

jeffmendoza:
First todo item on #719 

Also removes neo4j args from guacone, some commands were including those in args for some reason. Neo4j args should only be in the `guac` command from now on.


jeffmendoza:
Note: keeping the arg names from `guacone graphql` over the ones from `graphql_playground`

lumjjb:
oh i see... you're taking it out, i was confused with the naming

lumjjb:
Can we rename it to `guac-gql` binary? I think we'd want to ensure that the messaging is that GUAC is a framework and the ingestion pipeline is part of what's in that box which makes the magic work, and not something that we want people to be building into the ingestion API.

pxp928:
+1 this would be more clear.

jeffmendoza:
`guac` -> `guacgql` done.

pxp928 (approved):
LGTM

PR #746, opened 2023-04-19
fix issue with dependent package purl on deps.dev and remove duplicates for vulnerability and hasSourceAt

pxp928:

- Also adds edge filters to certifier queries
- remove vulnerability collection from deps.dev as it ends up with duplicate data from the osv certifier
- check if a source has already been ingested for a purl. If it has, do not re-ingest as it duplicates "hasSourceAt" (different timestamps)

### DIFF
--- a/internal/testing/testdata/testdata.go
+++ b/internal/testing/testdata/testdata.go
@@ -964,8 +964,7 @@ var (
 		   "tag":null,
 		   "type":"git"
 		},
-		"UpdateTime":"2022-11-21T17:45:50.52Z",
-		"Vulnerabilities":null
+		"UpdateTime":"2022-11-21T17:45:50.52Z"
 	 }`
 
 	CollectedForeignTypes = `{
@@ -1057,8 +1056,7 @@ var (
 				 "tag":null,
 				 "type":"git"
 			  },
-			  "UpdateTime":"2022-11-21T17:45:50.52Z",
-			  "Vulnerabilities":null
+			  "UpdateTime":"2022-11-21T17:45:50.52Z"
 		   }
 		],
 		"Scorecard":{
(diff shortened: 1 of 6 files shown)


internal/testing/testdata/testdata.go
```diff
@@ -982,9 +982,7 @@ var (
 			  "CurrentPackage":{
 				 "name":"foreign-types-shared",
 				 "namespace":"",
-				 "qualifiers":[
-					
-				 ],
+				 "qualifiers":null,
```
lumjjb, line 985: do not merge until we resolve issue with duplicates from deps.dev

pxp928: Ready to review! No more duplicates for `hasSourceAt` and `CertifyVuln`

mihaimaruseac (approved):

pkg/ingestor/parser/deps_dev/deps_dev.go
```diff
@@ -182,5 +138,16 @@ func (d *depsDevParser) GetIdentities(ctx context.Context) []common.TrustInforma
 }
 
 func (d *depsDevParser) GetIdentifiers(ctx context.Context) (*common.IdentifierStrings, error) {
-	return nil, fmt.Errorf("not yet implemented")
+	idstrings := &common.IdentifierStrings{}
+	for _, depComp := range d.packComponent.DepPackages {
```
pxp928, line 142: This does cause deps.dev to spin for a long time. Do we want this for the beta release?

pxp928: Future implementation would be that people would connect to the public instance to get this data and not keep going down the chain?

lumjjb (approved):

pkg/ingestor/parser/deps_dev/deps_dev.go
lumjjb: deps.dev can handle pretty big QPS from my knowledge, however, i think based off https://github.com/google/deps.dev/issues/12, it seems like we have all the information we need to construct all that is needed... so if we implement a cache check in the deps.dev collector it should not result in a big hit rate (like for the sourceinput one but we can make it more generic if we have all the data available).

pxp928: Based on discussion, commented out the csub identifiers from deps.dev parser until the issue https://github.com/guacsec/guac/issues/769 is addressed.

PR #760, opened 2023-04-20
Cleaned up guacone CLI opts. Categorized collectors and certifiers.

jeffmendoza:
`pkg/cli/store.go` contains a "FlagStore". All CLI flags are defined here, and the commands build from there by flag name. Commands don't need a "flag" struct, they always just pull from viper with `viper.GetType("name")`

 `pkg/cli/init.go` contains the Config file initialization routine that all commands had duplicated before. We can add more duplication to `pkg/cli` that we find.

`guacone`'s `files` and `image` are moved under `guacone collect`.

`guacone`'s `osv` and `scorecard` are moved under `guacone certifier` and share the `interval` and `poll` options there.

The only top-level `guacone` flag is `gql-endpoint`. The others are underneath the sub command they are for.

**TODO**
I didn't change any flag names, as to not change any existing `guac.yaml` files. We should go over these and think about changing them. Some are `camelCase` and others `dash-case`. Some are "namespaced" (ex: `gql-backend`, `gql-port`) which makes them nicer to group up in the `guac.yaml` and help identify them there without the context of which command they are for.

Also, `poll` default is still true. I want it to be true for `guaccollect` and false for `guacone collect`. Therefore we need different names. I'll think about this and change it soon.

### DIFF
--- a/cmd/guaccollect/cmd/files.go
+++ b/cmd/guaccollect/cmd/files.go
@@ -44,7 +44,7 @@ var filesCmd = &cobra.Command{
 	Short: "take a folder of files and create a GUAC graph utilizing Nats pubsub",
 	Run: func(cmd *cobra.Command, args []string) {
 
-		opts, err := validateFlags(
+		opts, err := validateFilesFlags(
 			viper.GetString("natsaddr"),
 			viper.GetBool("poll"),
 			args)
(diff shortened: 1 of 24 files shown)


lumjjb (commented):
LGTM, there seems to be some places in docs that need to be updated, like https://github.com/guacsec/guac/blob/main/demo/bad_packages.md, can you do a quick grep and see if there's any others that require it?

pxp928 (approved):
LGTM

jeffmendoza:
> can you do a quick grep and see if there's any others that require it?
Rebased, grepped and updated.

pxp928 (approved):

lumjjb (approved):

Issue #749, opened 2023-04-19
deps.dev collector on maven packages?


lumjjb:
It seems for me that I'm not getting any hits on deps.dev collector for maven packages ingested.. Need to validate this.

pxp928:
is the purl coming in as for example `pkg:maven/org.apache.xmlgraphics/batik-anim@1.9.1`?

pxp928:
It is likely related to how they have been stored in deps.dev. Using ":" to separate the namespace and name. For example: `javax.activation:javax.activation-api`

pxp928:
I will make a PR to update and some unit tests for each type of "system" in deps.dev

lumjjb:
relabelled this to data quality issue and keeping it open and remove "bug" label after PR is fixed

gaganhr94:
Looks like this issue is resolved and can be closed

PR #767, opened 2023-04-21
Deps.dev system check

pxp928:
Fix for now for https://github.com/guacsec/guac/issues/749 where the maven does not use "/" but ":" as a separator between the namespace and name.

Add more unit tests to cover each type of system supported by deps.dev.

~Depends on #746~


pkg/ingestor/parser/deps_dev/deps_dev.go
```diff
@@ -182,5 +138,16 @@ func (d *depsDevParser) GetIdentities(ctx context.Context) []common.TrustInforma
 }
 
 func (d *depsDevParser) GetIdentifiers(ctx context.Context) (*common.IdentifierStrings, error) {
-	return nil, fmt.Errorf("not yet implemented")
+	idstrings := &common.IdentifierStrings{}
```
pxp928, line 141: Depends on #746

pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -302,11 +308,20 @@ func (d *depsCollector) collectAdditionalMetadata(ctx context.Context, pkgType s
 
 func getVersionKey(pkgType string, namespace *string, name string, version *string) (*pb.VersionKey, error) {
 	queryName := ""
-	if namespace != nil && *namespace != "" {
-		queryName = strings.TrimSuffix(*namespace, "/") + "/" + name
+	if pkgType != "maven" {
```
lumjjb, line 311: is it worth creating an issue with label "data-quality" to talk about this? or is this specific to deps.dev?

lumjjb (approved):

pkg/handler/collector/deps_dev/deps_dev.go
pxp928: This is specific to deps.dev

pkg/ingestor/parser/deps_dev/deps_dev.go
pxp928: rebased!

PR #771, opened 2023-04-24
add doc for using helm chart

sunnyyip:


docs/HelmChart.md
```diff
@@ -0,0 +1,317 @@
+# Deploying GUAC with Helm Chart
+
+This tutorial introduces the full GUAC components deployment and how to deploy
+it with the [GUAC Helm Chart](https://github.com/kusaridev/helm-charts/tree/main/charts/guac).
```
lumjjb, line 4: We can probably prefix this with "Kusari's third party GUAC Helm Chart". Since technically ownership is not part of this repo, it acts as a way to link and shout out to Kusari's work here as well. thoughts? Or is the idea to donate the helm chart and then make an announcement around that? @sunnyyip @mlieberman85 @trmiller

docs/HelmChart.md
```diff
@@ -0,0 +1,317 @@
...
+info  - Thank you for testing `appDir` please leave your feedback at https://nextjs.link/app-feedback
+warn  - Experimental features are not covered by semver, and may cause unexpected or broken application behavior. Use at your own risk.
+
+{"level":"info","ts":1682341559.5920317,"caller":"cmd/root.go:92","msg":"Using config file: /workspace/guac.yaml"}
+```
+
+#### Ingesting sample data
+We are ingesting the sboms in guac-data/docs dir in this example.
+```bash
+docker run --rm -v $PWD:/data --entrypoint /cnb/process/guacone  ghcr.io/kusaridev/guac:v0.0.1-beta.5 files /data/docs --gql-endpoint http://host.docker.internal:8080/query
```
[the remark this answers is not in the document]
sunnyyip: It's either or in this base between 
1. using a pre-built container image, which the helm chart itself uses as well (current approach)
2. instruct people to clone, build, and find the bin

I'd expect people who picked the helm chart route to be familiar with container/docker run
Also, the Compose.md uses similar command but uses a locally built image from make. We should probably keep them in sync if we were to keep or change this.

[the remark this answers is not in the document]
pxp928: cool. That works.

docs/HelmChart.md
sunnyyip: added kusari to the helm chart link

trmiller (approved):

docs/HelmChart.md
```diff
@@ -0,0 +1,317 @@
...
+to be used, or indication that more data about a software identifier is desired.
+For example, in parsing an SBOM, if it sees the use of a package with a PURL,
+the ingestor creates and entry to the CollectSub service to indicate more
+information about the PURL is desired (via gRPC).
+
+The collectors all subscribe to this service and will automatically retrieve
+more information about the PURL (or other identifier/datasource) if it knows how
+to handle it. For example, the deps.dev collector would know how to handle
+PURLs, and then retrieve more information about the PURL entries created from
+the ingestor parsing the SBOM.
```
lumjjb, line 104: i talked to our TW about this, she recommended deduplicating this and moving it out, so let's leave it out of this document

docs/HelmChart.md
```diff
@@ -0,0 +1,317 @@
+# Deploying GUAC with Helm Chart
+
+This tutorial introduces the full GUAC components deployment and how to deploy
+it with [Kusari's GUAC Helm Chart](https://github.com/kusaridev/helm-charts/tree/main/charts/guac).
```
lumjjb, line 4: can we add a note here, that since this is not part of the repo there may be drift between documentation, and if anything is not working, it is best to refer to the original repository docs.

docs/HelmChart.md
sunnyyip: make sense...we were expecting to externalize the common content out of this doc and Compose.md.

docs/HelmChart.md
sunnyyip: note added

mihaimaruseac (approved):

PR #779, opened 2023-04-25
sign image and upload sbom

sunnyyip:


.github/workflows/release.yaml
```diff
@@ -15,16 +15,24 @@
 name: release-guac-image
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - 'v*'
```
jeffmendoza, line 20: Do we want to create releases manually or have the releases created by pushing tags?

sunnyyip: It looks like a tag is a pre-req of creating a release. I changed the trigger of the image build workflow to react to tag push. So we could potentially test the image prior to cutting a release or build beta images that is not intended for a release.

We could add logic to auto-generate the release if we want. Right now we can still cut the release manually and have the image tag matches it.

jeffmendoza: Let's add a doc (release.md / developer.md ?) with our intended workflow:

1. Push tag
2. verify images
3. Create release in GitHub

... etc.

sunnyyip: Yup, makes sense and we should make a decision on how we want to do 3.

.github/workflows/release.yaml
```diff
@@ -34,16 +42,39 @@ jobs:
...
       - name: Build and publish image
+        id: build_n_publish_image 
         run: |
           #!/usr/bin/env bash
           set -euo pipefail
-          pack build ${IMAGE}:${GITHUB_REF_NAME} --builder ${BUILDER} --buildpack ${BUILDPACK} --publish
+          VERSION=$(git describe --tags --always)
+          COMMIT=$(git rev-parse HEAD)
+          BUILD=$(date +%FT%T%z)
+          PKG=github.com/${{ github.repository }}/pkg/version
```
jeffmendoza, line 59: This needs to match the package name in the [go.mod](https://github.com/guacsec/guac/blob/main/go.mod#L1). If someone forks it and tries to make a release, it won't get the verison/commit/date.

jeffmendoza (approved):

.github/workflows/release.yaml
jeffmendoza: Added #791

.github/workflows/release.yaml
sunnyyip: good point...I'll put the repo name back

PR #782, opened 2023-04-26
add helper function to convert version ranges to constraints

lumjjb:
Since there are libraries that seem to be able to check for a constraint of the form
">=1.0, <6.0" or equivalent, this helper just translates to a set of those constraint ranges. They can then be fed into a library like https://pkg.go.dev/github.com/hashicorp/go-version#section-readme to do the version inclusion.

@mihaimaruseac @pxp928 can you take a quick look at the tests and see if this makes sense.. (trying to get something in between practical and reuse of libraries in both golang and nodejs so we can get this implemented before beta)

### DIFF
--- a/pkg/misc/depversion/depversion_test.go
+++ b/pkg/misc/depversion/depversion_test.go
@@ -1,0 +1,245 @@
+package depversion
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/guacsec/guac/internal/testing/ptrfrom"
+)
+
+func Test_VersionRangeParse(t *testing.T) {
+	testCases := []struct {
+		input  string
+		expect VersionMatchObject
+	}{
+		{
+			// deps.dev test set
+			input: "",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=0"},
+				},
+			},
+		},
+		{
+			input: "1.7.21",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{"=1.7.21"},
+				},
+			},
+		},
+		{
+			input: "3.0.3",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{"=3.0.3"},
+				},
+			},
+		},
+		{
+			input: "<2.0,>=0.12",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{"<2.0,>=0.12"},
+				},
+			},
+		},
+		{
+			input: ">=1.0.0",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=1.0.0"},
+				},
+			},
+		},
+		// not even proper semver... We will conver it to semver
+		{
+			input: ">=1.0.0rc8",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=1.0.0-rc8"},
+				},
+			},
+		},
+		{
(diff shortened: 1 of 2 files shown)


pkg/misc/depversion/depversion_test.go
```diff
@@ -0,0 +1,245 @@
...
+				VRSet: []VersionRange{
+					{">=1,<2"},
+				},
+			},
+		},
+		{
+			input: "^1.0.25",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=1.0.25,<2"},
```
mihaimaruseac, line 102: I think this is `<1.1` instead of `<2`. At least in Python (where they use `~` isntead) and Haskell (where they use `^>=`)

pkg/misc/depversion/depversion_test.go
```diff
@@ -0,0 +1,245 @@
...
+				},
+			},
+		},
+		{
+			input: "^3.0.0 || ^4.0.0",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=3.0.0,<4"},
+					{">=4.0.0,<5"},
```
mihaimaruseac, line 111: Here too, I think it should be `<3.1` and `<4.1` (increase by 1 the component that comes before the last specified component)

pkg/misc/depversion/depversion_test.go
```diff
@@ -0,0 +1,245 @@
...
+				},
+			},
+		},
+
+		// NPM test set
+		{
+			input: "1.0.0 - 2.9999.9999",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{"=1.1.2"},
```
mihaimaruseac, line 138: Hmm, isn't this a range between 1.0.0 and 2.999.9999 (that is `<3`)?

pkg/misc/depversion/depversion_test.go
```diff
@@ -0,0 +1,245 @@
...
+				VRSet: []VersionRange{
+					{">=1.2,<1.3"},
+				},
+			},
+		},
+		{
+			input: "~1.2.3",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=1.2.3,<2"},
```
mihaimaruseac, line 194: `<1.3` for sure

pkg/misc/depversion/depversion_test.go
```diff
@@ -0,0 +1,245 @@
...
+				VRSet: []VersionRange{
+					{"=1.1.2"},
+				},
+			},
+		},
+		{
+			input: ">=1.0.2 <2.1.2",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=1.0.2 <2.1.2"},
```
pxp928, line 146: should there be a comma separator here instead of a space?

pkg/misc/depversion/depversion_test.go
lumjjb: from npm docs, https://github.com/guacsec/guac/pull/782: "Thus, if you've worked with every 1.x version of the host package, use "^1.0" or "1.x" to express this. "....

oof.... :|

lumjjb: i think looks like, if we match exactly just "^" and version or "~" with version then we treat it as npm... and otherwise, look for operators like `~=` matching to other ecosystems.... lack of consistency is :|

pxp928 (approved):
LGTM! We can refine as needed but nice work!

PR #792, opened 2023-04-28
build: improve building in non-docker environments

ctron:
Having podman installed, there are two little pain points. For one, the makefile explicitly checks for a binary named "docker". This change makes this configurable and allows the user to use "make CONTAINER=podman" to override the (still default) docker name.

Second, the dockerfiles don't use fully qualified docker images. Even with docker, this can lead to unexpected results, as it's not guaranteed that you get the images from docker.io. That's why on e.g. Fedora, an interactive chooser shows up, asking the user for confirmation. Or, if the user already did choose as system wide default, one might not get the correct images.

This can easily be solved by using fully qualified image names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,8 @@ LDFLAGS="-X $(PKG).version=$(VERSION) -X $(PKG).commit=$(COMMIT) -X $(PKG).date=
 
 .DEFAULT_GOAL := build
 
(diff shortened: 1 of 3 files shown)


pxp928 (approved):
LGTM! Thanks for the PR @ctron!

mihaimaruseac (approved):

PR #820, opened 2023-05-05
Rename flags

### LINKED ISSUE #809: CLI UX -- Flag names
I propose these guidelines:
- **Consistent name** - For example, the Graph QL address is needed in most commands, the flag name should be the same across all commands.
- **Consistent syle** - whatever it is, make is consistent, I propose `dash-style`, therefore don't use `camelCase` anywhere.
- **Descriptive on its own** - The flag names are also used in the `guac.yaml` config file. Therefore a name should be self descriptive. Good: `nats-addr`, Bad: `type`. If it is something that has the same meaning everywhere, it is ok to be short: ex: `interval`.
- **Namespaced** - If appropriate, group a group of flags that go together with a prefix, ex: `neo4j`.
- **Short versions** - Service oriented flags don't need short versions, user oriented flags should.
- **User oriented bools default false** - Name bools so that the default is false. Because long names must be descriptive and possibly namespaced (for the config file), they are more cumbersome to type. You can't set a bool to false with the short version, only with long (`--long-name=false`), so default to false, and the short version can be used to enable it.
- **Required args** - required args should be positional. Options should be optional with good defaults.

All the current flags are in https://github.com/guacsec/guac/blob/main/pkg/cli/store.go This helps with the consistant name, and to make sure the flags are not used for different meanings in two places.

Proposed new flag names:

### guacgql - service so no short opts
neo4j-addr - old: gdbaddr
neo4j-user - old: gdbuser
neo4j-pass - old: gdbpass
neo4j-realm - old: realm
gql-backend
gql-listen-port - old: gql-port, change to be consistent with csub-listen-port
gql-debug
gql-test-data - old: gql-testdata

### guaccsub - service so no short opts
csub-listen-port

### multiple - likely use default or set in guac.yaml, no short opts
gql-addr old: gql-endpoint, change to be consistent with nats-addr and csub-addr
nats-addr old: natsaddr, adopt dash-style
csub-addr

### guaccollect - service so no short opts
use-csub
service-poll - old: poll, make separate from user oriented poll in guacone. default true for service mode

### guacone collect files, these are rare? so no short opts
verifier-key-path - old: verifier-keyPath
verifier-key-id - old: verifier-keyID

### guacone collect/certifier , yes short opts
poll - p - default false for user mode
interval - i

### guacone certify, user cmd so yes short opts
cert-bad - b - old: good - flip to bad to default false (old good was default true)
node-type - t - old: type - Question: is this required and should be positional?
justification - j
package-name - n - old: pkgName

### guacone query, user cmd so yes short opts
vuln-id - v - old: vulnerabilityID
search-depth - d - old: depth
num-path - n - old: path
Move purl to be positional.

jeffmendoza:
# Description of the PR

Fixes #809

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/deps_dev.go
+++ b/cmd/guaccollect/cmd/deps_dev.go
@@ -50,10 +50,10 @@ var depsDevCmd = &cobra.Command{
 		logger := logging.FromContext(ctx)
 
 		opts, err := validateDepsDevFlags(
-			viper.GetString("natsaddr"),
+			viper.GetString("nats-addr"),
 			viper.GetString("csub-addr"),
 			viper.GetBool("use-csub"),
-			viper.GetBool("poll"),
+			viper.GetBool("service-poll"),
 			args)
 		if err != nil {
 			fmt.Printf("unable to validate flags: %v\n", err)
--- a/cmd/guaccollect/cmd/files.go
+++ b/cmd/guaccollect/cmd/files.go
@@ -48,8 +48,8 @@ var filesCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 
 		opts, err := validateFilesFlags(
-			viper.GetString("natsaddr"),
-			viper.GetBool("poll"),
+			viper.GetString("nats-addr"),
+			viper.GetBool("service-poll"),
 			args)
 		if err != nil {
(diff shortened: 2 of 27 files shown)


mihaimaruseac (approved):

pxp928 (approved):
LGTM! This is much more consistent. Thanks @jeffmendoza

PR #855, opened 2023-05-18
fix version handling for top level and dependency components

pxp928:
# Description of the PR

fix version handling for top level and dependency components for both file and container types. Making it uniform for both top level and dependency components

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
+++ b/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
@@ -266,43 +266,45 @@ func (s *cyclonedxParser) getPackageElement(elementID string) []*model.PkgInputS
 }
 
 func guacCDXFilePurl(fileName string, version string, topLevel bool) string {
+	escapedName := asmhelpers.SanitizeString(fileName)
 	if topLevel {
 		if version != "" {
 			splitVersion := strings.Split(version, ":")
 			if len(splitVersion) == 2 {
 				s := fmt.Sprintf(topCdxPurlGuac+"%s:%s", strings.ToLower(splitVersion[0]), splitVersion[1])
-				s += fmt.Sprintf("#%s", fileName)
+				s += fmt.Sprintf("#%s", escapedName)
 				return s
 			}
 		}
-		return topCdxPurlGuac + fileName
+		return topCdxPurlGuac + escapedName
 	} else {
 		if version != "" {
 			splitVersion := strings.Split(version, ":")
 			if len(splitVersion) == 2 {
-				return asmhelpers.GuacFilePurl(splitVersion[0], splitVersion[1], &fileName)
+				return asmhelpers.GuacFilePurl(splitVersion[0], splitVersion[1], &escapedName)
 			}
 		}
-		return asmhelpers.PurlFilesGuac + fileName
+		return asmhelpers.PurlFilesGuac + escapedName
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -257,25 +265,44 @@ func (s *cyclonedxParser) getPackageElement(elementID string) []*model.PkgInputS
...
+			if len(splitVersion) == 2 {
+				return asmhelpers.GuacFilePurl(splitVersion[0], splitVersion[1], &fileName)
+			}
+		}
 		return asmhelpers.PurlFilesGuac + fileName
 	}
 }
 
-func guacCDXPkgPurl(componentName string, version string, tag string) string {
+func guacCDXPkgPurl(componentName string, version string, tag string, topLevel bool) string {
```
lumjjb, line 290: nit: would it help to have an additional sanitization here to run a HTML encoding here in each value here so that we don't run into the edge cases where a special character is included

This should make it so that it runs into less issues in the future following: https://github.com/package-url/purl-spec/blob/master/PURL-SPECIFICATION.rst#character-encoding

>the '#', '?', '@' and ':' characters must NOT be encoded when used as separators. They may need to be encoded elsewhere

lumjjb (approved):
minor nit; else lgtm

mihaimaruseac (approved):

pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
pxp928: Hmmm...in order to do that, it would make sense to use the packeurl-go function to create the purl for us. This does handle the encoding piece. In that case, we should also do it on the SPDX side:

https://github.com/guacsec/guac/blob/08f656c1b0258c16452d80d2efc2dccf4a5f8777/pkg/assembler/helpers/purl.go#L42-L63

pxp928: added `escapedName := url.QueryEscape(componentName)` is this what you were thinking?

pxp928: Actually, it might make sense to create the purl from the `packageurl-go` library.

https://github.com/guacsec/guac/blob/08f656c1b0258c16452d80d2efc2dccf4a5f8777/pkg/assembler/helpers/purl.go#L42-L63

pxp928: pass in the type, namespace, name...etc and it does the encoding for us based on the purl spec. This would also be have to be done on the SPDX side also.

lumjjb: yea - agree, that would make sense here to do it here!

pxp928: Thinking through this more. This might be the better approach:

```
if p.name != "" {
		var ns []string
		for _, item := range strings.Split(p.name, "/") {
			ns = append(ns, url.QueryEscape(item))
		}
		purl = purl + strings.Join(ns, "/") + "/"
	}
```

pxp928: @lumjjb updated, PTAL

lumjjb: LGTM

PR #857, opened 2023-05-18
Modify getTopLevelPackage to check for DESCRIBES/DESCRIBED_BY

### LINKED ISSUE #708: SPDX parser top level heuristic change to check "DESCRIBES" relationship
The SPDX top level purl construction for `guac:spdx/...` is a heuristic for documents that don't properly include DESCRIBES relationships to talk about what package they are describing. We should add some logic in the SPDX parser to not define this top level package otherwise and just walk the tree from the packages that the document describes.

https://github.com/guacsec/guac/pull/707/files/78907de874199482af3549719a1e25f2735db0b9#r1162839713

CrossedSecurity:
relationship when populating pUrl

# Description of the PR

getTopLevelPackage will now check DESCRIBES and DESCRIBED_BY relationships to populate the pUrl. It will fall back to the original method of generating pUrl if neither are available. Added test cases for both of these options.

Fixes #708

# PR Checklist

- [x ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ PKG=github.com/guacsec/guac/pkg/version
 LDFLAGS="-X $(PKG).Version=$(VERSION) -X $(PKG).Commit=$(COMMIT) -X $(PKG).Date=$(BUILD)"
 
 CONTAINER ?= docker
-CPUTYPE=$(shell uname -m)
+CPUTYPE=$(shell uname -m | sed 's/x86_64/amd64/')
 GITHUB_REPOSITORY ?= guacsec/guac
 LOCAL_IMAGE_NAME ?= local-organic-guac
 


lumjjb (commented):
Thanks for the PR @CrossedSecurity !! Mostly looks good, there are some edge cases that need to be handled which touches some other parts of the code - i've added some comments! Do let me know if you need more details and happy to chat through them! Have a good weekend!!

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -74,21 +76,51 @@ func (s *spdxParser) getTopLevelPackage() error {
...
+				for _, ext := range pac.PackageExternalReferences {
+					if ext.RefType == spdx_common.TypePackageManagerPURL {
+						pUrl = ext.Locator
+					}
+				}
+			}
+		}
+
+		// Construct a default pUrl if none exist for the id.
+		pUrl = "pkg:guac/spdx/" + id
```
lumjjb, line 114: hmm looks like this should be before the for loop? Also, i think we can use [GuacPkgPurl](https://github.com/guacsec/guac/blob/main/pkg/assembler/helpers/purl.go#LL178C2-L178C2) here

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -74,21 +76,51 @@ func (s *spdxParser) getTopLevelPackage() error {
...
+	if len(spdxIds) == 0 {
+		// If there is no relationship that describes the document, we take a best guess for the pUrl.
+		spdxIds = append(spdxIds, "pkg:guac/spdx/"+s.spdxDoc.DocumentName)
+		s.topLevelIsHeuristic = true
+	}
 
-	if purl != "" {
-		topPackage, err := asmhelpers.PurlToPkg(purl)
+	for _, id := range spdxIds {
+		// Check if we can find a pUrl from a package before constructing a default one.
```
lumjjb, line 101: i think this should be handled by the logic later on in the parsing of the document. So it's not needed here i believe, instead to get the pkgInputSpec you can retrieve it via `packagePackages[spdxid]`.

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -181,11 +213,18 @@ func (s *spdxParser) GetPredicates(ctx context.Context) *assembler.IngestPredica
...
 
-	toplevel := s.getPackageElement("DOCUMENT")
+	topLevel := s.getPackageElement("DOCUMENT")
 	// adding top level package edge manually for all depends on package
-	if toplevel != nil {
-		preds.IsDependency = append(preds.IsDependency, common.CreateTopLevelIsDeps(toplevel[0], s.packagePackages, s.filePackages, "top-level package GUAC heuristic connecting to each file/package")...)
-		preds.HasSBOM = append(preds.HasSBOM, common.CreateTopLevelHasSBOM(toplevel[0], s.doc))
+	if topLevel != nil {
+		for _, topLevelPkg := range topLevel {
+			preds.HasSBOM = append(preds.HasSBOM, common.CreateTopLevelHasSBOM(topLevelPkg, s.doc))
```
lumjjb, line 220: nit : paired with the comment of just needing the spdx ids, we can retrieve the topLevelPkg from packagesPackages

lumjjb (commented):
minor comments, else LGTM!

pkg/ingestor/parser/spdx/parse_spdx.go
CrossedSecurity: note: only need this if we use heuristic, otherwise we can look up in packagePackages

CrossedSecurity:
Not sure why this force pushed main in guacsec. Probably a bad git command - it should need to be approved before push though?

lumjjb:
don't think it force pushed main - just your branch? don't see it in main branch

PR #873, opened 2023-05-19
Fix duplicate vulnerability entries due to different attestation times

dejanb:
# Description of the PR

There is an issue when using osv certifier that produces duplicate "certify vulnerability" entries.
There could be more issues related to this, but one that I have identified so far is that every attestation is created with it's own time instead of the unique scan time. That prevents ingest logic to properly identify it as duplicate.

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/inmem/certifyVuln.go
+++ b/pkg/assembler/backends/inmem/certifyVuln.go
@@ -165,7 +165,7 @@ func (c *demoClient) ingestVulnerability(ctx context.Context, packageArg model.P
 		if noKnownVulnID != 0 && noKnownVulnID == v.noKnownVulnID {
 			vulnMatch = true
 		}
-		if vulnMatch && packageID == v.packageID && certifyVuln.TimeScanned.UTC() == v.timeScanned && certifyVuln.DbURI == v.dbURI &&
+		if vulnMatch && packageID == v.packageID && certifyVuln.TimeScanned.Equal(v.timeScanned) && certifyVuln.DbURI == v.dbURI &&
 			certifyVuln.DbVersion == v.dbVersion && certifyVuln.ScannerURI == v.scannerURI && certifyVuln.ScannerVersion == v.scannerVersion &&
 			certifyVuln.Origin == v.origin && certifyVuln.Collector == v.collector {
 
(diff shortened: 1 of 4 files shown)


mihaimaruseac (approved):
We probably should compare time as we compare floats, but for now this is good.

pkg/assembler/backends/inmem/certifyVuln.go
```diff
@@ -165,7 +165,7 @@ func (c *demoClient) ingestVulnerability(ctx context.Context, packageArg model.P
 		if noKnownVulnID != 0 && noKnownVulnID == v.noKnownVulnID {
 			vulnMatch = true
 		}
-		if vulnMatch && packageID == v.packageID && certifyVuln.TimeScanned.UTC() == v.timeScanned && certifyVuln.DbURI == v.dbURI &&
+		if vulnMatch && packageID == v.packageID && certifyVuln.TimeScanned.Equal(v.timeScanned) && certifyVuln.DbURI == v.dbURI &&
```
pxp928, line 168: hmmm..I don't understand why there are duplicate nodes in the first place. No matter what the timestamp says, there should just be one `certifyVuln` node per package version. The OSV certifier has the ability to check the timestamp and create a new `certifyVuln` node if that is out of date (based on user input), but that is disabled by default.

pxp928 (approved):
LGTM

pkg/assembler/backends/inmem/certifyVuln.go
dejanb: @pxp928 As far as I saw, osv certifier uses `PackageNode`s which have multiple instances for a purl (with different hashes). I'm not sure if this is correct or not, but it ends up in trying to ingest vuln multiple times. I think there's more room in improving this flow.

pxp928: So it checks at the package version level (which should be unique). So each package version should only have one `certifyVuln` and not be ingested multiple times.

dejanb: @pxp928 I think the issue originates somewhere here

https://github.com/guacsec/guac/blob/main/pkg/certifier/osv/osv.go#L84

As there could be multiple nodes for the purl, so the certifier will create multiple documents. I can dig deeper into this next week.

pxp928: ah yes I see, we could remove the need to do it for each hash. As it's not needed.

PR #877, opened 2023-05-19
make startedOn and finishedOn optional fields for hasSLSA

pxp928:
# Description of the PR

make startedOn and finishedOn optional fields for hasSLSA

# PR Checklist

### DIFF
--- a/cmd/guacgql/cmd/ingest.go
+++ b/cmd/guacgql/cmd/ingest.go
@@ -91,7 +91,8 @@ func ingestScorecards(ctx context.Context, client graphql.Client) {
 
 func ingestSLSA(ctx context.Context, client graphql.Client) {
 	logger := logging.FromContext(ctx)
-
+	startTime := time.Now()
+	finishTime := time.Now().Add(10 * time.Second)
 	predicate := []model.SLSAPredicateInputSpec{
 		{
 			Key:   "buildDefinition.externalParameters.repository",
@@ -124,8 +125,8 @@ func ingestSLSA(ctx context.Context, client graphql.Client) {
 		BuildType:     "Test:SLSA",
 		SlsaPredicate: predicate,
 		SlsaVersion:   "v1",
-		StartedOn:     time.Now(),
-		FinishedOn:    time.Now().Add(10 * time.Second),
+		StartedOn:     &startTime,
+		FinishedOn:    &finishTime,
 		Origin:        "Demo ingestion",
 		Collector:     "Demo ingestion",
 	}
@@ -155,8 +156,8 @@ func ingestSLSA(ctx context.Context, client graphql.Client) {
 		BuildType:     "Test:SLSA2",
 		SlsaPredicate: predicate,
 		SlsaVersion:   "v1",
-		StartedOn:     time.Now(),
-		FinishedOn:    time.Now().Add(10 * time.Second),
+		StartedOn:     &startTime,
+		FinishedOn:    &finishTime,
 		Origin:        "Demo ingestion",
 		Collector:     "Demo ingestion",
 	}
--- a/cmd/guacgql/cmd/ingest.go
+++ b/cmd/guacgql/cmd/ingest.go
@@ -156,8 +156,6 @@ func ingestSLSA(ctx context.Context, client graphql.Client) {
 		BuildType:     "Test:SLSA2",
 		SlsaPredicate: predicate,
 		SlsaVersion:   "v1",
-		StartedOn:     &startTime,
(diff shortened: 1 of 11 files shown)


cmd/guacgql/cmd/ingest.go
```diff
@@ -155,8 +156,8 @@ func ingestSLSA(ctx context.Context, client graphql.Client) {
 		BuildType:     "Test:SLSA2",
 		SlsaPredicate: predicate,
 		SlsaVersion:   "v1",
-		StartedOn:     time.Now(),
-		FinishedOn:    time.Now().Add(10 * time.Second),
+		StartedOn:     &startTime,
+		FinishedOn:    &finishTime,
```
mihaimaruseac, line 160: Let's remove them from one of the fields, this way we'd be notified when we accidentally make these fields mandatory again

pkg/assembler/graphql/schema/hasSLSA.graphql
```diff
@@ -115,8 +115,8 @@ input SLSAInputSpec {
   buildType: String!
   slsaPredicate: [SLSAPredicateInputSpec!]!
   slsaVersion: String!
-  startedOn: Time!
-  finishedOn: Time!
+  startedOn: Time
```
mihaimaruseac, line 118: Nit: trailing whitespace

mihaimaruseac (approved):

pkg/assembler/backends/inmem/hasSLSA.go
```diff
@@ -305,8 +305,8 @@ func (c *demoClient) convSLSA(in *hasSLSAStruct) (*model.HasSlsa, error) {
 			BuildType:     in.buildType,
 			SlsaPredicate: in.predicates,
 			SlsaVersion:   in.version,
-			StartedOn:     in.start,
-			FinishedOn:    in.finish,
+			StartedOn:     &in.start,
+			FinishedOn:    &in.finish,
```
jeffmendoza, line 309: Shouldn't these be set to nil if the start/finish time are 0?

cmd/guacgql/cmd/ingest.go
pxp928: done. Removed it from one test and left both the start and end time for the other.

pkg/assembler/graphql/schema/hasSLSA.graphql
pxp928: done

pkg/assembler/backends/inmem/hasSLSA_test.go
```diff
@@ -78,9 +79,11 @@ func TestHasSLSA(t *testing.T) {
 				{
 					Subject: a1out,
 					Slsa: &model.Slsa{
-						BuiltBy:   b1out,
-						BuiltFrom: []*model.Artifact{a2out},
-						BuildType: "test type",
+						BuiltBy:    b1out,
+						BuiltFrom:  []*model.Artifact{a2out},
+						BuildType:  "test type",
+						StartedOn:  &time.Time{},
```
jeffmendoza, line 85: Yeah, we shouldn't need these in the expected output if they are nil in the input.

pxp928: they are not nil. Changing them to:

```
func nilToZeroValue(inputTime *time.Time) time.Time {
	if inputTime == nil {
		return time.Time{}
	}
	return *inputTime
}
```

pkg/assembler/backends/inmem/hasSLSA.go
pxp928: storing them as time.Time{}

pxp928: similar to how we handle the version being nil

jeffmendoza: If the times are optional we need to set them to nil on the output (model.HasSlsa) if they were nil coming in (model.SLSAInputSpec)

mihaimaruseac: Yeah, these are leaf types, so they should be stored as input.

For version, namespace, etc., we did a canonicalization to empty string because those were intermediate nodes and reference keys, so we couldn't use nil.

jeffmendoza (approved):

pkg/assembler/backends/inmem/hasSLSA.go
pxp928: Thanks for fixing this up @jeffmendoza! I was thinking to match the others. It makes more sense to keep it nil.

Issue #169, opened 2022-10-20
Metadata document Quality Thread


lumjjb:
This issue keeps track of data quality and parsing metadata documents. If there are issues processing certain documents, do let us know and provide an example document that we can test with.

1. Type of Document:
2. Exhibited behavior (error, missing info, etc.):
3. Expected behavior:
4. Link to metadata document:
5. Link to relevant issues:

electricgull:
1. CycloneDX
2. segfault when running against a CycloneDx 1.3 bom
3. no error and ingests the data.
4. https://github.com/JupiterOne/graph-github/blob/guac-bom-test-cg/reports/bom.json

How to reproduce 

git clone git@github.com:JupiterOne/graph-github.git
docker run --rm -v ${PWD}/src jupiterone/node-cdx-bom:latest '/src/reports/bom.json'
../guacsec/guac/bin/guacone files --creds neo4j:s3cr3t reports/

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x68 pc=0x1978521]

Let me know if additional information is needed.

nadgowdas:
thanks for reporting, let me check!

nchelluri:
1. SPDX
2. If a `relationships` value is null, then the upstream SPDX parser will panic
3. The upstream parser should not panic, and the relationship in question would not be created
4. [spdx-null-rel.json](https://github.com/guacsec/guac/files/15099756/spdx-null-rel.json)
5. Saw this when implementing a fix for https://github.com/guacsec/guac/issues/1821

PR #603, opened 2023-03-20
Update CDX parser to generate GraphQL nodes (#584)

dejanb:
Rewrote cdx parser to work with new ontology:

* Tests have been adapted
* Except `Test_addEdgesRecursive` which seems to me is not relevant anymore. Let me know if you see it differently.
* I tried to carry on some of the improvements from #390 
* More improvements and usage of helpers functions can be done later on

Issue #199, opened 2022-11-03
[BUG] CycloneDX ingestion segfault


lumjjb:
More details:

https://github.com/guacsec/guac/issues/169#issuecomment-1302558947 (opened by @electricgull)

lumjjb:
FYI: @nadgowdas

nadgowdas:
This is an interesting case. For SBOMs we are expecting expecting the identify root component (which is typically an image) and is captured in an SBOM (atleast the ones that are produced by `syft`). 
The SBOM in this case, seem to have been generated from package.json (@electricgull  can you share how this SBOM was generated ?) and does not capture root component in the metadata.

Current SBOM: https://github.com/JupiterOne/graph-github/blob/guac-bom-test-cg/reports/bom.json#L6-L11
Syft Image SBOM: https://github.com/guacsec/guac/blob/main/internal/testing/processor/testdata/alpine-cyclonedx.json#L6-L20

We need to support all these SBOMs variations. In this case, the question is what should be the root component ? 

We can add SBOM file with its sha as the root element. Thought ? @lumjjb @pxp928

electricgull:
We are using this docker container created from here:
https://github.com/JupiterOne/node-cdx-bom

https://github.com/JupiterOne/node-cdx-bom/blob/main/Dockerfile
https://github.com/JupiterOne/node-cdx-bom/blob/main/bin/node-cdx-bom

lumjjb:
I think for us to create the edges, we would need the top level component. I think in this case, a missing [top level "component"](https://github.com/CycloneDX/specification/blob/1.4/schema/bom-1.4.schema.json#L136-L140) would probably mean that we would only create nodes for each individual component but not link it to any top level component so it doesn't exists...

@stevespringett do you have any recommendations for CDX here?

PR #896, opened 2023-05-30
Fix crash when parsing valid CycloneDX formats missing optional metadata fields

kurt-r2c:
# Description of the PR

Fixes https://github.com/guacsec/guac/issues/169.

As of 0d557cbe66b42053e72d5f25c1bba94c905dcede, `guac` still crashes on the example from #169 (referenced also in #199): https://github.com/JupiterOne/graph-github/blob/43fa5c43078eb16fad904fd1567133f9c296ea7f/reports/bom.json#L6-L11

image

The root cause is this line of code: https://github.com/guacsec/guac/blob/cb2b6fe5f6449d5225f1e3f55ad5f2d4741dc5d5/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go#L205

The CycloneDX specification as of 1.4 does not *require* the `component` field in `metadata`, nor does it *require* the BOM ref field: https://cyclonedx.org/docs/1.4/json/#metadata_component_bom-ref.

`guac` was encountering a NPE when trying to read these uninitialized fields from the CycloneDX struct.

Some SBOM generators that only operate on package lockfiles (such as https://pypi.org/project/cyclonedx-bom/) do not generate the `component` metadata field, nor are they required to by the spec.

This change makes the existence of this field optional and "fails closed" to the `nil` handler introduced by https://github.com/guacsec/guac/pull/603 - I believe this preserves the **intent** of the design while gracefully handling the non-existence of this metadata field.

The test I added is not great - it does check that a schema without these fields in metadata *parses* but the predicate output is `nil` due to the existing handlers, which I did not want to make sweeping changes to. I leave this up to the maintainers to decide how this *should* behave.

# PR Checklist


pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go
```diff
@@ -202,7 +202,13 @@ func (c *cyclonedxParser) GetPredicates(ctx context.Context) *assembler.IngestPr
 
 	preds := &assembler.IngestPredicates{}
 
-	toplevel := c.getPackageElement(string(c.cdxBom.Metadata.Component.BOMRef))
+	var toplevel []*model.PkgInputSpec = nil
+
+	if c.cdxBom.Metadata.Component != nil && c.cdxBom.Metadata.Component.BOMRef != "" {
```
pxp928, line 207: Hello @kurt-r2c! Thanks for the PR! This is something that warrants further discussion on how it should be handled in GUAC.

lumjjb: I am wondering what the use case of having a CDX document without the top level component is for... if it is the case where it is interpretted as here's a list of all the components I use, then it should parse to a bunch of singletons, however, if it is not , then my thought would be to either use the generic `guac/cdx/<name>` or to just reject it. let's make sure we have this on the priority for triage.

pxp928:
Hey @kurt-r2c, could you elaborate a bit more on your use case here? Based on the PR, it would generate `component` nodes without a top level `component` field in `metadata`. This would result in a disconnected node without a parent. How would you tie this information back to the package that the SBOM is being generated for?

kurt-r2c:
@pxp928 I was just trying to ingest a Python CycloneDX SBOM generated for a project using https://pypi.org/project/cyclonedx-bom/ to start exploring the tool. The SBOM output from `cyclonedx-bom` is valid according to the spec - if guac needs additional context to function, it needs to explicitly require some additional arguments/constraints.

lumjjb:
Yea let's perhaps add a message and link to this issue for discussion in the warning message. In case someone else comes along with a use case and a way to handle it.

pxp928:
Thanks @kurt-r2c for opening the PR and bringing this issue to our attention. We have created issue #976 to track this and also PR https://github.com/guacsec/guac/pull/992 to ensure that GUAC fails with an error message. Closing this PR as it it no longer needed.

PR #900, opened 2023-05-31
Fix parallel ingest when guacgql is in docker.

jeffmendoza:
Parallel ingest is failing half way through when running guacgql in docker. Works fine when running directly, or in k8s with the helm chart. This suggests a resource limitation in the docker networking setup.

Taking a look at the default Go client networking setup, it tries to re-use connections, but limits the number of idle connections per sever to 2. We are trying to do around 20, so they are constantly being closed and re-opened, since we only allow 2 to sit idle.

Bumping the number of idle connections allowed per server fixes the issue for me.

https://pkg.go.dev/net/http#DefaultTransport
https://pkg.go.dev/net/http#Transport
https://pkg.go.dev/net/http#pkg-constants (see DefaultMaxIdleConnsPerHost)

# PR Checklist

### DIFF
--- a/cmd/guacone/cmd/files.go
+++ b/cmd/guacone/cmd/files.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


pxp928 (approved):
LGTM

PR #907, opened 2023-06-02
Included tests and handled stack overflow error for parser

nathannaveen:
# Description of the PR

- Included tests for `parseHelper`
- Included tests for `parse`
- A graph nodes children is pointing back at itself can cause an infinite loop, resulting in a stack overflow error (i.e. root points to its child, and the child points back at the root, `root.Next = root.Child` and `root.Child.Next = root`). To solve this issue, I have included a map to store all visited nodes so that we don't keep calling nodes that have already been called. I have also included a test for a stack overflow.

# PR Checklist


pkg/ingestor/parser/parser_test.go
```diff
@@ -15,210 +15,192 @@
...
-// 				if len(d) != len(tt.want) {
-// 					t.Errorf("ParseDocumentTree() = %v, want %v", d, tt.want)
-// 				}
-// 				for i := range d {
-// 					compare(t, d[i].Edges, tt.want[i].Edges, d[i].Nodes, tt.want[i].Nodes)
-// 				}
-// 				return nil
-// 			}
-//
-// 			err = Subscribe(ctx, transportFunc)
```
pxp928, line 201: We should revive this test as well to test the functionality for the NATS Subscriber.

pxp928 (commented):
Thanks @nathannaveen! This looks good so far.

pkg/ingestor/parser/internal/mock/types.go
```diff
@@ -0,0 +1,95 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: /Users/nathannaveen/go/src/github.com/nathannaveen/guac/pkg/ingestor/parser/common/types.go
+
+// Package mocks is a generated GoMock package.
+package mocks
```
lumjjb, line 5: does this come with a configuration file? want to ensure that this gets updated when it needs to be. Should we add a make command for this? 

Also a broader question, on testing consistency, i don't think it matters too much now that we have tests that use the declaration type tests definitions vs the just code that does comparison (like current other tests). I do worry a bit on consistency and maintainability. Thoughts @pxp928 @mihaimaruseac @mihaimaruseac @jeffmendoza

lumjjb (approved):
minor nits and some open questions, else lgtm

lumjjb (approved):
minor nits and some open questions, else lgtm

lumjjb (approved):
minor nits and some open questions, else lgtm

pkg/ingestor/parser/internal/mock/types.go
mihaimaruseac: +1 to adding to Makefile and [CI](https://github.com/guacsec/guac/blob/131eeab8164633412cf15109e7d98fab5ce46dc3/.github/workflows/ci.yaml#L96-L97) to make sure it is always regenerated as needed.

Regarding test suites: I'd recommend striving for consistency, as that makes it easier to transfer knowledge from one area of GUAC to another.

pkg/ingestor/parser/parser_test.go
nathannaveen: I don't want to make this PR too large, so I will do that in a subsequent PR.

pkg/ingestor/parser/internal/mock/types.go
nathannaveen: What do you mean when you say the following? Could you please explain?
> Also a broader question, on testing consistency, i don't think it matters too much now that we have tests that use the declaration type tests definitions vs the just code that does comparison (like current other tests). I do worry a bit on consistency and maintainability. Thoughts @pxp928 @mihaimaruseac @mihaimaruseac @jeffmendoza

nathannaveen: I will add this to the CI and makefile.

pkg/ingestor/parser/parser.go
```diff
@@ -143,22 +143,33 @@ func ParseDocumentTree(ctx context.Context, docTree processor.DocumentTree) ([]a
 	return assemblerInputs, identifierStrings, nil
 }
 
-func (t *docTreeBuilder) parse(ctx context.Context, root processor.DocumentTree) error {
+// visitedKey is used to keep track of the document nodes that have already been visited to avoid infinite loops.
+// The key is a combination of the document type, format, and source information. This is used instead of
+// processor.Document because maps cannot use slices as keys.
+type visitedKey struct {
+	Type              processor.DocumentType
```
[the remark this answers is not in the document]
nathannaveen: You are right; I have made all of them internal.

pkg/ingestor/parser/parser_test.go
```diff
@@ -15,210 +15,192 @@
...
+	type fields struct {
+		identities    []common.TrustInformation
+		graphBuilders []*common.GraphBuilder
+	}
+	tests := []struct {
+		name            string
+		fields          fields
+		root            processor.DocumentTree
+		registerDocType processor.DocumentType
+		makeOverflow    bool
```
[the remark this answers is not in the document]
nathannaveen: 👍 I have added comments to `registerDocType` and `makeOverflow`.

pkg/ingestor/parser/internal/mock/types.go
nathannaveen: I have updated the Makefile, but didn't need to update the CI because I added the mockgen to the generate which is already being called in the CI.

pxp928: > What do you mean when you say the following? Could you please explain?

So the other tests [collector](https://github.com/guacsec/guac/blob/main/pkg/handler/collector/collector_test.go) and [processor](https://github.com/guacsec/guac/blob/main/pkg/handler/processor/process/process_test.go) all do some type of comparison to check that it received the proper documents/artifacts that the results should be as part of the unit tests. This is a deviation from those so the question comes to whether we want to keep them all consistent such that it's easier to understand/maintain (and for doing knowledge transfer in the future).

Makefile
```diff
@@ -91,6 +91,7 @@ fmt-md:
 .PHONY: generate
 generate:
 	go generate ./...
+	mockgen -source=pkg/ingestor/parser/common/types.go -destination=pkg/ingestor/parser/internal/mock/types.go -package=mocks -copyright_file=pkg/ingestor/parser/internal/mock/license.txt
```
pxp928, line 94: This can be moved to a `mockgen.go` similar to what we did here for the [scorecard mock gen](https://github.com/guacsec/guac/blob/main/internal/testing/mocks/mockgen.go). So that when `go generate ./...` is called, it will run the proper mockgen.

pkg/ingestor/parser/internal/mock/types.go
```diff
@@ -0,0 +1,110 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// Code generated by MockGen. DO NOT EDIT.
+// Source: pkg/ingestor/parser/common/types.go
+
+// Package mocks is a generated GoMock package.
+package mocks
```
pxp928, line 20: Could you move this into `internal/testing` or a subfolder in `internal/testing/mocks` to keep it all in once place?

pxp928 (commented):
Couple of comments! LGTM otherwise besides the open question.

pkg/ingestor/parser/internal/mock/types.go
nathannaveen: Thanks for the explanation, Parth! 

So, what is the outcome? Should I rewrite my tests in a similar way to the standard?

Makefile
nathannaveen: 👍

pkg/ingestor/parser/internal/mock/types.go
nathannaveen: Moved the file to `internal/testing/mocks`.

pxp928 (approved):
LGTM! Thanks @nathannaveen

.github/scripts/excluded_from_copyright
```diff
@@ -49,3 +49,4 @@
 ./pkg/assembler/graphql/resolvers/path.resolvers.go
 ./pkg/assembler/graphql/generated/path.generated.go
 ./internal/testing/mocks/scorecard.go
+./internal/testing/mocks/documentparser.go
```
pxp928, line 52: small nit: new line at end of file else LGTM

pxp928 (approved):

pkg/ingestor/parser/internal/mock/types.go
nathannaveen: Ok, sure, I will add the "comparison" checks for the `Test_ParserSubscribe` in the next PR.

pxp928:
@nathannaveen If you rebase on main, we fixed the CI issues. After that, it should auto-merge this PR.

PR #910, opened 2023-06-04
Implement RDMS backend (postgres/mysql/sqlite)

ivanvanderbyl:
# Description of the PR

(WIP) Implements a backend using [Ent](https://entgo.io) that supports common RDMS such as Postgres, MySQL, Sqlite. Tested only with Postgres at the moment.

# PR Checklist

## To implement:

- [x] Migration script (Atlas or ent)
- [x] Test harness using [txdb](github.com/DATA-DOG/go-txdb)
- [x] Artifacts
- [x] Builders
- [x] Cve
- [x] Ghsa
- [x] Osv
- [x] Packages
- [x] Sources
- [x] CertifyBad
- [x] CertifyGood
- [ ] CertifyVEXStatement
- [x] CertifyVuln
- [x] HasSBOM
- [x] HasSlsa
- [x] HasSourceAt
- [x] HashEqual
- [x] IsDependency
- [x] IsOccurrence
- [x] IsVulnerability
- [x] PkgEqual
- [ ] Scorecards
- [x] IngestArtifact
- [x] IngestArtifacts
- [x] IngestBuilder
- [x] IngestCve
- [x] IngestGhsa
- [x] IngestMaterials
- [x] IngestOsv
- [x] IngestPackage
- [x] IngestPackages
- [x] IngestSource
- [ ] CertifyScorecard
- [x] IngestCertifyBad
- [x] IngestCertifyGood
- [x] IngestDependency
- [x] IngestDependencies
- [x] IngestHasSbom
- [x] IngestHasSourceAt
- [x] IngestHashEqual
- [x] IngestIsVulnerability
- [x] IngestOccurrence
- [ ] IngestOccurrences
- [ ] IngestPkgEqual
- [x] IngestSLSA
- [ ] IngestVEXStatement
- [x] IngestVulnerability
- [ ] Neighbors
- [ ] Node
- [ ] Nodes
- [ ] Path
- [x] FindSoftware
- [ ] Implement new model.MatchType on deps
- [ ] Implement new vulnerability changes

*Note to reviewers:* A large chunk of this is code generated using Ent, which results in a large change surface. You'll notice that most ent code is identical for each node type, so you can skim it easily. The actual `Backend` implementation is in `pkg/assembler/backends/ent/backend` with the generated code in the parent directory.


pxp928:
Thanks, @ivanvanderbyl! We never knew about this. This could be a game changer for backend resolver implementation work.

lumjjb:
Nice!!!! This is super awesome!!! Looking forward to diving deeper into this! Let us know if you'd like to spend time to run through anything that you have questions on!

mihaimaruseac:
This is amazing! Thank you for starting this!

ivanvanderbyl:
@lumjjb one area that isn't so clear is the schema, I'm inferring a lot of this from reading the inmem and neo4j implementations. Do you have a design document for this?

Another area to discuss is how Ent can optimally work with `gqlgen`. Given the current Backend interface, if we make a query like (take from the docs):

```gql
packages(pkgSpec: { type: "deb" }) {
    type
    namespaces {
      namespace
    }
  }
```

The backing Ent query should look like:

```go
records,err := client.Package.Query().
  Where(package.TypeEQ(typ)).
  WithNamespaces().
  All(ctx)
```

Key to this is knowing that the user asked for namespaces and only eagerly querying that edge when asked. The current Backend API I believe makes this possible? Ensuring we don't create an N+1 query problem, which is quite easy to do with gqlgen 🦶🏻 🔫 

A larger design decision would  be to have Ent generate all this glue code between Ent and gqlgen if you don't mind the interface coupling (https://entgo.io/docs/tutorial-todo-gql-field-collection)

lumjjb:
>Key to this is knowing that the user asked for namespaces and only eagerly querying that edge when asked. The current Backend API I believe makes this possible?

Yea if I understand it right. The software trees like packages, etc. all have this behavior. 

>A larger design decision would be to have Ent generate all this glue code between Ent and gqlgen if you don't mind the interface coupling (https://entgo.io/docs/tutorial-todo-gql-field-collection)

I do personally like being able to manually have control over this since, I do envision that some graphql APIs could be tricky to optimize with ent generated code. With current APIs it looks like a nice fit, but I think we would add some more "analytics" type queries which may make things a bit tricky.


>Do you have a design document for this?

hmm, have you looked at some of these docs by any chance? If it still isn't clear maybe we can have a call and answer some questions and then add to the docs to make things clearer!

- [GraphQL Doc](https://docs.guac.sh/graphql/)
- [Ontology doc](https://docs.guac.sh/guac-ontology/)
- [Ontology Definitions](https://docs.guac.sh/guac-ontology-definition/)

pxp928:
> Yea if I understand it right. The software trees like packages, etc. all have this behavior.
> 
> >

+1 Via the graphQL preloads, we can get the query fields that the user requested and optimize based on that. We did something similar in the neo4j implementation.

mihaimaruseac:
> Key to this is knowing that the user asked for namespaces and only eagerly querying that edge when asked. The current Backend API I believe makes this possible?

Yes, we have a [`getPreloads`](https://github.com/guacsec/guac/blob/08bbd69f6da9fb73c9b7bcb57b743912e6752e1c/pkg/assembler/backends/neo4j/backend.go#L128) function for this, but this can be adapted if needed

ivanvanderbyl:
👍🏻  for preloads.

Thanks for pointing me to the docs, hadn't dug deep enough to find them yet. Big help! I've added the software tree packages and a tests. I've also configured this to do an upsert so that you can call ingest multiple times without duplicating each tree node.

I think this is starting to take shape.

ivanvanderbyl:
I got a PoC working with ingesting packages and querying packages 😁 
image

pkg/assembler/backends/ent/backend/jeffwork.go
```diff
@@ -0,0 +1,353 @@
...
+		ID:            nodeID(o.ID),
+		Subject:       sub,
+		Artifact:      toModelArtifact(o.Edges.Artifact),
+		Justification: o.Justification,
+		Origin:        o.Origin,
+		Collector:     o.Collector,
+	}, nil
+}
+
+func toModelIsDependency(ctx context.Context, id *ent.Dependency) (*model.IsDependency, error) {
```
ivanvanderbyl, line 289: I'm conflicted about whether transform funcs should be allowed to query data or assume it's provided up front. I can see a lot of potential performance problems if you unwittingly make a query in each loop.

ivanvanderbyl: After reasoning about this yesterday, I think running queries inside the transform funcs is a bad idea as it will nearly always be suboptimal, instead we should eager load all the edges ahead of time using `With<EdgeType>(...)`

We can still build queries from reusable query funcs without executing them.

lumjjb:
@ivanvanderbyl would you be interested in showing this off at the next community meeting on July 20th? :D

ivanvanderbyl:
@lumjjb Yeah! absolutely keen to show it off. I just added the meeting to my calendar, it's usually at 5am here (NZST), but I might be in Vancouver by then and TZs will be a little more friendly. Otherwise I can pre-record something.

lumjjb:
Awesome! I've put you on the schedule if timezone matches up that's great - else we'd be happy to reschedule to a time that works better for you for one instance!

lumjjb:
Sent you a cal invite as well!

pkg/assembler/backends/ent/backend/cve.go
```diff
@@ -0,0 +1,148 @@
...
+		Year:  spec.Year,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return collect(results, toModelCVE), nil
+}
+
+func (b *EntBackend) IngestCve(ctx context.Context, spec *model.CVEInputSpec) (*model.Cve, error) {
```
noqcks, line 34: I think you might want it like this instead, otherwise client is not passed correctly, and then remove the `WithinTX` from `upsertAdvisory`

```
func (b *EntBackend) IngestCve(ctx context.Context, spec *model.CVEInputSpec) (*model.Cve, error) {
	advisory, err := WithinTX(ctx, b.client, func(ctx context.Context) (*ent.SecurityAdvisory, error) {
		client := ent.FromContext(ctx)
		return upsertAdvisory(ctx, client, advisoryQuerySpec{
			CveID: &spec.CveID,
			Year:  &spec.Year,
		})
	})
	if err != nil {
		return nil, err
	}
	return toModelCVE(advisory), nil
}
```

ivanvanderbyl: I guess for consistency that makes sense... `upsertAdvisory` simply wraps this logic for CVEs, GHSAs, and OSV since they are stored in the same table. At the moment this is the only callsite for upserts, but if we end up needing it to not be transactional I think that change would make sense.

pkg/assembler/backends/ent/backend/package.go
```diff
@@ -0,0 +1,377 @@
...
+	for i, s := range q {
+		results[i] = &model.PackageQualifierSpec{
+			Key:   s.Key,
+			Value: &s.Value,
+		}
+	}
+	return results
+}
+
+func packageVersionQuery(spec *model.PkgInputSpec) predicate.PackageVersion {
```
noqcks, line 205: I believe these `where` predicates are failing to match two package versions that are identical but which have different qualifiers.

Running this input against this data using `getPkgVersion` fn will result in an error as it expects only one result, but will return two

Input:
```
namespace := "debian"
version := "3.11.2-1+b1"
subpath := ""
pkgin := model.PkgInputSpec{
		Type:       "deb",
		Namespace:  &namespace,
		Name:       "python3",
		Version:    &version,
		Qualifiers: []*model.PackageQualifierInputSpec{},
		Subpath:    &subpath,
	}
```

DB: 
```
[
	{
		"id": 64424511484,
		"version": "3.11.2-1+b1",
		"subpath": "",
		"qualifiers": [
			{
				"key": "arch",
				"value": "arm64"
			},
			{
				"key": "distro",
				"value": "debian-12"
			},
			{
				"key": "upstream",
				"value": "python3-defaults@3.11.2-1"
			}
		],
		"hash": "b1a934efb143f8d71e0d342c9f23066124d36f7e",
		"name_id": 51539609596,
		"tenant_id": 90194313216
	},
	{
		"id": 64424512478,
		"version": "3.11.2-1+b1",
		"subpath": "",
		"qualifiers": [
		],
		"hash": "14a747debb4ca0ca8b878d0e4913720e8a0e8b70",
		"name_id": 51539609596,
		"tenant_id": 90194313216
	}
]
```

mihaimaruseac:
I'll review this by the end of the week. Can you fix the merge conflict please?

ivanvanderbyl:
@mihaimaruseac done :)

cmd/guacgql/cmd/root.go
```diff
@@ -40,6 +40,12 @@ var flags = struct {
 	nPass  string
 	nRealm string
 
+	// Needed only if using ent backend
```
mihaimaruseac, line 43: I think at some point (not in this PR, just commenting here to start a discussion) we might want to separate these flags and have them active/present only if the backend to which they apply is included.

Thoughts @lumjjb @pxp928 @jeffmendoza ?

cmd/guacgql/cmd/server.go
```diff
@@ -111,16 +113,33 @@ func startServer(cmd *cobra.Command) {
 
 func validateFlags() error {
 	if flags.backend != neo4js &&
-		flags.backend != inmems && flags.backend != arango && flags.backend != neptune {
+		flags.backend != inmems && flags.backend != arango {
```
mihaimaruseac, line 116: I think we still need to keep `neptune` here?

container_files/guac/guac.yaml
```diff
@@ -17,7 +17,7 @@ csub-addr: guac-collectsub:2782
 csub-listen-port: 2782
 
 # graphql
-gql-backend: inmem
+gql-backend: ent # ent|neo4j|inmem
```
mihaimaruseac, line 20: I think here we need to keep inmem as default for now?

mihaimaruseac (commented):
Took a first pass through the first set of files. Will cover the next ones in my next review round

cmd/guacgql/cmd/root.go
pxp928: +1, otherwise we will end up with many unused flags that will cause confusion.

cmd/guacgql/cmd/server.go
ivanvanderbyl: Oh definitely, that's accidentally got dropped in the merge, will fix.

.goreleaser.yaml
```diff
@@ -103,7 +103,6 @@ builds:
     goarch:
       - amd64
       - arm64
-      - arm
```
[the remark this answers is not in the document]
mihaimaruseac: @pxp928 @lumjjb @jeffmendoza : should we keep arm support or first migrate to UUIDs and then continue with this PR?

pkg/assembler/backends/ent/artifact.go
```diff
@@ -0,0 +1,291 @@
...
+	"fmt"
+	"strings"
+
+	"entgo.io/ent"
+	"entgo.io/ent/dialect/sql"
+	"github.com/guacsec/guac/pkg/assembler/backends/ent/artifact"
+)
+
+// Artifact is the model entity for the Artifact schema.
+type Artifact struct {
```
mihaimaruseac, line 15: Question: When the GraphQL schema changes will this be autogenerated or we'll have to update?

cmd/guacgql/cmd/root.go
lumjjb: yea i like the prefix backend address idea!

pkg/assembler/backends/ent/artifact.go
ivanvanderbyl: This schema will need to be manually edited and then run the code gen for ent.

.goreleaser.yaml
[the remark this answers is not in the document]
pxp928: How much work is it to change to UUIDs? I rather keep the arm support. If possible can that be commented out, create an issue, and fixed in another PR or is that not possible?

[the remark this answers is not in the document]
ivanvanderbyl: It's not a lot, just need to change all the IDs to UUIDs and regenerate. It'll probably break a few things in the process but I can work around that. I was thinking this over and it _might_ be easier to fix it in Ent instead, as it's definitely a bug. I'll raise this with the ent team.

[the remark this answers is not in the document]
ivanvanderbyl: On second thought, I took a look at the implementation in Ent and it should work fine on 32bit builds. I think this error might have been related to something else, so I'll add arm back in.

container_files/guac/guac.yaml
pxp928: @ivanvanderbyl can you change this to be inmem by default

mrizzi:
> @pxp928 makes sense. How are you handling the database services for testing arango/neo etc? Is it just a matter of adding a postgres service to Github actions?

Maybe it's out of scope for this PR but I would suggest to evaluate testcontainers.
It would help a lot without having to rely on GH actions but relying on a solution that would easily work also locally for developers.

pxp928:
> Maybe it's out of scope for this PR but I would suggest to evaluate testcontainers.
It would help a lot without having to rely on GH actions but relying on a solution that would easily work also locally for developers.

Interesting. I like the idea!

mrizzi:
So if @ivanvanderbyl is more confident with GH Services, for this PR, it's fine to add them in order to have the Postresql instance to run tests with.

Then, later on, I can work on adding Testcontainers support to guac tests for Ent and maybe also for ArangoDB to have consistency across development and testing with different backends.

jeffmendoza (approved):
Lets merge this into main and continue there for the finishing touches.

pxp928 (approved):

cmd/guacgql/cmd/root.go
pxp928: @ivanvanderbyl did you want to create an issue to capture this needed change?

.goreleaser.yaml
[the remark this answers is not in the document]
pxp928: @ivanvanderbyl can you create an issue for this also

PR #942, opened 2023-06-14
Reorg graphQL to ingest nouns and verbs separately 

pxp928:
# Description of the PR

Based on testing, it makes sense to ingest the nouns and verbs separately. This allows for improvement in the ingestion time. This PR will also de-duplicate the nouns such that there is no need to ingest if it already exists.

# PR Checklist


pxp928:
Looking at the issue here.

mihaimaruseac (commented):
So, on a quick glance this breaks invariants since we now allow callers to directly ingest nouns without a link to the document from which the noun was created.

I thought we were discussing changing the client queries to ingest nouns first and then verbs, but still linked to the doc.

Also, this increases number of round trips to the GraphQL server

pxp928:
> So, on a quick glance this breaks invariants since we now allow callers to directly ingest nouns without a link to the document from which the noun was created.

Thats true...so what would the alternative be, ingest nouns first and then verbs but don't keep ingesting nouns as they should already be there?

jeffmendoza:
> Also, this increases number of round trips to the GraphQL server

Overall performance limitations with ingestion seems to be DB lookups/upserts. We are trying to avoid ingestion the same noun 10,000 times, if it has 10,000 dependencies.

Round trips should be mitigated with either

- Presently: Concurrent gql calls
- Possibly batching gql calls
- Using the in-cluster ingestor once the nats message size issue is figured out.

pxp928:
@mihaimaruseac can you review this? I can get the other changes staged on top of these.

mihaimaruseac:
Will do this today, sorry, I forgot about it.

pxp928:
We will have to restrict ingest nouns without a link to the document. Keep it as an internal API that we do not expose out for other integrations to use.

mihaimaruseac (approved):

Issue #941, opened 2023-06-14
[feature] can we have snapshot daily build 


tuananh:
Currently it's moving quite fast and we would love to try snapshot daily build to get the latest fixes and feats.

pxp928:
That's a good idea @tuananh! We can start doing that if it helps the community.

sunnyyip:
At the moment, goreleaser is run by the release workflow at https://github.com/guacsec/guac/blob/main/.github/workflows/release.yaml#L56-L78. The type of build it runs (snapshot/release) is based on the event that triggered the workflow.

I'd like to move all the goreleaser commands into the Makefile and have the release workflow consume it. Similar to how we build the local-organic-guac image for local testing - https://github.com/guacsec/guac/blob/main/Makefile#L105-L121
This would allow people to test the builds more easily and consolidate all goreleaser commands into one place.

With that, a nightly build would be a make target with specific goreleaser inputs defined for a specific github event triggers representing the nightly build - e.g. push of a 'nightly' tag - and the build should replace the nightly build artifacts (bins/container image) produced from the previous night.

PR #948, opened 2023-06-18
ci: add schedule daily snapshot build

tuananh:
# Description of the PR

Snapshot build logic is already there, i just add schedule to trigger every day at 0:00

Fixes: #941

# PR Checklist

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -16,6 +16,8 @@ name: release
 
 on:
   workflow_dispatch: # testing only, trigger manually to test it works
+  schedule:
+    - cron: 0 0 * * * # every day at 0:00, build latest on default branch (main)
   push:
     branches:
       - main


sunnyyip:
The snapshot build in the current release workflow is meant to test the build-ability as code get merged into main. It doesn't releaser any form of artifacts at the moment. We'll need to come up with a strategy to name the nightly release and how to trigger it in the context of the current goreleaser config.

.github/workflows/release.yaml
```diff
@@ -16,6 +16,8 @@ name: release
 
 on:
   workflow_dispatch: # testing only, trigger manually to test it works
+  schedule:
+    - cron: 0 0 * * * # every day at 0:00, build latest on default branch (main)
```
lumjjb, line 20: nit: can we change this to have a bit of randomness since its likely a lot of people will be doing things at 0:00, maybe 0 23 or something?

lumjjb (commented):
minor nit: I think it would be nice to also add a description tag of nightly build. Would like to get review from @sunnyyip before merging this since he's been owning this bit of the code.

tuananh:
> The snapshot build in the current release workflow is meant to test the build-ability as code get merged into main. It doesn't release any form of artifacts at the moment. We'll need to come up with a strategy to name the nightly release and how to trigger it in the context of the current goreleaser config.

this one https://github.com/guacsec/guac/blob/8282449c65a3d0a79016a3ce9fb8434916f2662b/.goreleaser.yaml#L203 does not release image?

sunnyyip:
> > The snapshot build in the current release workflow is meant to test the build-ability as code get merged into main. It doesn't release any form of artifacts at the moment. We'll need to come up with a strategy to name the nightly release and how to trigger it in the context of the current goreleaser config.
> 
> this one
> 
> https://github.com/guacsec/guac/blob/8282449c65a3d0a79016a3ce9fb8434916f2662b/.goreleaser.yaml#L203
> 
> does not release image?

@tuananh It doesn't - the flag merely defines the snapshot tag. 

At the moment, goreleaser is run by the release workflow at https://github.com/guacsec/guac/blob/main/.github/workflows/release.yaml#L56-L78. The type of build it runs (snapshot/release) is based on the event that triggered the workflow.

I'd like to move all the goreleaser commands into the Makefile and have the release workflow consume it. Similar to how we build the local-organic-guac image for local testing - https://github.com/guacsec/guac/blob/main/Makefile#L105-L121
This would allow people to test the builds more easily and consolidate all goreleaser commands into one place.

With that, a nightly build would be a make target with specific goreleaser inputs defined for a specific github event triggers representing the nightly build - e.g. push of a 'nightly' tag - and the build should replace the nightly build artifacts (bins/container image) produced from the previous night.

PR #964, opened 2023-06-21
Implement search dependent packages + testing

rmetzman:
# Description of the PR

 * Code will be used for patch planning in order to traverse packages through isDep relationships 
 * Uses code from file "guac/guac/cmd/guacone/cmd/vulnerability.go" function searchDependencyPackages but modified to focus on packages rather than vulnerabilities
 * Added testing

# PR Checklist

- [ x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.


pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+		topTopPkg1           model.PkgInputSpec
+		topTopPkg2           model.PkgInputSpec
+		topTopPkg3           model.PkgInputSpec
+		topPkg               model.PkgInputSpec
+		pkg                  model.PkgInputSpec
+		depPkg1              model.PkgInputSpec
+		depPkg2              model.PkgInputSpec
+		dependencyDirect     model.IsDependencyInputSpec
+		dependencyNotInRange model.IsDependencyInputSpec
+		dependencyIndirect   model.IsDependencyInputSpec
```
lumjjb, line 64: how about `assembler.IngestPredicates` here? may streamline the test cases a bit more so its easier to add future test cases.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,128 @@
...
+		if maxDepth != 0 && nowNode.depth >= maxDepth {
+			break
+		}
+
+		isDependencyNeighborResponses, err := model.Neighbors(ctx, gqlclient, now, []model.Edge{model.EdgePackageIsDependency})
+		if err != nil {
+			return nil, fmt.Errorf("failed getting package parent:%w", err)
+		}
+		for _, neighbor := range isDependencyNeighborResponses.Neighbors {
+			if isDependency, ok := neighbor.(*model.NeighborsNeighborsIsDependency); ok {
```
lumjjb, line 73: would be helpful to refactor this to case on each type of node so it is easier for one to see the behavior for each node, and then look up the behavior that way... For example, having a switch statement on type(neighbor), each calling to its own function.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,128 @@
...
+)
+
+type DfsNode struct {
+	expanded bool // true once all node neighbors are added to queue
+	parent   string
+	depth    int
+}
+
+// TODO: make more robust usuing predicates
+func searchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Client, startID string, stopID string, maxDepth int) (map[string]DfsNode, error) {
```
lumjjb, line 34: probably should be a public function.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+
+	"github.com/Khan/genqlient/graphql"
+	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"github.com/guacsec/guac/pkg/misc/depversion"
+)
+
+type DfsNode struct {
+	expanded bool // true once all node neighbors are added to queue
+	parent   string
```
lumjjb, line 29: in order for this to be used as a return statement from searchDependenciesFromStartNode, it probably requires maybe at least the parent node to be public. This could implementation could potentially hide behind an interface as well.

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+			Origin:         "Demo ingestion",
+			Collector:      "Demo ingestion",
+		},
+	}}
+	for _, ingest := range ingestDependencies {
+		_, err1 := model.IsDependency(context.Background(), client, ingest.topPkg, ingest.pkg, ingest.dependencyDirect)
+		if err1 != nil {
+			logger.Errorf("Error in ingesting: %v\n", err1)
+		}
+		_, err2 := model.IsDependency(context.Background(), client, ingest.pkg, ingest.depPkg1, ingest.dependencyDirect)
```
lumjjb, line 138: re-use err instead of redefining a new variable since it is no longer used after each err check.

pkg/guacanalytics/patchPlanning_test.go
rmetzman: Fixed (on new PR #969)

pkg/guacanalytics/patchPlanning.go
rmetzman: Fixed (on new PR #969)

pkg/guacanalytics/patchPlanning.go
rmetzman: Fixed (on new PR #969 )

Issue #963, opened 2023-06-21
[feature] Add parallel ingestion for assembler being used by NATS


pxp928:
**Is your feature request related to a problem? Please describe.**

Currently, the assembler being used by [guacIngest](https://github.com/guacsec/guac/blob/main/cmd/guacingest/cmd/ingest.go#L185) is not running in parallel. [guacone collect](https://github.com/guacsec/guac/blob/main/cmd/guacone/cmd/files.go#L250) is utilizing the [parallel assembler](https://github.com/guacsec/guac/blob/main/pkg/assembler/clients/helpers/parallel.go#L28).

**Describe the solution you'd like**

1. They both need to be consistent and use the parallel assembler to speed up the ingestion of data. 
2. Tests need to be conducted to ensure functionality does not break for [guacIngest](https://github.com/guacsec/guac/blob/main/cmd/guacingest/cmd/ingest.go#L185) by running:
- `make container`
- `docker compose up --force-recreate`
- `go run ./cmd/guaccollect files ../guac-data/docs/spdx/spdx_vuln.json`

Check that there is no errors in the docker logs for ingestion and it returns logs similar to this:

```
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373183.041415,"caller":"parser/parser.go:126","msg":"parsing document tree with root type: SPDX"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373183.0902941,"caller":"helpers/assembler.go:33","msg":"assembling CertifyScorecard: 0"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373183.0903711,"caller":"helpers/assembler.go:40","msg":"assembling IsDependency: 7161"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373185.0002403,"caller":"helpers/assembler.go:47","msg":"assembling IsOccurence: 3878"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373185.9143217,"caller":"helpers/assembler.go:54","msg":"assembling HasSLSA: 0"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373185.9145687,"caller":"helpers/assembler.go:61","msg":"assembling CertifyVuln: 0"}
artifact-ff-guac-ingestor-1      | {"level":"info","ts":1687373185.914619,"caller":"helpers/assembler.go:68","msg":"assembling IsVuln: 0"}
[...]

PR #967, opened 2023-06-22
update nats workflow to use parallel assembler

pxp928:
# Description of the PR
update nats workflow to use parallel assembler

closes https://github.com/guacsec/guac/issues/963

# PR Checklist

### DIFF
--- a/cmd/guacingest/cmd/ingest.go
+++ b/cmd/guacingest/cmd/ingest.go
@@ -19,11 +19,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/Khan/genqlient/graphql"
 	"github.com/guacsec/guac/pkg/assembler"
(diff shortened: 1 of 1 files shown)


cmd/guacingest/cmd/ingest.go
```diff
@@ -179,9 +177,26 @@ func getIngestor(ctx context.Context, transportFunc func([]assembler.IngestPredi
...
-	return f, nil
+func getAssembler(ctx context.Context, graphqlEndpoint string) func([]assembler.IngestPredicates) error {
+	// Same as https://pkg.go.dev/net/http#DefaultTransport but with greater
+	// MaxIdleConnsPerHost so that we effectively re-use connections when doing
+	// parallel ingestion calls.
+	var dialer = &net.Dialer{
+		Timeout:   30 * time.Second,
+		KeepAlive: 30 * time.Second,
+	}
+	var customTransport http.RoundTripper = &http.Transport{
```
lumjjb, line 188: nit: feel like these would be nice if they were changable via config file, would help in perf testing as well.

lumjjb (approved):

cmd/guacingest/cmd/ingest.go
pxp928: Sure can do that. The only modification that was made here was with `MaxIdleConnsPerHost`. Did we want to expose the others also? cc @jeffmendoza

mihaimaruseac (approved):

cmd/guacingest/cmd/ingest.go
jeffmendoza: Just a "parallel: yes/no" should be fine imo.

But if you want to add the # outstanding config, it would be here: https://github.com/guacsec/guac/blob/main/pkg/assembler/clients/helpers/parallel.go#L37-L39 Then the `MaxIdleConnsPerHost` needs to be greater than that by some margin.

pxp928:
Closing this PR as parallel assembler is not needed. Rather need parallel document ingestion.

PR #969, opened 2023-06-22
Implement search dependent packages + testing (updated)

rmetzman:
# Description of the PR

 * Code will be used for patch planning in order to traverse packages through isDependency relationships 
 * Uses code from file "guac/guac/cmd/guacone/cmd/vulnerability.go" function searchDependencyPackages but modified to focus on packages rather than vulnerabilities
 * Added testing
 * This is the same as PR #964 in order to fix git issues

# PR Checklist


pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+		pkgResponse, err := model.Packages(ctx, gqlclient, pkgFilter)
+
+		if err != nil {
+			t.Errorf("Error getting id for isDep test case: %s\n", err)
+		}
+
+		pkgIds = append(pkgIds, pkgResponse.Packages[0].Namespaces[0].Names[0].Versions[0].Id)
+	}
+
+	testCases := []struct {
```
lumjjb, line 257: probably will be helpful to have the ingestion be part of the test case since it probably will be good to have multiple graphs being tested as well.

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+		startID  string
+		stopID   string
+		maxDepth int
+	}{
+		{
+			// 1: test case with two dependencies at the same depth, no stopID and no limiting maxDepth
+			startID:  pkgIds[0],
```
lumjjb, line 264: the start node probably should be defined such that it's not dependent on the array index of the graph. It will be nice as a test to be able to read the test case and determine what the test is doing.

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+		pkgIds = append(pkgIds, pkgResponse.Packages[0].Namespaces[0].Names[0].Versions[0].Id)
+	}
+
+	testCases := []struct {
+		startID  string
+		stopID   string
+		maxDepth int
+	}{
+		{
+			// 1: test case with two dependencies at the same depth, no stopID and no limiting maxDepth
```
lumjjb, line 263: nit: Having the test name itself be part of the test case will also allow it to be run as a test using t.Run(...)

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,460 @@
...
+			if tt.stopID == "" && tt.maxDepth == 10 {
+				for k, v := range gotMap {
+					if !v.expanded {
+						t.Errorf("All nodes should be expanded but this node was not: node %s \n", k)
+					}
+				}
+			}
+
+			// Test Case 1
+			if tt.startID == pkgIds[0] {
```
lumjjb, line 317: tests are repeated code and should probably be a loop instead over the test cases slice.

lumjjb (commented):
Good work - almost there!

pkg/guacanalytics/patchPlanning_test.go
rmetzman: Some of them are slightly different but I generalized the cases in order to reduce seperating out each test case / repeating code.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+
+		nodes := map[string]model.NeighborsResponse{}
+
+		isDependencyNeighborResponses, err := model.Neighbors(ctx, gqlclient, now, []model.Edge{model.EdgePackageIsDependency})
+		if err != nil {
+			return nil, fmt.Errorf("failed getting package parent:%w", err)
+		}
+
+		nodes["isDep"] = *isDependencyNeighborResponses
```
pxp928, line 79: It would be more robust to check the type similar to:
```
if isDependency, ok := neighbor.(*model.NeighborsNeighborsIsDependency); ok {
```
and use that for the switch statement instead of using the map.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+				Name:      &isDependency.DependentPackage.Namespaces[0].Names[0].Name,
+			}
+
+			depPkgResponse, err := model.Packages(ctx, gqlclient, dependentPkgFilter)
+			if err != nil {
+				return fmt.Errorf("error querying for dependent package: %w", err)
+			}
+
+			depPkgVersionsMap := map[string]string{}
+			depPkgVersions := []string{}
```
pxp928, line 115: this is a change I need to make upstream also but its better to initialize the slice via `var depPkgVersions []string`

pxp928 (commented):
Looking good! Almost there!

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+		nodes := map[string]model.NeighborsResponse{}
+
+		isDependencyNeighborResponses, err := model.Neighbors(ctx, gqlclient, now, []model.Edge{model.EdgePackageIsDependency})
+		if err != nil {
+			return nil, fmt.Errorf("failed getting package parent:%w", err)
+		}
+
+		nodes["isDep"] = *isDependencyNeighborResponses
+
+		for predicateType, response := range nodes {
```
lumjjb, line 81: this will always be one? since each iteration will only look at a set of nodes?

lumjjb: this should iterate over the loop of neighbors here instead?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+			switch predicateType {
```
lumjjb, line 82: can a type switch be used here instead?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+			break
+		}
+
+		if maxDepth != 0 && nowNode.depth >= maxDepth {
+			break
+		}
+
+		nodes := map[string]model.NeighborsResponse{}
+
+		isDependencyNeighborResponses, err := model.Neighbors(ctx, gqlclient, now, []model.Edge{model.EdgePackageIsDependency})
```
lumjjb, line 74: probably want to get all the nodes here and the case on them.. so you don't want to make multiple calls here.

this would be `NeighborResponses` instead of `isDependencyNeighborResponses`..

pkg/guacanalytics/patchPlanning.go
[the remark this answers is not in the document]
lumjjb: see above comment, you'd probably want to only have 1 neighbors call instead of multiple

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,151 @@
...
+func exploreIsDependency(isDependencyNeighborResponses model.NeighborsResponse, ctx context.Context, gqlclient graphql.Client) error {
+	for _, neighbor := range isDependencyNeighborResponses.Neighbors {
+		if isDependency, ok := neighbor.(*model.NeighborsNeighborsIsDependency); ok {
+			dependentPkgFilter := &model.PkgSpec{
+				Type:      &isDependency.DependentPackage.Type,
+				Namespace: &isDependency.DependentPackage.Namespaces[0].Namespace,
+				Name:      &isDependency.DependentPackage.Namespaces[0].Names[0].Name,
+			}
+
+			depPkgResponse, err := model.Packages(ctx, gqlclient, dependentPkgFilter)
```
lumjjb, line 109: what about handling this in the outer function? since this is doing exactly that of exploring neighbors?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+	nodeType string // packageName, packageVersion
+}
+
+var path []string
+var nodeMap map[string]DfsNode
+var now string
+var nowNode DfsNode
+var queue []string
+
+// TODO: make more robust usuing predicates
```
lumjjb, line 40: nit: typo using

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+	for len(queue) > 0 {
+		now = queue[0]
+		queue = queue[1:]
+		nowNode = nodeMap[now]
+
+		if stopID == now {
+			break
+		}
+
+		if maxDepth != 0 && nowNode.depth >= maxDepth {
```
lumjjb, line 71: hmm, what's the case for checking maxDepth != 0 ?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+			break
+		}
+
+		neighborsResponse, err := model.Neighbors(ctx, gqlclient, now, []model.Edge{})
+
+		if err != nil {
+			return nil, fmt.Errorf("failed getting package parent:%w", err)
```
lumjjb, line 78: error message should be getting neighbors, not just parents, and not specifically to a package

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+
+		// case on predicates and nodeType
+		for _, neighbor := range neighborsResponse.Neighbors {
+			if nowNode.nodeType == "packageVersion" {
```
lumjjb, line 83: would be good if this can be refactored to another function with nowNode and neighbor

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+				if isDependency, ok := neighbor.(*model.NeighborsNeighborsIsDependency); ok {
```
lumjjb, line 84: use switch on type here would be useful https://go.dev/tour/methods/16

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+	nodeMap = map[string]DfsNode{}
+
+	nodeMap[startID] = DfsNode{
+		nodeType: startType,
+	}
+	queue = append(queue, startID)
+
+	for len(queue) > 0 {
+		now = queue[0]
+		queue = queue[1:]
```
lumjjb, line 64: where else are nodes being added to the frontier here? it seems like this is only a single step?

lumjjb (changes requested):

pkg/guacanalytics/patchPlanning.go
lumjjb: e.g. https://go.dev/tour/methods/4

lumjjb: The current implementation makes this function not async safe as it uses a global variable.

rmetzman: Done -- changed to pass a a pointer to a struct.

pkg/guacanalytics/patchPlanning.go
rmetzman: Done.

pkg/guacanalytics/patchPlanning.go
rmetzman: Done.

pkg/guacanalytics/patchPlanning.go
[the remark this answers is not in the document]
rmetzman: Done.

pkg/guacanalytics/patchPlanning.go
rmetzman: Done.

pkg/guacanalytics/patchPlanning.go
rmetzman: Done.

pkg/guacanalytics/patchPlanning.go
rmetzman: Done.

lumjjb (approved):
LGTM! Thanks!

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+		q.nodeMap[q.now] = q.nowNode
+	}
+
+	return q.nodeMap, nil
+
+}
+
+func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueValues, neighbor model.NeighborsNeighborsNode) (queueValues, error) {
+	// case on predicates and nodeType
+	switch q.nowNode.nodeType {
```
pxp928, line 104: not sure I understand the reason for this switch statement. Can you explain your thoughts here?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+		if err != nil {
+			return *q, fmt.Errorf("error querying neighbor: %w", err)
+		}
+
+		q.path = append(q.path, isDependency.Id, matchingDepPkgVersionID,
+			depPkgResponse.Packages[0].Namespaces[0].Names[0].Id, depPkgResponse.Packages[0].Namespaces[0].Id,
+			depPkgResponse.Packages[0].Id)
+
+		dfsNVersion, seenVersion := q.nodeMap[matchingDepPkgVersionID]
+		dfsNName, seenName := q.nodeMap[depPkgResponse.Packages[0].Namespaces[0].Names[0].Id]
```
pxp928, line 154: Do we need to check the name and version both?

pxp928 (commented):
Looks good. Just some questions

pkg/guacanalytics/patchPlanning.go
rmetzman: Later I may add for the "packageName" nodes to be cased on for predicated attached to the name level where that would need to be searched as well

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+
+		if stopID == q.now {
+			break
+		}
+
+		if q.nowNode.depth >= maxDepth {
+			break
+		}
+
+		neighborsResponse, err := model.Neighbors(ctx, gqlclient, q.now, []model.Edge{})
```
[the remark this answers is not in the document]
rmetzman: Yes, it is to make it more generic and then in the explore methods I add the filtering for the edge type.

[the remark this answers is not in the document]
lumjjb: yea i think the idea here is to be more generic (with a couple other PRs coming in right after)

pkg/guacanalytics/patchPlanning.go
pxp928: Meaning that the package version will have all the dependencies. Not at the name level? So what is the reason to check the name level also?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+
+}
+
+func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueValues, neighbor model.NeighborsNeighborsNode) (queueValues, error) {
+	// case on predicates and nodeType
+	switch q.nowNode.nodeType {
+	case "packageVersion":
+		switch neighbor := neighbor.(type) {
+		case *model.NeighborsNeighborsIsDependency:
+			q, err := exploreIsDependency(ctx, gqlclient, q, *neighbor)
```
pxp928, line 108: as you are already passing in a reference of queries, there is no reason to return it?

pkg/guacanalytics/patchPlanning.go
rmetzman: I believe you are correct, at least for now when implementing dependencies only they are attached at the version.  Just changed (as well as removing the nodeType switch you commented on earlier).

pkg/guacanalytics/patchPlanning.go
rmetzman: Yes, good point.  Fixed.

pkg/guacanalytics/patchPlanning.go
pxp928: yeah, I would have a check based on whether it is required or can be at the "packageName" level. Else you are unnecessarily computing.

pkg/guacanalytics/patchPlanning.go
pxp928: Yeah, I would have a check based on whether it is required or can be at the "packageName" level. Else you are unnecessarily computing.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+type queueValues struct {
+	path    []string
+	nodeMap map[string]DfsNode
+	now     string
+	nowNode DfsNode
+	queue   []string
+}
+
+// TODO: make more robust using predicates
+func SearchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Client, startID string, stopID string, startType string, maxDepth int) (map[string]DfsNode, error) {
```
pxp928, line 42: is `startType` being used?

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,399 @@
...
+
+			if stopID == "" && tt.maxDepth == 10 {
+				for k, v := range gotMap {
+					if !v.expanded {
+						t.Errorf("All nodes should be expanded but this node was not: node %s \n", k)
+					}
+				}
+			}
+
+			if diff := cmp.Diff(tt.expectedLen, len(gotMap)); len(diff) > 0 {
```
pxp928, line 313: can you compare a sorted list? To make sure we get the path we expect back when searching for a node

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,162 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/Khan/genqlient/graphql"
+	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"github.com/guacsec/guac/pkg/misc/depversion"
+)
+
+type DfsNode struct {
```
pxp928, line 27: if the struct is public, the values also have to be public if you need to export them and use them else where

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,399 @@
...
+func ingestTestData(graphInput string, ctx context.Context, client graphql.Client) {
+	logger := logging.FromContext(ctx)
+	switch graphInput {
+	case "isDependency":
+		for _, ingest := range isDepTestData.IsDependency {
+
+			_, err := model.IngestPackage(context.Background(), client, *ingest.Pkg)
+
+			if err != nil {
+				logger.Errorf("Error in ingesting package: %v\n", err)
```
pxp928, line 183: can return the errors here and fail the test.

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,399 @@
...
+	cf()
+}
+
+func startTestServer() (*http.Server, *zap.SugaredLogger) {
+	ctx := logging.WithLogger(context.Background())
+	logger := logging.FromContext(ctx)
+
+	srv, err := getGraphqlTestServer()
+	if err != nil {
+		logger.Errorf("unable to initialize graphql server: %v", err)
```
pxp928, line 352: return the error and exit the test if it fails

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,399 @@
...
+	case <-done:
+	case <-time.After(5 * time.Second):
+		logger.Warnf("forcibly shutting down gql http server")
+		cf()
+		server.Close()
+	}
+	cf()
+}
+
+func startTestServer() (*http.Server, *zap.SugaredLogger) {
```
pxp928, line 346: no need to return the logger

pxp928 (approved):
LGTM! We can keep iterating on this as we add the others.

pkg/guacanalytics/patchPlanning.go
lumjjb: +1 probably shouldn't require being passed in since it can be retrieved from `startID`

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+type queueValues struct {
+	path    []string
+	nodeMap map[string]DfsNode
+	now     string
+	nowNode DfsNode
+	queue   []string
+}
+
+// TODO: make more robust using predicates
+func SearchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Client, startID string, stopID string, startType string, maxDepth int) (map[string]DfsNode, error) {
```
lumjjb, line 42: if stopID is optional it should be a pointer instead to indicate if it's required

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+func exploreIsDependency(ctx context.Context, gqlclient graphql.Client, q *queueValues, isDependency model.NeighborsNeighborsIsDependency) (queueValues, error) {
+	dependentPkgFilter := &model.PkgSpec{
+		Type:      &isDependency.DependentPackage.Type,
+		Namespace: &isDependency.DependentPackage.Namespaces[0].Namespace,
+		Name:      &isDependency.DependentPackage.Namespaces[0].Names[0].Name,
+	}
+
+	depPkgResponse, err := model.Packages(ctx, gqlclient, dependentPkgFilter)
```
lumjjb, line 120: I think this is redundant since it should be addressable via `allIsDependencyTreeDependentPackage` embedding of NeighborsNeighborsIsDependency

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		return *q, fmt.Errorf("error querying for dependent package: %w", err)
+	}
+
+	depPkgVersionsMap := map[string]string{}
+	var depPkgVersions []string
+	for _, depPkgVersion := range depPkgResponse.Packages[0].Namespaces[0].Names[0].Versions {
+		depPkgVersions = append(depPkgVersions, depPkgVersion.Version)
+		depPkgVersionsMap[depPkgVersion.Version] = depPkgVersion.Id
+	}
```
lumjjb, line 130: This will probably be incorrect since it is querying for all package versions within the graph given the input filter. This will fail if the input is not that of matching the versionRange of `IsDepedency`.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return q.nodeMap, nil
+
+}
+
+func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueValues, neighbor model.NeighborsNeighborsNode) (queueValues, error) {
+	// case on predicates and nodeType
+	switch q.nowNode.nodeType {
+	case "packageVersion":
```
lumjjb: technically this shouldnt find any `IsDeps` nodes since it attaches to `packageName` for dependents

lumjjb (changes requested):

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,399 @@
...
+		maxDepth       int
+		expectedLen    int
+		graphInput     string
+	}{
+
+		{
+			name:           "1: test case with two dependencies at the same depth, no stopID and no limiting maxDepth",
+			startType:      "deb",
+			startNamespace: ptrfrom.String("ubuntu"),
+			startName:      "dpkg",
```
lumjjb, line 224: start node is missing a version, since it needs to know which version to start off from.

pkg/guacanalytics/patchPlanning.go
rmetzman: Resolved issue by adding a nodeVersions field to the DfsNode to be checked instead.

pkg/guacanalytics/patchPlanning.go
rmetzman: Resolved, changed to packageName.

pkg/guacanalytics/patchPlanning_test.go
rmetzman: Fixed.

pkg/guacanalytics/patchPlanning.go
rmetzman: Resolved, I made everthing public that should be able to be used elsewhere.

pkg/guacanalytics/patchPlanning_test.go
rmetzman: I added an expected outputs field to resolve this.

pkg/guacanalytics/patchPlanning.go
rmetzman: Fixed.

pkg/guacanalytics/patchPlanning.go
rmetzman: So in the end I think you would require both the version and the name to be added because some relationships like isDependencies are attached at the packageName level while others like isOccurrences are attached at the package version and both may be relevant for exploring the graph.

pkg/guacanalytics/patchPlanning.go
rmetzman: Fixed (changed in tests as well).

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if len(nodePkg.AllPkgTree.Namespaces[0].Names[0].Versions) < 1 {
+		var versionsList []string
+		for _, versionEntry := range nodePkg.AllPkgTree.Namespaces[0].Names[0].Versions {
+			versionsList = append(versionsList, versionEntry.Version)
+		}
+		q.nodeMap[startID] = DfsNode{
+			NodeType:     "packageName",
+			nodeVersions: versionsList,
```
lumjjb, line 77: nit: add a TODO here to note that it needs to handle the case where there's circular depedencies that introduce more versions to the version list on a node that requires revisiting

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,187 @@
...
+	return nil
+}
+
+func exploreIsDependency(ctx context.Context, gqlclient graphql.Client, q *queueValues, isDependency model.NeighborsNeighborsIsDependency) error {
+	doesRangeInclude, err := depversion.DoesRangeInclude(q.nowNode.nodeVersions, isDependency.VersionRange)
+
+	if err != nil {
+		return err
+	}
+	if doesRangeInclude {
```
lumjjb, line 155: nit: to make it more readable, you can do the inverse for block which include everything. So for this, something like

```
if !doesRangeInclude {
    return nil
}

// do other stuff assuming doesRangeInclude==true
```

pkg/guacanalytics/patchPlanning.go
```diff
@@ -0,0 +1,187 @@
...
+			if err != nil {
+				return err
+			}
+		}
+	}
+
+	return nil
+}
+
+func exploreIsDependency(ctx context.Context, gqlclient graphql.Client, q *queueValues, isDependency model.NeighborsNeighborsIsDependency) error {
```
lumjjb, line 149: nit `exploreIsDependencyFromDepPkg`

pkg/guacanalytics/patchPlanning_test.go
```diff
@@ -0,0 +1,481 @@
...
+	for _, tt := range testCases {
+		t.Run(fmt.Sprintf("Test case %s\n", tt.name), func(t *testing.T) {
+			err = ingestTestData(ctx, gqlclient, tt.graphInput)
+
+			if err != nil {
+				t.Errorf("Error ingesting test data: %s", err)
+				return
+			}
+
+			var startIDs []*string
```
lumjjb, line 296: this should probably just be a string and then if the return from `getPackageIDs` is more than 1 then it should error out and not continue, then you can make the assumption going forward based on the type and not having to deal with addressing the 0th entry each time, which makes it so you have an implicit contract on the slice throughout the code. 

E.g. instead of *[]string where we need to assert it is non-nil and len==1, if we see *string, the type tells us that it is a singleton

lumjjb (approved):
Minor nits , else LGTM! Thanks @rmetzman !!

PR #979, opened 2023-06-23
getTopLevelPackage will now check DESCRIBES and DESCRIBED_BY relation…

### LINKED ISSUE #708: SPDX parser top level heuristic change to check "DESCRIBES" relationship

CrossedSecurity:
…ships to populate the pUrl. It will fall back to the original method of generating pUrl if neither are available. Added test cases for both of these options.

Fixes #708

This is a fresh branch for https://github.com/guacsec/guac/pull/857

# Description of the PR

# PR Checklist

- [x ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.


pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -175,14 +196,20 @@ func (s *spdxParser) getFileElement(elementID string) []*model.PkgInputSpec {
...
-
 	preds := &assembler.IngestPredicates{}
 
-	toplevel := s.getPackageElement("DOCUMENT")
+	topLevel := s.getPackageElement("DOCUMENT")
 	// adding top level package edge manually for all depends on package
-	if toplevel != nil {
-		preds.IsDependency = append(preds.IsDependency, common.CreateTopLevelIsDeps(toplevel[0], s.packagePackages, s.filePackages, "top-level package GUAC heuristic connecting to each file/package")...)
-		preds.HasSBOM = append(preds.HasSBOM, common.CreateTopLevelHasSBOM(toplevel[0], s.doc))
+	if topLevel != nil {
```
lumjjb, line 203: nit; i think this technically should never be nil now, so maybe we can throw an error if it is

lumjjb (approved):
Minor nits, else LGTM! Thanks @CrossedSecurity for helping with this!

pkg/ingestor/parser/spdx/parse_spdx_test.go
```diff
@@ -46,7 +55,134 @@ func Test_spdxParser(t *testing.T) {
...
+				},
+			},
+			wantPredicates: &assembler.IngestPredicates{
+				HasSBOM: []assembler.HasSBOMIngest{
+					{Pkg: pUrlToPkgDiscardError("pkg:guac/spdx/sbom-sha256%3Aa743268cd3c56f921f3fb706cc0425c8ab78119fd433e38bb7c5dcd5635b0d10")},
+				},
+			},
+			wantErr: false,
+		},
+	}
```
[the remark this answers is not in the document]
CrossedSecurity: done

pkg/ingestor/parser/spdx/parse_spdx.go
CrossedSecurity: This method doesn't return an error so I'll leave it

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -125,6 +133,19 @@ func (s *spdxParser) getPackages() error {
 		}
 
 	}
+
+	// If there is no top level Spdx Id that can be derived from the relationships, we take a best guess for the SpdxId.
+	if _, ok := s.packagePackages[string(s.spdxDoc.SPDXIdentifier)]; !ok {
```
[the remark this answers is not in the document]
CrossedSecurity: done

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -175,14 +205,20 @@ func (s *spdxParser) getFileElement(elementID string) []*model.PkgInputSpec {
...
 	// adding top level package edge manually for all depends on package
-	if toplevel != nil {
-		preds.IsDependency = append(preds.IsDependency, common.CreateTopLevelIsDeps(toplevel[0], s.packagePackages, s.filePackages, "top-level package GUAC heuristic connecting to each file/package")...)
-		preds.HasSBOM = append(preds.HasSBOM, common.CreateTopLevelHasSBOM(toplevel[0], s.doc))
+	if topLevel != nil {
+		for _, topLevelPkg := range topLevel {
+			preds.HasSBOM = append(preds.HasSBOM, common.CreateTopLevelHasSBOM(topLevelPkg, s.doc))
+		}
+
+		if s.topLevelIsHeuristic {
```
pxp928, line 217: Will have to do something similar to CDX also. Currently we are just creating the `CreateTopLevelIsDeps` without any checks in the dependencies.

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -166,6 +189,13 @@ func (s *spdxParser) getPackageElement(elementID string) []*model.PkgInputSpec {
 	return nil
 }
 
+func (s *spdxParser) getTopLevelPackageElement(elementID string) []*model.PkgInputSpec {
+	if packNode, ok := s.topLevelPackages[string(elementID)]; ok {
+		return packNode
+	}
+	return nil
```
pxp928, line 196: This should be an error message. As if there is no top-level package, we should not proceed. We are doing this in CDX as well: https://github.com/guacsec/guac/blob/165ec24a416eac6ba94924982a939e3954710235/pkg/ingestor/parser/cyclonedx/parser_cyclonedx.go#L110-L112

lumjjb: This is for the name case in SPDX Syft, i think we still want to have the same behavior?

pkg/ingestor/parser/spdx/parse_spdx.go
CrossedSecurity: https://github.com/guacsec/guac/issues/996

pkg/ingestor/parser/spdx/parse_spdx.go
CrossedSecurity: Ok, since GetPredicates does not return an error I logged an error message and returned empty predicates.

This code path isn't actually reachable today, but I didn't want to put a panic because if something changes in the future and GUAC parses a bad SBOM (likely) it would crash guac

pxp928 (approved):

Issue #975, opened 2023-06-22
[feature] Generic Package and Source metadata


lumjjb:
There have been some requests for being able to have custom metadata for packages and source. These can be custom metadata that is used in an organization or perhaps not yet implemented by the GUAC ontology. 

Change the graphQL interface to introduce metadata nodes that will attach to packages, source or artifacts. These metadata will have a name type, value type, and an optional timestamp for the metadata.

PR #1002, opened 2023-06-30
add generic metadata gql

lumjjb:
# Description of the PR

Part 1 of #975 

# PR Checklist


mihaimaruseac (approved):

pkg/assembler/graphql/schema/metadata.graphql
```diff
@@ -0,0 +1,44 @@
...
+The metadata applies to a subject which is a package, source, or artifact.
+If the attestation targets a package, it must target a PackageName or a
+PackageVersion. If the attestation targets a source, it must target a
+SourceName.
+"""
+type HasMetadata {
+  id: ID!
+  subject: PackageSourceOrArtifact!
+  justification: String!
+  key: String!
```
pxp928, line 40: should this be an array of key-value pairs? similar to how we have `PackageQualifier`?

lumjjb: i think it would be different nodes, and they would collapse to be evaluated since they could come from multiple sources.

.github/scripts/excluded_from_copyright
```diff
@@ -48,5 +48,6 @@
 ./pkg/assembler/graphql/resolvers/certifyVEXStatement.resolvers.go
 ./pkg/assembler/graphql/resolvers/path.resolvers.go
 ./pkg/assembler/graphql/generated/path.generated.go
+./pkg/assembler/graphql/generated/metadata.generated.go
```
lumjjb, line 51: Adding a comment to block the PR.. I just realized that timestamp probably will be important here, thoughts @pxp928 @mihaimaruseac

pkg/assembler/graphql/schema/metadata.graphql
pxp928: What if one source has a bunch of data? Just thinking if it will cause number is nodes to grow unnecessarily.

.github/scripts/excluded_from_copyright
pxp928: +1

mihaimaruseac: Yeah, let's record timestamps

pkg/assembler/graphql/schema/metadata.graphql
mihaimaruseac: We don't really have good support for arbitrary lists in nodes in the backend. In GraphQL we can make one single type that contains a list, but in the backend we will need to implement these separately

pxp928: Why not do something similar to package qualifiers and hassbom annotations. Seems like having multiple key value pairs per source makes more sense.

mihaimaruseac: We still cannot search for all packages that have a certain list of qualifiers and others (like search for "arch=some_value" but also get package that has "os=windows" as an additional qualifier)

pxp928: If that is the case, we should remove all key/value pair lists from graphQL?

mihaimaruseac: Here it's more likely to want to search for a single key, than in the package qualifier.

pxp928: hmm...we should remove annotations from hasSBOM. This can be used instead. I will make a PR for that, as that will also be something we would want to search for using a single key.

pxp928 (approved):

pkg/assembler/graphql/schema/metadata.graphql
pxp928: created issue https://github.com/guacsec/guac/issues/1018

.github/scripts/excluded_from_copyright
lumjjb: done!

PR #1021, opened 2023-07-05
add support for using Tilt for local development

j-white:
# Description of the PR

Here we add a Tiltfile and supporting resources to allow using Tilt for local development:
* add K8S manifests for basic stack
* enable the use of environment variables in Viper: prefixed with GUAC_
* add an HTTP based healthcheck to the GQL service
* integrate Tilt with Kind in GHA for CI

Can be used like this:
```
% kind create cluster
% tilt up
````

And then all the Tilt fun:
image

# PR Checklist

### CODE REVIEW

File: .github/workflows/ci.yaml [hunk 1]
```diff
@@ -19,6 +19,7 @@ on:
   push:
     branches:
       - main
+      - features/tilt-dev
```
lumjjb, line 22: hmm, is this a development artifact or does this have another intent?

cmd/guacgql/cmd/server.go
```diff
@@ -62,6 +62,8 @@ func startServer(cmd *cobra.Command) {
 		srv.Use(tracer)
 	}
 
+	http.HandleFunc("/healthz", healthHandler)
```
lumjjb, line 65: 🚀

lumjjb (commented):
awesome! Thanks for the PR! This looks great! 

@sunnyyip looks like there's a change in the env variable prefixes, does this affect your helm deployment?

If this PR merges, would you be able to add some documentation in https://github.com/guacsec/guac-docs as well?

pxp928:
This is very cool! Thanks @j-white

.github/workflows/ci.yaml
j-white: dev artifact, will remove

j-white:
> If this PR merges, would you be able to add some documentation in https://github.com/guacsec/guac-docs as well?

Of course. Happy to write some docs on how to use it for a local dev loop. I can start a draft pending further review here.

Tiltfile
```diff
@@ -0,0 +1,10 @@
+allow_k8s_contexts('kind-kind')
+analytics_settings(enable=False)
+
+load('ext://helm_remote', 'helm_remote')
+helm_remote('nats', repo_url='https://nats-io.github.io/k8s/helm/charts/', repo_name='nats', values='tilt-nats-values.yaml')
+
+custom_build('local-organic-guac', "export GUAC_IMAGE=\"$EXPECTED_IMAGE\" && make container", deps='pkg/', tag="latest")
```
j-white, line 7: The build could be made faster if we have `goreleaser` only build images for the current architecture instead of all archs. Any ideas if this is possible and what flags we would set?

lumjjb: yea - i do like this suggestion as i also have long build times. @sunnyyip what do you think of maybe having another make command or an ENV variable for building all archs?

sunnyyip: The container build runs on top of "release" which builds **bins** for all os/arch defined in .goreleaser.yaml and copy them into container images. 
goreleaser supports splitting builds on a per GOOS/GOARCH basis but it's a pro feature that won't be available for all users. 
Beyond that we have to edit the .goreleaser.yaml to override building only the current GOOS/GOARCH for each bin. i.e. use the builds.targets in https://goreleaser.com/customization/builds/. This can be messy as we have to manipulate the list of build configs in the yaml and will likely need additional tools installed.

As far as binary goes, we are already building for the host os/arch only using the "[--single-target](https://goreleaser.com/cmd/goreleaser_build/)" flag. 
One workaround I can think of is to use docker build and package the goreleaser output bins into the local container image. This won't be fast for the very first build since pulling the builder images and run build inside it takes time. Subsequent builds will be much faster tho.

sunnyyip:
> awesome! Thanks for the PR! This looks great!
> 
> @sunnyyip looks like there's a change in the env variable prefixes, does this affect your helm deployment?
> 
> If this PR merges, would you be able to add some documentation in https://github.com/guacsec/guac-docs as well?

This won't impact the [Helm Chart](https://github.com/kusaridev/helm-charts/tree/main/charts/guac) as we take config inputs via [guac.yaml as a ConfigMap](https://github.com/kusaridev/helm-charts/blob/main/charts/guac/templates/guac-cm.yaml).

tilt-nats-values.yaml
sunnyyip, line 1: let's put this file in the k8s dir to keep things together?

sunnyyip (commented):
LGTM. Thanks @j-white for the PR!
Mind if we keep all the k8s related config under the k8s dir?

tilt-nats-values.yaml
j-white: done

Tiltfile
j-white: Thanks @sunnyyip - sounds like there's no easy way now. We can optimize this later :)

j-white:
I consolidated the k8s bits and rebased on `main`.
Draft PR for docs up here: https://github.com/guacsec/guac-docs/pull/73

I've been using this while working on a new storage engine, flows well.

sunnyyip (approved):

pxp928 (commented):
LGTM! I will give @lumjjb final approval.

.github/workflows/ci.yaml
```diff
@@ -122,3 +122,85 @@ jobs:
...
         python-version: '3.10'
     - run: ./internal/testing/e2e/e2e
+
+  tilt-ci:
+    name: Run 'tilt ci'
+    runs-on:
+      labels: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      - name: Install tools
```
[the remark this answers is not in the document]
j-white: Built and tested with Mac M1.

k8s/tilt-nats-values.yaml
```diff
@@ -0,0 +1,19 @@
+nats:
```
lumjjb, line 1: qq: can we use the config file? https://github.com/guacsec/guac/blob/main/container_files/nats/js.conf? 

If its hard to use it, can we maybe sync up values to 64M?

lumjjb (approved):
lgtm - minor comment on NATs config

k8s/tilt-nats-values.yaml
j-white: no easy way to re-use that file via the Helm chart for NATS, so I opted to update it instead

pxp928: @j-white just need to sign the commit and should be good to merge!

PR #1036, opened 2023-07-10
Plan out HasSLSA exploration in path planning search

rmetzman:
# Description of the PR

Adding casing and comments to plan out code for HasSLSA predicate for patch planning search method

# PR Checklist


pkg/guacanalytics/patchPlanning.go
```diff
@@ -142,8 +142,18 @@ func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueVal
 				return err
 			}
 		}
-	}
+	// two cases one after the other work like an OR statement
```
lumjjb, line 145: I don't think this is true. https://go.dev/play/p/Adq3QyWdUAy

pkg/guacanalytics/patchPlanning.go
```diff
@@ -190,3 +200,26 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
 
 	return nil
 }
+
+// TODO: impelement this functions
+func exploreIsOccurrence(ctx context.Context, gqlclient graphql.Client, q *queueValues, isOccurrence model.NeighborsNeighborsIsOccurrence) error {
+	// Step 1: Find Artifact attached to package through IsOccurence
+	// -> call .Artifact on isOccurrence
```
lumjjb, line 207: Do we need to call `query Artifact` here?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -190,3 +200,26 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
+	// Step 2: Find HasSLSA where Artifact is the builtFrom
```
lumjjb, line 208: shouldn't this be `exploreHasSLSA` instead?

pkg/guacanalytics/patchPlanning.go
```diff
@@ -190,3 +200,26 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
+	// -> call .Neighbors on the artifact id with the edge type specified as HasSLSAArtifact and loop through results
```
lumjjb, line 209: Calling `.Neighbors` and acting upon it sounds a lot like the main loop, what is the special case here?

pkg/guacanalytics/patchPlanning.go
rmetzman: Do you mean I should make a helper called exploreHasSLSA or that the func this is in should be called exploreHasSLSA instead of exploreIsOccurrence?

pkg/guacanalytics/patchPlanning.go
rmetzman: I believe if I simply input isOccurrence.Artifact.Id into .Neighbors that would be sufficent to access the artifact we need.

pkg/guacanalytics/patchPlanning.go
rmetzman: The main loop found the IsOccurrence (say this is called A) attached to the original package or source but here we need to call .Neighbors twice more to 
1) Find the HasSLSA connected to the artifact connected to A
2) Find the IsOccurrence connected to the Artifact that is the subject of the HasSLSA

In short, there is a longer chain to traverse in the case of an IsOccurrence to reach the relevent package affected.

It would look something like this:

Package (or Source) ----> isOccurrence (A) ----> Artifact ----> HasSLSA (B) ----> Artifact (the subject of the SLSA) ----> isOccurrence (C) ----> Package (or Source)

The first .Neighbors is for isOccurrence A where the second/third .Neighbors are for B and C.  I was planning to do this in order to get from the original Package (or Source) to the Package (or Source) at the end of the chain that needs to be added to the queue.

pkg/guacanalytics/patchPlanning.go
rmetzman: Got it.  I was confusing with java I believe where this may work (https://stackoverflow.com/questions/18412936/or-operator-in-switch-case).

I'll use a comma instead with is Go's syntax for this I believe.

pkg/guacanalytics/patchPlanning.go
lumjjb: why do we need to call `Artifact` here though?

pkg/guacanalytics/patchPlanning.go
lumjjb: How about the case of `PkgEqual`, would `PkgEqual` also explore `IsDepedency`?

pkg/guacanalytics/patchPlanning.go
rmetzman: I want to call .Artidact so I can reach the Artifact so that we can find any HasSLSAs it is connected to, to find the relevent affected packages

pkg/guacanalytics/patchPlanning.go
rmetzman: I haven't planned out or taken into account in the code a PkgEqual case but in theory it should

PR #1022, opened 2023-07-05
Add HasSLSA test case(s)

rmetzman:
# Description of the PR

Add HasSLSA test case(s) to check understanding of the problem before implementing code for this predicate

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [x] All new changes are covered by tests
- [x] If GraphQL schema is changed, `make generate` has been run
- [x] If `collectsub` protobuf has been changed, `make proto` has been run
- [x] All CI checks are passing (tests and formatting)
- [x] All dependent PRs have already been merged

PR #1046, opened 2023-07-11
Implement HasSLSA predicate case + tests (patchPlanning search method)

rmetzman:
# Description of the PR

This PR add functionality and test cases to the search fucntion to deal with the HasSLSA predicate (coming from an artifact attached to a package) for the purpose of patchPlanning

- Handles artifact node types
- Handles isOccurrence relationships
- Handles hasSLSA relationships

Based off of pseudocode in Draft PR #1036
Based off of tests in Draft PR #1022

# PR Checklist


pkg/guacanalytics/patchPlanning.go
```diff
@@ -28,8 +28,9 @@ type DfsNode struct {
 	Expanded     bool // true once all node neighbors are added to queue
 	Parent       string
 	Depth        int
-	NodeType     string   // packageName, packageVersion
-	nodeVersions []string // for a packageName, what was the packageVersion associated with this version
+	NodeType     string   // packageName, packageVersion, sourceName, artifact
```
lumjjb, line 31: may be good to specify enum here

pkg/guacanalytics/patchPlanning.go
```diff
@@ -158,35 +171,49 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
-			Parent:       q.now,
-			Depth:        q.nowNode.Depth + 1,
-			NodeType:     "packageName",
-			nodeVersions: []string{isDependency.Package.Namespaces[0].Names[0].Versions[0].Version},
+func exploreIsOccurrenceFromSubject(ctx context.Context, gqlclient graphql.Client, q *queueValues, isOccurrence model.NeighborsNeighborsIsOccurrence) {
+	addNodeToQueue(q, q.now, q.nowNode.Depth+1, "artifact", false, nil, isOccurrence.Artifact.Id)
+}
+
+func exploreHasSLSAFromArtifact(ctx context.Context, gqlclient graphql.Client, q *queueValues, hasSLSA model.NeighborsNeighborsHasSLSA) {
+	addNodeToQueue(q, q.now, q.nowNode.Depth+1, "artifact", true, nil, hasSLSA.Subject.Id)
```
lumjjb, line 185: probably need to make sure that it is coming from the materials and not the subject here since neighbor doesn't distinguish if its the subject or the materials

pkg/guacanalytics/patchPlanning.go
```diff
@@ -158,35 +171,49 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
+func exploreIsOccurrenceFromArtifact(ctx context.Context, gqlclient graphql.Client, q *queueValues, isOccurrence model.NeighborsNeighborsIsOccurrence) {
+	// TODO: take into account PkgEqual case where there is another alias of the artifact package that we don't know about
+	// Right now we only explore isOccurrences if their artifact is from a HasSLSA
+	if q.nowNode.slsaSubject {
+		switch subject := isOccurrence.Subject.(type) {
+		case *model.AllIsOccurrencesTreeSubjectPackage:
+			addNodeToQueue(q, q.now, q.nowNode.Depth+1, "packageVersion", false, []string{subject.Namespaces[0].Names[0].Versions[0].Version}, subject.Namespaces[0].Names[0].Versions[0].Id)
+			addNodeToQueue(q, q.now, q.nowNode.Depth+1, "packageName", false, []string{subject.Namespaces[0].Names[0].Versions[0].Version}, subject.Namespaces[0].Names[0].Id)
+		case *model.AllIsOccurrencesTreeSubjectSource:
+			addNodeToQueue(q, q.now, q.nowNode.Depth+1, "sourceName", false, nil, subject.Namespaces[0].Names[0].Id)
```
lumjjb, line 197: May be good to have this be done by the helper function since it is consistently plus 1, else there's a case where it isn't.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -130,10 +140,10 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Clie
 
 }
 
-func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueValues, neighbor model.NeighborsNeighborsNode, nodeType string) error {
+func caseOnPredicates(ctx context.Context, gqlclient graphql.Client, q *queueValues, neighbor model.NeighborsNeighborsNode, Type NodeType) error {
```
lumjjb, line 143: camlCase usually for function variables.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -158,35 +180,52 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
+			addNodeToQueue(q, q.now, q.nowNode.Depth, SourceName, false, nil, subject.Namespaces[0].Names[0].Id)
 		}
-		q.nodeMap[isDependency.Package.Namespaces[0].Names[0].Id] = dfsNName
 	}
+}
 
-	if !dfsNVersion.Expanded {
-		q.queue = append(q.queue, isDependency.Package.Namespaces[0].Names[0].Versions[0].Id)
-	}
+func addNodeToQueue(q *queueValues, parent string, depth int, Type NodeType, slsa bool, versions []string, id string) {
```
lumjjb, line 214: this could be a pointer receiver for  the queue struct, so you don't have to pass in all the variables.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -158,35 +180,52 @@ func exploreIsDependencyFromDepPkg(ctx context.Context, gqlclient graphql.Client
...
-	if !seenName {
-		dfsNName = DfsNode{
-			Parent:       q.now,
-			Depth:        q.nowNode.Depth + 1,
-			NodeType:     "packageName",
-			nodeVersions: []string{isDependency.Package.Namespaces[0].Names[0].Versions[0].Version},
+func exploreIsOccurrenceFromArtifact(ctx context.Context, gqlclient graphql.Client, q *queueValues, isOccurrence model.NeighborsNeighborsIsOccurrence) {
+	// TODO: take into account PkgEqual case where there is another alias of the artifact package that we don't know about
+	// Right now we only explore isOccurrences if their artifact is from a HasSLSA
+	if q.nowNode.slsaSubject {
```
lumjjb, line 203: hmm, why is `slsaSubject` used here to detemine if the edge is explored?

rmetzman: slsaSubject is the field I added to denote that an Artifact is linked to the subject of a HasSLSA.

For the moment, the functionality is to only explore an isOccurrence from an artifact if the Artifact is linked to a HasSLSA because then we are exploring packaged and paths no necissarily affected. However, it depends what we want the desired behavior to be if this should be taken into account.  This TODO is to take into account cases that may be affected by this nuance: // TODO: take into account PkgEqual case where there is another alias of the artifact package that we don't know about.

lumjjb: I don't believe this to be true right now. Can you help me understand where this is happening?

"For the moment, the functionality is to only explore an isOccurrence from an artifact if the Artifact is linked to a HasSLSA because then we are exploring packaged and paths no necissarily affected. "

lumjjb: in the diagram that showed me earlier *

lumjjb: by assumption all nodes that are explored from are nodes of interest, which should be the invariant of things in the queue. Would being a slsaSubject change this?

rmetzman: image
For this diagram, pkg4 will not be added to the queue to be explored because the artifact that the isOccurrence is linked to isn't the subject of a HasSLSA.  

This is what I would think should be desired behavior (expect for the one case I have in the TODO) because in this diagram is Artifact1 is produced from Pkg1, it isn't relevent for the search the fact that Artifact1 is also produced by Pkg4 because that wouldn't mean Pkg4 is also affected by the vulnerability too.

rmetzman: Resolved (for this example, pkg4 should be explored because essentially pkg1 and pkg4 are equal and thus is pkg1 is vulnerable so is pkg4).

pkg/guacanalytics/patchPlanning.go
```diff
@@ -60,35 +69,35 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Clie
 
 	// TODO: add functionality to start with other nodes?
 	if len(nodePkg.AllPkgTree.Namespaces) < 1 {
-		return nil, fmt.Errorf("Start by inputting a packageName or packageVerion node")
+		return nil, fmt.Errorf("Start by inputting a packageName or packageVersion node")
```
lumjjb, line 72: nit: start err messages with lower case for consistency

pkg/guacanalytics/patchPlanning.go
lumjjb: i dont believe this is resolved. Refer to https://go.dev/tour/methods/4

rmetzman: I think this is an outdated version of the file, on the latest I have this: 
func addNodeToQueue(q *queueValues, nodeType NodeType, versions []string, id string)
so I think this follows the same pointer receiver for the struct as you linked instead of passing in all the variables?

lumjjb: ok that's fine

pkg/guacanalytics/patchPlanning.go
```diff
@@ -130,24 +139,36 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlclient graphql.Clie
...
-	}
+	case PackageVersion, SourceName:
+		switch neighbor := neighbor.(type) {
+		case *model.NeighborsNeighborsIsOccurrence:
+			exploreIsOccurrenceFromSubject(ctx, gqlClient, q, *neighbor)
+		}
+	case Artifact:
+		switch neighbor := neighbor.(type) {
+		case *model.NeighborsNeighborsHasSLSA:
+			exploreHasSLSAFromArtifact(ctx, gqlClient, q, *neighbor)
```
pxp928, line 162: So is the thought process here that I start with a package name or package version and find every dependency, every artifact that they touch? I thought we wanted to go back up the tree to determine the root node that needs to be updated?

pxp928 (approved):
LGTM. I am confused by what the `SearchDependenciesFromStartNode` is trying to return.

pkg/guacanalytics/patchPlanning.go
rmetzman: So the user would input a package version and or package name that is affected by a vulnerability or malware.  Then it would return every dependency or artifact that is also affected by this vulnerability for the purposed of path planning.  Eventually,  I'll want to return some sort of subgraph and blast radius.  I may change exactly the data structure returned though later so it'll be easier to make a visualizer output from.

pxp928: It should be going back up the dependency tree. Not down it.

lumjjb: intention is to find the blast radius. So getting subjects from materials and SLSA and getting `pkg` from `depPkg` for IsDependency. Does that make more sense in this context?

lumjjb (approved):

pkg/guacanalytics/patchPlanning.go
pxp928:
```
switch nodeType {
	case PackageName:
		switch neighbor := neighbor.(type) {
		case *model.NeighborsNeighborsIsDependency:
			err := exploreIsDependencyFromDepPkg(ctx, gqlClient, q, *neighbor)

			if err != nil {
				return err
			}
		}
	case PackageVersion, SourceName:
		switch neighbor := neighbor.(type) {
		case *model.NeighborsNeighborsIsOccurrence:
			exploreIsOccurrenceFromSubject(ctx, gqlClient, q, *neighbor)
		}
```
Based on the switch statement, it is moving in the right direction. Not an issue.

PR #1049, opened 2023-07-12
Involve `since_time` for collect-sub entry getting.

bobmcwhirter:
Check `since_time` for the getCollectEntries call.


pkg/collectsub/server/db/simpledb/simpledb.go
```diff
@@ -52,15 +53,18 @@ func (s *simpleDb) containsEntry(e *pb.CollectEntry) bool {
 func (s *simpleDb) AddCollectEntries(ctx context.Context, entries []*pb.CollectEntry) error {
 	s.lock.Lock()
 	defer s.lock.Unlock()
+	var asOf int64
```
lumjjb: lint fix
`asOf := time.Now().Unix()` or `var asOf int64 = time.Now().Unix()`

pkg/collectsub/collectsub/collectsub.pb.go
```diff
@@ -98,6 +98,7 @@ type CollectEntry struct {
 
 	Type  CollectDataType `protobuf:"varint,1,opt,name=type,proto3,enum=guacsec.guac.collect_subscriber.schema.CollectDataType" json:"type,omitempty"`
 	Value string          `protobuf:"bytes,2,opt,name=value,proto3" json:"value,omitempty"`
+	AsOf  int64           `protobuf:"int64,3,opt,name=as_of,proto3" json:"as_of,omitempty"`
```
lumjjb, line 101: this file should be accompanied with the collectsub.proto file change and this change should propagate with running  `make proto`. If you don't have the proto utils set up i can help you generate the code and send it over as well.

lumjjb (commented):
minor nits, should be good otherwise! Thanks for adding this, super useful!

pkg/collectsub/server/db/simpledb/simpledb.go
lumjjb: resolved

pkg/collectsub/collectsub/collectsub.pb.go
pxp928: I don't think there is a need to change this as it is already part of the filters.

bobmcwhirter: Installed and `make proto`'d now.

pxp928: yeah, there is no reason to change the proto file. It already had the time value, just not being utilized.

pkg/collectsub/server/db/simpledb/simpledb.go
```diff
@@ -52,15 +53,17 @@ func (s *simpleDb) containsEntry(e *pb.CollectEntry) bool {
 func (s *simpleDb) AddCollectEntries(ctx context.Context, entries []*pb.CollectEntry) error {
```
pxp928, line 53: This has to be updated to match the interface: `GetCollectEntries(context.Context, []*pb.CollectEntryFilter, int64) ([]*pb.CollectEntry, error)`

pkg/collectsub/collectsub/collectsub.pb.go
lumjjb: yea i think the question is whether `asOf` wants to be part of the proto, which then if the time it was added wants to be returned, it would need to be changed in [CollectEntry](https://github.com/guacsec/guac/blob/main/pkg/collectsub/collectsub/collectsub.proto#L30)

lumjjb (approved):

pxp928 (approved):
LGTM, small nit

pkg/collectsub/collectsub/collectsub.proto
```diff
@@ -30,6 +30,7 @@ enum CollectDataType {
 message CollectEntry {
     CollectDataType type = 1;
     string value = 2;
+    int64 as_of = 3;
```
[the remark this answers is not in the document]
lumjjb: either is fine , we can go with since_time

PR #1052, opened 2023-07-13
Ingest and Query vulnerabilities for arango backend

pxp928:
# Description of the PR

Ingest (including bulk) and Query vulnerabilities for arango backend

# PR Checklist


pkg/assembler/backends/arangodb/ghsa.go
```diff
@@ -0,0 +1,142 @@
...
+		} else {
+			sb.WriteString(string(bs) + ",")
+		}
+	}
+	sb.WriteString("]")
+
+	query := `
+UPSERT { ghsaId:doc.ghsaId } 
+INSERT { ghsaId:doc.ghsaId } 
+UPDATE {} IN ghsas OPTIONS { indexHint: "byGhsaID" }
```
lumjjb, line 88: hmm interesting, did we have to force the use of the index here? curious about that.

pkg/assembler/backends/arangodb/backend.go
```diff
@@ -303,10 +317,18 @@ func GetBackend(ctx context.Context, args backends.BackendArgs) (backends.Backen
 		// repeat this for the collections where an edge is going into
 		hasSBOMEdges.To = []string{hasSBOMsStr}
 
+		var certifyVulnEdges driver.EdgeDefinition
+		certifyVulnEdges.Collection = certifyVulnEdgesStr
+		// define a set of collections where an edge is going out...
+		certifyVulnEdges.From = []string{pkgVersionsStr, certifyVulnsStr}
+
+		// repeat this for the collections where an edge is going into
+		certifyVulnEdges.To = []string{certifyVulnsStr, cvesStr, ghsasStr, osvsStr}
```
lumjjb, line 326: hmm, how would we handle the `NoVuln` case here, thoughts?

lumjjb (approved):

pkg/assembler/backends/arangodb/backend.go
pxp928: oh good call. I missed `noVuln`. Let me update to account for that.

pkg/assembler/backends/arangodb/ghsa.go
pxp928: We should not need to (as its only one field) but did it for consistency.

pkg/assembler/backends/arangodb/backend.go
pxp928: Based on the discussion, `noVuln` will be added as a field on `certifyVuln`. That is not currently implemented by this PR (only adds CVE, GHSA and OSV). Another PR will capture `certifyVuln` along with `noVuln` field within it.

PR #1094, opened 2023-07-25
[WIP] Adding CVE score by version on VulnerabilityMetadataInputSpec

### LINKED ISSUE #1072: [feature] Adding the CVSS/CVE score
The default graphql schema `CertifyVuln` do not contain any information related to CVSS or score  (it exists only in the scorecard).

I would like to access this information in the vulnerability field or even a related `score` mapping. 

```
    vulnerability {
      __typename
      ... on CVE {
        id
        year
        cveId
        score // new cvss field
        severity // new cvss field
      }
```

Having the score for filtering and enumerating the CVEs is a fundamental requirement

knabben:
# Description of the PR

This PR intends to add the CVE score into VulnerabilityMetadata.

NOTE: Inmem changes still pending

Fixes #1072

# PR Checklist

### DIFF
--- a/pkg/assembler/graphql/schema/certifyVuln.graphql
+++ b/pkg/assembler/graphql/schema/certifyVuln.graphql
@@ -99,6 +99,14 @@ type VulnerabilityMetaData {
   origin: String!
   "GUAC collector for the document"
   collector: String!
+  "cvss 2.0 score, defaults to empty string"
+  cvss2Score: Float!
+  "cvss 3.0 score, defaults to empty string"
+  cvss3Score: Float!
+  "additional information, defaults to empty string"
+  additionalInformation: String!
+  "links pointing to other resources, defaults to empty list"
+  links: [String!]!
 }
 
 """
@@ -129,7 +137,16 @@ VulnerabilityMetaDataInput represents the input for certifying vulnerability
 scans in mutations.
 """
 input VulnerabilityMetaDataInput {
-  timeScanned: Time!
+  "cvss 2.0 score, defaults to empty string"
+  cvss2Score: Float!
+  "cvss 3.0 score, defaults to empty string"
+  cvss3Score: Float!
+  "additional information, defaults to empty string"
+  additionalInformation: String!
+  "links pointing to other resources, defaults to empty list"
+  links: [String!]!
+  ollector: String!
+  timeStamp: Time!
   dbUri: String!
   dbVersion: String!
   scannerUri: String!
--- a/pkg/assembler/clients/generated/operations.go
+++ b/pkg/assembler/clients/generated/operations.go
@@ -23869,7 +23869,16 @@ const (
 // VulnerabilityMetaDataInput represents the input for certifying vulnerability
 // scans in mutations.
 type VulnerabilityMetaDataInput struct {
-	TimeScanned    time.Time `json:"timeScanned"`
+	// cvss 2.0 score, defaults to empty string
+	Cvss2Score float64 `json:"cvss2Score"`
+	// cvss 3.0 score, defaults to empty string
+	Cvss3Score float64 `json:"cvss3Score"`
+	// additional information, defaults to empty string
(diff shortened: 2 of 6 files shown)


pkg/assembler/graphql/schema/certifyVuln.graphql
```diff
@@ -99,6 +99,14 @@ type VulnerabilityMetaData {
   origin: String!
   "GUAC collector for the document"
   collector: String!
+  "cvss 2.0 score, defaults to empty string"
```
pxp928, line 102: This should be broken out into its own node that ties to the vulnerability. We should rename the `VulnerabilityMetaData` from `certifyVuln` to something like `scannerMetadata`. Move `VulnerabilityMetaData` into its own graphQL schema file.

pkg/assembler/graphql/schema/certifyVuln.graphql
```diff
@@ -129,7 +137,16 @@ VulnerabilityMetaDataInput represents the input for certifying vulnerability
 scans in mutations.
 """
 input VulnerabilityMetaDataInput {
```
pxp928, line 139: this would be used to create a new `VulnerabilityMetaData` so this is where we would specify the defaults if not specified by the user.  This would move into the new `VulnerabilityMetaData` graphQL definition file but here you would need to add in the defaults:
```
type VulnerabilityMetadataInputSpec {
...
cvss2Score: Float = 0.0
cvss3Score: Float = 0.0
links: [String!] = []
additionalInfo: String = ""
...
```

pkg/assembler/graphql/schema/certifyVuln.graphql
```diff
@@ -129,7 +137,16 @@ VulnerabilityMetaDataInput represents the input for certifying vulnerability
...
+  "cvss 2.0 score, defaults to empty string"
+  cvss2Score: Float!
+  "cvss 3.0 score, defaults to empty string"
+  cvss3Score: Float!
+  "additional information, defaults to empty string"
+  additionalInformation: String!
+  "links pointing to other resources, defaults to empty list"
+  links: [String!]!
+  ollector: String!
+  timeStamp: Time!
```
pxp928, line 149: will also have to add a filter called `VulnerabilityMetaDataSpec` (for consistency, for now, prob need to rename it to filter in the future).

```
type VulnerabilityMetadataSpec {
  id: ID
  vulnerability: Vulnerability
  cvss2Score: Float
  cvss3Score: Float
  timeStamp: Time
  additionalInformation: String
  links: [String!] = []
  origin: String
  collector: String
}
```

Do we need to add in a `matchOnlyEmptyLinks`? Seems like in this case that would not be needed. @mihaimaruseac?

Finally will also have to add the query and mutation similar to the `certifyVuln` but update fr the `Vulnerability Metadata`:

```
extend type Query {
  "Returns all vulnerability certifications matching the input filter."
  CertifyVuln(certifyVulnSpec: CertifyVulnSpec): [CertifyVuln!]!
}

extend type Mutation {
  "Adds a certification that a package has been scanned for vulnerabilities."
  ingestVulnerability(pkg: PkgInputSpec!, vulnerability: VulnerabilityInput!, certifyVuln: VulnerabilityMetaDataInput!): CertifyVuln!
}
```

pxp928 (changes requested):
@knabben Great start! The issue was not clear about this but, we need to break this out into its own node. Otherwise, you are so far on the right track to make updates to the graphQL API!

pkg/assembler/graphql/schema/certifyVuln.graphql
mihaimaruseac: I don't think we need the `machOnlyEmptyLinks` here, since these are not actual identifiers (we can have the same set of links on two separate nodes)

pxp928:
@knabben updated the issue but wanted to state here that we might break the nodes out even more. Will update the issue on a proper path for this.

PR #1095, opened 2023-07-25
[patchPlanning] Make Parent field in BfsNode into a list

rmetzman:
# Description of the PR

Resolve TODO comment to turn parents into a list
Make parents into a list in case a node is discovered from multiple nodes (descendent)
Will be useful for patch planning purposes in order to convert the nodeMap into a subgraph

# PR Checklist


pkg/guacanalytics/patchPlanning.go
```diff
@@ -90,12 +90,14 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlClient graphql.Clie
 		var versionsList []string
 		versionsList = append(versionsList, nodePkg.AllPkgTree.Namespaces[0].Names[0].Versions[0].Version)
 		q.nodeMap[startID] = BfsNode{
-			Type: PackageVersion,
+			Type:   PackageVersion,
+			Parent: []string{""}, // empty string signifies starting node
```
mihaimaruseac, line 94: Thinking out loud: could we mark starting node by empty list of parents instead?

mihaimaruseac (approved):

pkg/guacanalytics/patchPlanning.go
rmetzman: Yes, but I thought maybe it would be useful to have a way of knowing that this node also was added at the start?

mihaimaruseac: I still think you can detect that (`len(node.Parent) == 0` to test if node was from the start).

It will also simplify adding parents, in all cases it will just be `appned(node.Parent, q.now)`, so it can reduce a lot of duplication below

rmetzman: But what if a parent node is then discovered to be connected as a neighbor of one of the other nodes?  Then len(node.Parent) != 0 even if it is the start.

mihaimaruseac: I must misunderstand something. `Parent` is the BFS parent, right? You cannot have a parent for the start node in BFS, since it immediately gets expanded and when you expand a node you only add to the queue the neighbors that are not already expanded.

rmetzman: But technically can't the parent node be the depPkg for let's say nodeA and then nodeA is the depPkg for nodeB but then nodeB is the depPkg for the parent node so you can actually rediscover the parent from another node?  Maybe that wouldn't actually ever happen?

rmetzman: Okay, I believe we have a TODO to deal with cylical dependencies.  I changed the start noded to be empty lists though.

mihaimaruseac (approved):

Issue #1007, opened 2023-07-03
[bug] nil pointer when attempting to `ingestCertifyGood`


bobmcwhirter:
NPE when attempting to `ingestCertifyGood` without a pkgMatchType Spec

**To Reproduce**

While this works:
```
mutation CertifyGoodM1($package: PkgInputSpec!, $origin: String!, $collector: String!, $justification: String!) {
  ingestCertifyGood(
    subject: {
      package: $package
    }
    pkgMatchType: { pkg: SPECIFIC_VERSION }
    certifyGood: {
      justification: $justification
      origin: $origin
      collector: $collector
    }
  ) {
    ...allCertifyGoodTree
  }
}
```

The variant without specifying `pkgMatchType`

```
mutation CertifyGoodM1($package: PkgInputSpec!, $origin: String!, $collector: String!, $justification: String!) {
  ingestCertifyGood(
    subject: {
      package: $package
    }
    certifyGood: {
      justification: $justification
      origin: $origin
      collector: $collector
    }
  ) {
    ...allCertifyGoodTree
  }
}
```

causes an nil pointer.

**Expected behavior**
Either not a nil pointer, or a sensible error indicating that `pkgMatchType` is non-optional.

**Screenshots**

```
github.com/guacsec/guac/pkg/assembler/backends/inmem.(*demoClient).ingestCertifyGood(0x140002188c0, {0x103760fa0, 0x14000357bc0}, {0x140003279a0, 0x0, 0x0}, 0x0, {{0x140000262bc, 0x4}, {0x14000026170, ...}, ...}, ...)
	/Users/bob/repos/trustification/guac/pkg/assembler/backends/inmem/certifyGood.go:82 +0x67c
github.com/guacsec/guac/pkg/assembler/backends/inmem.(*demoClient).IngestCertifyGood(0x140003990b8?, {0x103760fa0?, 0x14000357bc0?}, {0x140003279a0?, 0x0?, 0x0?}, 0x1?, {{0x140000262bc, 0x4}, {0x14000026170, ...}, ...})
	/Users/bob/repos/trustification/guac/pkg/assembler/backends/inmem/certifyGood.go:62 +0x54
```

**GUAC version**

`main#HEAD` as of last week.

**Additional context**

Seems `matchFlags` is deemed optional here https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/certifyGood.graphql#L76

pxp928:
[...]

PR #1115, opened 2023-08-01
update API for bulk ingestion for CertifyBad/CertifyGood, add missing unit tests, update collections on arango

pxp928:
# Description of the PR

- update API for bulk ingestion for CertifyBad/CertifyGood
- add `inmem` bulk ingestion for certifyGood and CertifyBad
- add `inmem` unit tests for bulk ingestion of hashEqual, hasSLSA, hasSBOM, Scorecard, certifyGood, certifybad
- add bulk ingestion for certiyBad on arango backend
- Update collections on arango backend to be specific to each node type for easier querying

Note: `certifyBad` and arango `backend` will have to be expanded out to be reviewed

Fixes https://github.com/guacsec/guac/issues/1007

In the next PR:

- add query for certifyBad
- add query for certifyGood
- add ingestion for certifyGood and bulk certifyGood
- update assembler to use certifyBad/good bulk (with proper separation for
pkg Match flags)

# PR Checklist


cmd/guacgql/cmd/ingest.go
```diff
@@ -741,7 +742,7 @@ func ingestCertifyBad(ctx context.Context, client graphql.Client) {
 	ingestCertifyBad := []struct {
 		name         string
 		pkg          *model.PkgInputSpec
-		pkgMatchType *model.MatchFlags
+		pkgMatchType model.MatchFlags
```
lumjjb, line 745: i assume this change was because we needed batch to force the definition of something per index of the list, can we set a default value for this?

pkg/assembler/graphql/schema/certifyBad.graphql
```diff
@@ -44,6 +44,18 @@ input PackageSourceOrArtifactInput {
...
 }
 
+"""
+PackageSourceOrArtifactInputs allows using PackageSourceOrArtifact union as
+input type to be used in bulk mutations.
+
+Exactly one list must be specified.
+"""
+input PackageSourceOrArtifactInputs {
+  packages: [PkgInputSpec!]
```
lumjjb, line 54: i think matchflags should be here since it is tagged to the input that is being processed. thoughts?

lumjjb (commented):
just a comment around MatchFlags otherwise lgtm

pkg/assembler/graphql/schema/certifyBad.graphql
mihaimaruseac: We could move it here instead of having it be a separate argument value

mihaimaruseac: Actually, no, we need these each time we can match on a package, so not only in this struct. In this case, there's no benefit between having it in each of the input structs or having it as an argument

mihaimaruseac (approved):

pkg/assembler/graphql/schema/certifyBad.graphql
lumjjb: but would you have the case where in one we want to match a version and the other a name??

pxp928: In the case of bulk ingestion, no you could not have one match version and another match name. This also relates to this issue: https://github.com/guacsec/guac/issues/1007 where the user does not specify the `pkgMatchType` and causes a nil pointer error.

cmd/guacgql/cmd/ingest.go
pxp928: Does it make sense to make `pkgMatchType` required field via the API?

cmd/guacgql/cmd/ingest.go
```diff
@@ -49,6 +49,7 @@ func ingestData(port int) {
 	ingestVulnerability(ctx, gqlclient)
 	ingestPkgEqual(ctx, gqlclient)
 	ingestCertifyBad(ctx, gqlclient)
+	ingestCertifyBads(ctx, gqlclient)
```
pxp928, line 52: adding a blocking comment here, while I update.

pxp928: Updated to make `pkgMatchType` mandatory and fix https://github.com/guacsec/guac/issues/1007

mihaimaruseac (approved):

lumjjb (approved):

PR #1130, opened 2023-08-07
:book: Included comments for the bfs on patchPlanning

nathannaveen:
# Description of the PR

* Included comments on `SearchDependenciesFromStartNode`

# PR Checklist


lumjjb (commented):
Besides the one on line 105, i think the others are self-explanatory from the name of the functions/variables and i'm on the preference of omitting the comments as it would help readability.

+1 would be really nice if the BFS part of it can be refactored to its own function.

nathannaveen:
> Besides the one on line 105, i think the others are self-explanatory from the name of the functions/variables and i'm on the preference of omitting the comments as it would help readability.

I agree with some of them, but I think that the one on line 119 is necessary for new contributors because I had a hard time understanding this at first.

Thoughts?
 
> +1 would be really nice if the BFS part of it can be refactored to its own function.

I will refactor it.

lumjjb:
Sounds good!

nathannaveen:
I have removed the extra comments and refactored the BFS.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -132,9 +144,7 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlClient graphql.Clie
 		q.nowNode.Expanded = true
 		q.nodeMap[q.now] = q.nowNode
 	}
-
-	return q.nodeMap, nil
-
+	return nil, nil
```
rmetzman, line 147: why do you return nil for the nodeMap?

nathannaveen: It is retuning `nil` because `q` is being mutated and we return `q` at the end of `SearchDependenciesFromStartNode`.

I just push a small change removing the first return value of `map[string]BfsNode`. The function `bfsOfDependencies` now only returns an error because if there is an error, `SearchDependenciesFromStartNode` will return `nil` and the error, and if there isn’t an error it will return the `q.nodeMap` and `nil`.

pkg/guacanalytics/patchPlanning.go
```diff
@@ -104,6 +104,17 @@ func SearchDependenciesFromStartNode(ctx context.Context, gqlClient graphql.Clie
...
+	err = q.bfsOfDependencies(ctx, gqlClient, stopID, maxDepth)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return q.nodeMap, path, nil
+
+}
+
+// bfsOfDependencies performs a breadth-first search on a graph to find dependencies
```
rmetzman, line 116: I think it is technically bfsOfDependents?

rmetzman: Because it finds the dependents i.e. which packages/artifact/sources are using it.

rmetzman: Maybe you should rename the search function SearchDependentsFromStartNode?  What do you think @lumjjb ?

nathannaveen: Good idea, I have done that.

nathannaveen: I don’t think that it needs to be called `SearchDependentsFromStartNode` because bfs specifies what type of search we are running.

rmetzman: Specifically the bfsOfDependencies should be bfsOfDependents.  In terms of the search change, I am fixing it here #1142.

nathannaveen: @rmetzman, Can this comment be resolved?

lumjjb (approved):

pxp928 (approved):

Issue #1132, opened 2023-08-07
[bug] toposort Loop Not Ending when Graph is Finished


nathannaveen:
**Describe the bug**

On line 30 of pkg/guacanalytics/toposort.go:

https://github.com/guacsec/guac/blob/bc5c042429c7fa93f993cfef3a7ad2b5f4b1059c/pkg/guacanalytics/toposort.go#L30

I think that it shouldn't be `numNodes <= totalNodes`, and should actually be `numNodes < totalNodes`. The reason for this being on line 41:

https://github.com/guacsec/guac/blob/bc5c042429c7fa93f993cfef3a7ad2b5f4b1059c/pkg/guacanalytics/toposort.go#L41-L43

We are checking for whether `len(foundIDs) == 0`. If `numNodes == totalNodes` then we have gone through all the nodes and have no more nodes to find, so `len(foundIDs)` will equal `0`. 

I am not sure whether this is intentional because the error message is `fmt.Errorf("error: cycle detected")`, even though in this case there will be no cycle.

nathannaveen:
cc @rmetzman

rmetzman:
https://github.com/guacsec/guac/pull/1129 This PR fixes the issue but it does not apply the change with the len(parents) from you other fix because the PR hasn't been merged yet.

nathannaveen:
This gets fixed with https://github.com/guacsec/guac/pull/1129.

PR #1131, opened 2023-08-07
Removed Empty String Parents for Root Nodes

nathannaveen:
# Description of the PR

* In pkg/guacanalytics/patchPlanning.go removed the empty string parents for root nodes. The explanation for this is: https://github.com/guacsec/guac/issues/1120#issuecomment-1667069470
* Changed q.now to a pointer
* Fixes #1120

# PR Checklist

- [ ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [ ] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #1133, opened 2023-08-07
Refactored and Included Tests for TopoSortFromBfsNodeMap

nathannaveen:
# Description of the PR

* Refactored the code for TopoSortFromBfsNodeMap for readability and time complexity purposes.
* Included tests for TopoSortFromBfsNodeMap

In regards of improving the time complexity here is what I did:

I changed `parentsMap` from a `map[string] []string` to a `map[string] parent`, with `parent` being: 

``` go
type parent struct {  
	parents map[string]bool // Consider the map[string]bool as a set, the value doesn't matter just the key  
}
```

So basically now `parentsMap = map[string] map[string] bool`. This so that we don't have to loop through every value in `parentsMap`. Doing this change will improve the time complexity by a lot, instead of looping through every value, we will only be doing an `O(1)` operation to remove each `id` from the maps.

I also added a `map[string] []string` called `childrensMap`, it is a map of parents to a list of children. This makes it so that we know which children have which parents. We can use this to only loop through the nodes that have a certain parent and not just loop through each and every node. 

This may be affected by: #1132 

# PR Checklist

### DIFF
--- a/pkg/guacanalytics/toposort.go
+++ b/pkg/guacanalytics/toposort.go
@@ -22,58 +22,65 @@ import (
 	"github.com/Khan/genqlient/graphql"
 )
 
-// TODO: add tests
-func ToposortFromBfsNodeMap(ctx context.Context, gqlClient graphql.Client, nodeMap map[string]BfsNode) (map[int][]string, []string, error) {
-	frontiers := make(map[int][]string)
-	parentsMap, infoNodes := copyParents(nodeMap)
-	frontierLevel := 0
+// TopoSortFromBfsNodeMap sorts the nodes such that it returns a map of level -> list of nodeIDs at that level
+func TopoSortFromBfsNodeMap(ctx context.Context, gqlClient graphql.Client, nodeMap map[string]BfsNode) (map[int][]string, []string, error) {
+	sortedNodes := make(map[int][]string) // map of level -> list of nodeIDs at that level
+	parentsMap, childrensMap, infoNodes := copyParents(nodeMap)
+	// parentsMap: map of nodeID (child) -> list of parents in the form of a map
+	// childrensMap: map of nodeID (parent) -> list of children in the form of a list
+	bfsLevel := 0
 	numNodes := 0
 	totalNodes := len(parentsMap)
 
 	for numNodes < totalNodes {
 		foundIDs := make(map[string]bool)
-		for id, parentsList := range parentsMap {
-			if len(parentsList) == 0 || (parentsList[0] == "" && len(parentsList) == 1) {
-				frontiers[frontierLevel] = append(frontiers[frontierLevel], id)
-				foundIDs[id] = true
+		for id, pMap := range parentsMap {
+			if pMap.parents != nil && len(pMap.parents) == 0 { // if this node has no parents, it is a root node
+				sortedNodes[bfsLevel] = append(sortedNodes[bfsLevel], id)
 				numNodes++
+				foundIDs[id] = true
+				delete(parentsMap, id)
+			}
+		}
+
+		for id := range foundIDs {
+			for _, childID := range childrensMap[id] { // loop through all the children of this node
+				delete(parentsMap[childID].parents, id) // remove this node from the map of parents of the child
 			}
 		}
 
 		if len(foundIDs) == 0 {
-			// TODO: print out offending cycle
-			return frontiers, infoNodes, fmt.Errorf("error: cycle detected")
+			return sortedNodes, infoNodes, fmt.Errorf("error: cycle detected")
 		}
 
-		for id := range foundIDs {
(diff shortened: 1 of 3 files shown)


pkg/guacanalytics/toposort.go
```diff
@@ -22,58 +22,65 @@ import (
...
+	// parentsMap: map of nodeID (child) -> list of parents in the form of a map
+	// childrensMap: map of nodeID (parent) -> list of children in the form of a list
+	bfsLevel := 0
 	numNodes := 0
 	totalNodes := len(parentsMap)
 
 	for numNodes < totalNodes {
 		foundIDs := make(map[string]bool)
-		for id, parentsList := range parentsMap {
-			if len(parentsList) == 0 || (parentsList[0] == "" && len(parentsList) == 1) {
```
nathannaveen, line 36: I removed `|| (parentsList[0] == "" && len(parentsList) == 1)` because https://github.com/guacsec/guac/pull/1131 is handling it.

pxp928:
@rmetzman Did you want to review this one as well?

pkg/guacanalytics/toposort.go
```diff
@@ -22,58 +22,65 @@ import (
...
-				if !foundIDs[parentID] {
-					newParentsList = append(newParentsList, parentID)
-				}
+			for _, childID := range childrensMap[id] { // loop through all the children of this node
+				delete(parentsMap[childID].parents, id) // remove this node from the map of parents of the child
 			}
+		}
 
-			parentsMap[id] = newParentsList
+		if len(foundIDs) == 0 {
```
rmetzman, line 52: Can you add back in the TODO comment please: "// TODO: print out offending cycle"

pkg/guacanalytics/toposort.go
```diff
@@ -22,58 +22,65 @@ import (
...
-		for id, parentsList := range parentsMap {
-			if len(parentsList) == 0 || (parentsList[0] == "" && len(parentsList) == 1) {
-				frontiers[frontierLevel] = append(frontiers[frontierLevel], id)
-				foundIDs[id] = true
+		for id, pMap := range parentsMap {
+			if pMap.parents != nil && len(pMap.parents) == 0 { // if this node has no parents, it is a root node
+				sortedNodes[bfsLevel] = append(sortedNodes[bfsLevel], id)
 				numNodes++
+				foundIDs[id] = true
+				delete(parentsMap, id)
```
rmetzman, line 42: Since you're looping through parentsMap would it be better to delete from the map outside of the loop ?

pkg/guacanalytics/toposort.go
```diff
@@ -22,58 +22,65 @@ import (
...
 	numNodes := 0
 	totalNodes := len(parentsMap)
 
 	for numNodes < totalNodes {
 		foundIDs := make(map[string]bool)
-		for id, parentsList := range parentsMap {
-			if len(parentsList) == 0 || (parentsList[0] == "" && len(parentsList) == 1) {
-				frontiers[frontierLevel] = append(frontiers[frontierLevel], id)
-				foundIDs[id] = true
+		for id, pMap := range parentsMap {
```
rmetzman, line 37: I think the fact that there is both a pMap and a parentsMap variable, it is kind of confusing from the naming to figure out what each represents.   Maybe it would be better to name pMap parents or parentsSet or something along those lines... not sure what the best way around it would be?

pkg/guacanalytics/toposort.go
nathannaveen: Added

pkg/guacanalytics/toposort.go
nathannaveen: Good catch, I have now got it to be deleted when looping through `foundIDs`.

pkg/guacanalytics/toposort.go
nathannaveen: I have renamed `pMap` to `p`. I didn’t rename it to `parents` because calling `parents.parents` would be even more confusing.

rmetzman:
Thank you! LGTM

pxp928 (approved):
LGTM

nathannaveen:
@lumjjb, @mihaimaruseac 👀

pkg/guacanalytics/toposort_test.go
```diff
@@ -0,0 +1,113 @@
...
+		{
+			name: "default",
+			args: args{
+				/*
+						1
+					   / \
+					  2   3
+					 / \   \
+					4   5   6
+				*/
```
lumjjb, line 47: :D

pxp928:
@nathannaveen rebase will auto merge.

nathannaveen:
> @nathannaveen rebase will auto merge.

Done, Thanks!

Issue #1087, opened 2023-07-24
[bug] Crash with version matching / parsing


jeffmendoza:
Problem is here:
https://github.com/guacsec/guac/blob/main/pkg/misc/depversion/depversion.go#L356

This isDependency was added by the deps.dev collector
```json
    "IsDependency": [
      {
        "id": "4360",
        "justification": "dependency data collected via deps.dev",
        "package": {
          "id": "2",
          "type": "npm",
          "namespaces": [
            {
              "id": "6",
              "namespace": "",
              "names": [
                {
                  "id": "476",
                  "name": "readable-stream",
                  "versions": [
                    {
                      "id": "3617",
                      "version": "1.0.34",
                      "qualifiers": [],
                      "subpath": ""
                    }
                  ]
                }
              ]
            }
          ]
        },
        "dependentPackage": {
          "id": "2",
          "type": "npm",
          "namespaces": [
            {
              "id": "6",
              "namespace": "",
              "names": [
                {
                  "id": "774",
                  "name": "string_decoder",
                  "versions": []
                }
              ]
            }
          ]
        },
        "dependencyType": "DIRECT",
        "versionRange": "~0.10.x",
        "origin": "deps.dev",
        "collector": "deps.dev"
      },
```
Note the `"versionRange": "~0.10.x",`. The `getConstraint()` function can handle a version range that starts with `~` or has a wildcard `x`, but not both.

It then tries to pass the wildcard semver to `parseSemver()` and that chokes with

````
panic: runtime error: index out of range [1] with length 0

goroutine 1 [running]:
github.com/guacsec/guac/pkg/misc/depversion.parseSemverHelper(0xc000373cc0, {0xc0004df687, 0x6})
	/home/jeffm/git/guac/pkg/misc/depversion/depversion.go:270 +0xc65
````

https://github.com/guacsec/guac/blob/main/pkg/misc/depversion/depversion.go#L249 Should check that `len(matches) > 0` and return error if not to avoid panic on the index on line 263.

Once this isDep is in your graph, any `guacone query vuln` that tries to traverse will fail.

jeffmendoza:
Here is another dep that causes issues
```json
      {
        "id": "439",
        "justification": "dependency data collected via deps.dev",
        "package": {
          "id": "2",
          "type": "pypi",
          "namespaces": [
            {
              "id": "3",
              "namespace": "",
              "names": [
                {
                  "id": "169",
                  "name": "s3transfer",
                  "versions": [
                    {
[...]

PR #1157, opened 2023-08-15
Fixed Part of SemVer Issue

nathannaveen:
# Description of the PR

* Fixes the first part of https://github.com/guacsec/guac/issues/1087
* Fixed the error for `versionRange` having a tilde `"~"` and wildcard `"x"` (ie. `"versionRange": "~0.10.x"`).
* Also included a test for tilde and wildcard

# PR Checklist

### DIFF
--- a/pkg/misc/depversion/depversion.go
+++ b/pkg/misc/depversion/depversion.go
@@ -159,6 +159,9 @@ var exactSvR = regexp.MustCompile(`^v?(?P<semver>(?P<major>0|[1-9]\d*)(\.(?P<min
 // check for 1.x or 1.0.x cases
 var exactSvRWithWildcard = regexp.MustCompile(`^v?(?P<semver>(?P<major>0|[1-9]\d*)(\.(?P<minor>x|0|[1-9]\d*))?(\.(?P<patch>0|x|[1-9]\d*))?(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)$`)
 
+// check for ~1.x or ~1.0.x cases
+var exactSvRWithWildcardAndTidle = regexp.MustCompile(`~.*[.]x`)
+
 // for bad semvers like v1.0.0rc8 that don't include prerelease dashes
 var almostExactSvR = regexp.MustCompile(`^v?(?P<beforerel>(?P<major>0|[1-9]\d*)\.(?P<minor>0|[1-9]\d*)\.(?P<patch>0|[1-9]\d*))(?P<afterrel>(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)$`)
 
@@ -188,7 +191,7 @@ func isSemver(s string) bool {
 }
 
 func isSemVerWildcard(s string) bool {
-	return !exactSvR.Match([]byte(s)) && exactSvRWithWildcard.Match([]byte(s))
+	return !exactSvR.Match([]byte(s)) && (exactSvRWithWildcardAndTidle.Match([]byte(s)) || exactSvRWithWildcard.Match([]byte(s)))
 }
 
 func isValidConstraint(s string) bool {
@@ -252,36 +255,27 @@ func parseSemver(s string) (semver, major, minor, patch, prerelease, metadata st
 }
 
 func parseSemverHelper(re *regexp.Regexp, s string) (semver, major, minor, patch, prerelease, metadata string, err error) {
-	tildeRegex := regexp.MustCompile(`^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>[xX])$`)
+	matches := re.FindStringSubmatch(s)
 
-	var matches []string
-	var semverIdx, majorIdx, minorIdx, patchIdx, prereleaseIdx, metadataIdx int
+	if len(matches) == 0 {
+		err = fmt.Errorf("did not match regex: %q %s", s, re)
+		return
+	}
+	semverIdx := re.SubexpIndex("semver")
+	majorIdx := re.SubexpIndex("major")
+	minorIdx := re.SubexpIndex("minor")
+	patchIdx := re.SubexpIndex("patch")
+	prereleaseIdx := re.SubexpIndex("prerelease")
+	metadataIdx := re.SubexpIndex("metadata")
 
-	if matches = re.FindStringSubmatch(s); len(matches) > 0 {
-		semverIdx = re.SubexpIndex("semver")
-		majorIdx = re.SubexpIndex("major")
-		minorIdx = re.SubexpIndex("minor")
-		patchIdx = re.SubexpIndex("patch")
-		prereleaseIdx = re.SubexpIndex("prerelease")
-		metadataIdx = re.SubexpIndex("metadata")
-		if semverIdx < 0 {
-			err = fmt.Errorf("unable to find semver")
-			return
-		}
+	if semverIdx < 0 {
+		err = fmt.Errorf("unable to find semver")
+		return
+	}
 
-		semver = matches[re.SubexpIndex("semver")]
-		if semver == "" {
-			err = fmt.Errorf("unable to find semver")
-			return
-		}
-	} else if matches = tildeRegex.FindStringSubmatch(s); len(matches) > 0 {
-		fmt.Println("tilde regex")
-		majorIdx = tildeRegex.SubexpIndex("major") + semverIdx
-		minorIdx = tildeRegex.SubexpIndex("minor") + semverIdx
-		patchIdx = tildeRegex.SubexpIndex("patch") + semverIdx
-		metadataIdx = tildeRegex.SubexpIndex("tilde_x") + semverIdx
-	} else {
-		err = fmt.Errorf("tilde regex failed")
+	semver = matches[re.SubexpIndex("semver")]
+	if semver == "" {
+		err = fmt.Errorf("unable to find semver")
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


nathannaveen:
@jeffmendoza Could you take a look at this?

pkg/misc/depversion/depversion_test.go
```diff
@@ -242,6 +242,14 @@ func Test_VersionRangeParse(t *testing.T) {
 				},
 			},
 		},
+		{
+			input: "~0.10.x",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=,<0.11.0"},
```
lumjjb, line 249: should this be >=0.10.0, <0.11.0?

pkg/misc/depversion/depversion.go
```diff
@@ -252,27 +252,36 @@ func parseSemver(s string) (semver, major, minor, patch, prerelease, metadata st
...
-	semver = matches[re.SubexpIndex("semver")]
-	if semver == "" {
-		err = fmt.Errorf("unable to find semver")
+		semver = matches[re.SubexpIndex("semver")]
+		if semver == "" {
+			err = fmt.Errorf("unable to find semver")
+			return
+		}
+	} else if matches = tildeRegex.FindStringSubmatch(s); len(matches) > 0 {
+		fmt.Println("tilde regex")
```
lumjjb, line 278: remove print statement

pkg/misc/depversion/depversion.go
```diff
@@ -252,27 +252,36 @@ func parseSemver(s string) (semver, major, minor, patch, prerelease, metadata st
...
-	minorIdx := re.SubexpIndex("minor")
-	patchIdx := re.SubexpIndex("patch")
-	prereleaseIdx := re.SubexpIndex("prerelease")
-	metadataIdx := re.SubexpIndex("metadata")
-
-	if semverIdx < 0 {
-		err = fmt.Errorf("unable to find semver")
-		return
-	}
+	tildeRegex := regexp.MustCompile(`^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>[xX])$`)
```
lumjjb, line 255: i'm not sure i see the tilde in this regex?

lumjjb (changes requested):

pkg/misc/depversion/depversion.go
nathannaveen: This is my first time writing regex, so I had some issues. I believe that I've managed to rectify all problems, but please point out any errors you might come across.

***

Previously when `s` hit line `337` (Currently, outside of this commit it is line `334`) it would match the regex, but with `s` having a tilde and a wildcard, it wouldn't. It would continue until it hit line `370` (Currently, outside of this commit it is line `367`) where it would check for a tilde. The issue with this is that `s` has a wildcard, but is skipping the wildcard check.

To remedy this I decided to update `isSemVerWildcard`. 

In order to do this I have created a new regex that checks for a tilde and than a `".x"` anywhere else in the string:

``` go
var exactSvRWithWildcardAndTidle = regexp.MustCompile(`~.*[.]x`) 
```

Then inside the function `isSemVerWildcard` I am checking `exactSvRWithWildcardAndTidle || exactSvRWithWildcard`. The `exactSvRWithWildcardAndTidle` checks cases that have a tilde AND a `"x"` (ie. `"~0.10.x"`), while `exactSvRWithWildcard` checks cases that only have a wildcard (ie. `"0.10.x"`). 

Note that `exactSvRWithWildcardAndTidle` won't be checking cases with just a tilde and no wildcard, `"~0.10.0"`. 

***

I also removed everything I did to `parseSemverHelper` because I am now just trimming off the beginning tilde.

pkg/misc/depversion/depversion_test.go
nathannaveen: Thanks, fixed!

pkg/misc/depversion/depversion.go
nathannaveen: Thanks, I forgot about that. I have also removed all changes from that function `parseSemverHelper`. Reason for it in the other conversation https://github.com/guacsec/guac/pull/1157#discussion_r1296433558.

pkg/misc/depversion/depversion.go
```diff
@@ -159,6 +159,9 @@ var exactSvR = regexp.MustCompile(`^v?(?P<semver>(?P<major>0|[1-9]\d*)(\.(?P<min
 // check for 1.x or 1.0.x cases
 var exactSvRWithWildcard = regexp.MustCompile(`^v?(?P<semver>(?P<major>0|[1-9]\d*)(\.(?P<minor>x|0|[1-9]\d*))?(\.(?P<patch>0|x|[1-9]\d*))?(?:-(?P<prerelease>(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+(?P<buildmetadata>[0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?)$`)
 
+// check for ~1.x or ~1.0.x cases
+var exactSvRWithWildcardAndTidle = regexp.MustCompile(`~.*[.]x`)
+
```
jeffmendoza, line 164: This regex looks too simplistic to catch what we want. I'd rather just test for and remove tilde if found. (below)

pkg/misc/depversion/depversion.go
```diff
@@ -188,7 +191,7 @@ func isSemver(s string) bool {
 }
 
 func isSemVerWildcard(s string) bool {
-	return !exactSvR.Match([]byte(s)) && exactSvRWithWildcard.Match([]byte(s))
+	return !exactSvR.Match([]byte(s)) && (exactSvRWithWildcardAndTidle.Match([]byte(s)) || exactSvRWithWildcard.Match([]byte(s)))
```
jeffmendoza, line 194: Instead of using the regex you added, just first check if the string starts with tilde, and remove if found, then use existing `exactSvRWithWildcard`

pkg/misc/depversion/depversion.go
```diff
@@ -332,7 +335,7 @@ func getConstraint(s string) (string, error) {
 	}
 	// check for 1.x minor and patch versions
 	if isSemVerWildcard(s) {
-		_, major, minor, _, _, _, err := parseWildcardSemver(s)
+		_, major, minor, _, _, _, err := parseWildcardSemver(strings.TrimPrefix(s, "~"))
```
jeffmendoza, line 338: Follow the same pattern already found in this file for consistency:
```
version := strings.TrimPrefix(s, "~")
_, major, minor, _, _, _, err := parseWildcardSemver(version)
```
Or change all the other cases to do the trim inline when passed to the parse function().

Also, add a comment here that we are not expecting ~ in all cases, but when found we just ignore it and assume the wildcard handling is correct.

jeffmendoza (changes requested):
Minor changes.

pkg/misc/depversion/depversion.go
nathannaveen: I have got the code to check for a tilde, and then remove the tilde. Because of this I have removed the regex. I have also combined the wildcard and tilde checks together because they both can start with a tilde.

pkg/misc/depversion/depversion.go
nathannaveen: I have got the code to ignore any tildes that are seen by the wildcard check.

pkg/misc/depversion/depversion.go
nathannaveen: The regex has been removed because they weren't very necessary https://github.com/guacsec/guac/pull/1157#pullrequestreview-1581436663. Instead the code is now ignoring the tilde when checking for a wildcard.

pkg/misc/depversion/depversion_test.go
```diff
@@ -242,6 +242,14 @@ func Test_VersionRangeParse(t *testing.T) {
...
 			},
 		},
+		{
+			input: "~0.10.x",
+			expect: VersionMatchObject{
+				VRSet: []VersionRange{
+					{">=0.10.0,<0.11.0"},
+				},
+			},
+		},
```
lumjjb, line 252: Looks like NPM has some weird cases where you can do things like `^1.3.x` which technically means `>=1.0, <2.0 `which is not obvious, can we help to add these checks and make sure we handle them properly?

```
^1.0.x ==> ">=1.0, <2.0
~1.x ==> ">=1.0, <2.0
^1.0.x ==> ">=1.0, <2.0
```
```

**Review by lumjjb — COMMENTED** [review]
minor request for test, else lgtm

File: pkg/misc/depversion/depversion_test.go (hunk 7 above)
Comment (nathannaveen) [reply pkg/misc/depversion/depversion_test.go:252]: @lumjjb I don’t understand the `^`. How does `^1.3.x` mean `>=1.0,<2.0`?

Comment (lumjjb) [reply pkg/misc/depversion/depversion_test.go:252]: neither do i... it should really be invalid, but unfortunately, that's allowed in NPM so we put it in and ran it and it seemed to behave that way... so those test cases we have above are based on our testing with npm.

Comment (nathannaveen) [reply pkg/misc/depversion/depversion_test.go:252]: So, how would to get these cases to work, should I remove the wildcard, because when I remove the `^`, I just get a regular wildcard case. For example when I run `^1.3.x` I get the output of `">=1.0.0,<1.1.0"`.

Comment (lumjjb) [reply pkg/misc/depversion/depversion_test.go:252]: >For example when I run ^1.3.x after removing the ^, I get the output of ">=1.0.0,<1.1.0".

Do you mean after removing the x? Removing `^` should get ">=1.3.0, <1.4.0" i think.


it looks like between the prefix and the wildcard, it takes the broader range. at least that's what i perceive the behavior to be (not documented unfortunately). 

One way to handle that i think it that if there is a prefix, then for the parsed wildcard, setting that particular version octet to x would handle this case. 

For example for ^1.3.x, major = 1, minor = 3, patch = x, in this case, since "^" is the prefix, set minor = x. 

The other way is to parse both and pick the wider range. That may be easier if there are some helpers to do so in the semver library.

Comment (nathannaveen) [reply pkg/misc/depversion/depversion_test.go:252]: @lumjjb These are my current test cases:

```
^1.0.x -> ">=1.0,<2.0"
^1.x -> ">=1.0,<2.0"
^1.3.x -> ">=1.0,<2.0"
```

Can I change it to be (Including the patch): 

```
^1.0.x -> ">=1.0.0,<2.0.0"
^1.x -> ">=1.0.0,<2.0.0"
^1.3.x -> ">=1.0.0,<2.0.0"
```

Or does it have to stay at only the minor version?

lumjjb: yea that's fine!

nathannaveen: Thanks, done that!

lumjjb (approved):
Awesome! thanks for this PR and helping handle those tricky cases @nathannaveen

PR #1168, opened 2023-08-18
Fixed the nil error in SLSA Parser

naveensrinivasan:
# Description of the PR

- Improve error messages for the `Parse` function
- Add a check for a `nil` document

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/slsa/parser_slsa.go
+++ b/pkg/ingestor/parser/slsa/parser_slsa.go
@@ -22,18 +22,17 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


mihaimaruseac (commented):
Is there a scenario where we'd have the doc be `nil`?

naveensrinivasan:
> Is there a scenario where we'd have the doc be `nil`?

Maybe not. I was using the func and realized that it could be nil.

mihaimaruseac:
It doesn't make sense to add a check for something that we know can never be false.

naveensrinivasan:
> It doesn't make sense to add a check for something that we know can never be false.

OK. I'll close it.

Issue #1050, opened 2023-07-13
[bug] Time of 0 regarded as null when querying time fields


lumjjb:
**Describe the bug**

After ingesting a predicate with a time field of 0 (e.g. "0001-01-01T00:00:00Z"), im getting this message, which shouldn't be the case? I get error "<TRUNCATE> ingestHasSourceAt.knownSince the requested element is null which the schema does not allow\n" when

However, when it is non-zero, the error isnt thrown
        "knownSince": "0001-01-01T00:00:01Z",

**To Reproduce**

Ingest a predicate with a 0 time value

```
mutation hsAMutate {
  
  ingestHasMetadata(subject: {
    source: {
  	type: "git",
    namespace:"github.com/googleapis",
    name:"google-cloud-go"
      }
  },  hasMetadata:{
    key:"key1",
    value:"value1",
    justification:"justification",
    origin:"origin",
    collector: "collector",
    timestamp: "0001-01-01T00:00:00Z"
    
  }) {
    
     __typename
  }
}
```

Try to retrieve it

```
query hSAMeta {
  HasMetadata(
    hasMetadataSpec: {}
  ) {
     __typename
    key
    timestamp
  }
}
```

The result is 
```
{
  "errors": [
    {
      "message": "the requested element is null which the schema does not allow",
      "path": [
        "HasMetadata",
        0,
        "timestamp"
      ]
    }
  ],
  "data": null
}
```

**Expected behavior**
Should not return an error
**Screenshots**

**GUAC version**
main branch

**Additional context**

PR #1169, opened 2023-08-21
changing mutationAPI to only return IDs instead of whole struct to fi…

### LINKED ISSUE #1116: [feature] Change mutation API to remove requirement to return all values except ID
Change mutation APIs to remove requirement to return the entire struct, but instead just IDs. This will make it so that mutations will have better execution time as usually results in INSERTs being used instead of UPSERTs with a followup retrieval if necessary.

A field is still required since GraphQL requires a return statement.

This stems from efforts in the https://github.com/guacsec/guac/pull/910, as well as some optimization efforts for implementation in ArangoDB.

For each mutation like the following example for HasSBOM, go from
```

extend type Mutation {
  "Certifies that a package or artifact has an SBOM."
  ingestHasSBOM(subject: PackageOrArtifactInput!, hasSBOM: HasSBOMInputSpec!): HasSBOM!
  "Bulk ingest that package or artifact has an SBOM."
  ingestHasSBOMs(subjects: PackageOrArtifactInputs!, hasSBOMs: [HasSBOMInputSpec!]!): [HasSBOM!]!
}
```
to
```
extend type Mutation {
  "Certifies that a package or artifact has an SBOM."
  ingestHasSBOM(subject: PackageOrArtifactInput!, hasSBOM: HasSBOMInputSpec!): string!
  "Bulk ingest that package or artifact has an SBOM."
  ingestHasSBOMs(subjects: PackageOrArtifactInputs!, hasSBOMs: [HasSBOMInputSpec!]!): [string!]!
}
```

Return a boolean instead. However IDs are still usable. If the ID is not returnable, returning an empty string is also fine.

**Additional context**

arorasoham9:
…x #1116

# Description of the PR

Initial changes needed for #1116 

# PR Checklist

### DIFF
--- a/pkg/assembler/clients/operations/trees.graphql
+++ b/pkg/assembler/clients/operations/trees.graphql
@@ -116,10 +116,10 @@ fragment AllIsOccurrencesTree on IsOccurrence {
   id
   subject {
     __typename
-    ...on Package {
+    ... on Package {
       ...AllPkgTree
     }
-   ...on Source {
+    ... on Source {
       ...AllSourceTree
     }
   }
@@ -259,7 +259,7 @@ fragment AllCertifyVuln on CertifyVuln {
     ...AllPkgTree
   }
   vulnerability {
-      ...AllVulnerabilityTree
+    ...AllVulnerabilityTree
   }
   metadata {
     dbUri
@@ -339,24 +339,24 @@ fragment AllCertifyVEXStatement on CertifyVEXStatement {
 # }
 
 fragment AllPointOfContact on PointOfContact {
-    id
-    subject {
-      __typename
-      ... on Package {
-        ...AllPkgTree
-      }
-      ... on Source {
-        ...AllSourceTree
-      }
-      ... on Artifact {
-        ...AllArtifactTree
-      }
+  id
+  subject {
+    __typename
+    ... on Package {
+      ...AllPkgTree
     }
+    ... on Source {
+      ...AllSourceTree
+    }
+    ... on Artifact {
+      ...AllArtifactTree
+    }
+  }
 
-    email
-    info
-    since
-    justification
-    origin
-    collector
+  email
+  info
+  since
+  justification
+  origin
+  collector
 }
--- a/pkg/assembler/clients/operations/trees.graphql
+++ b/pkg/assembler/clients/operations/trees.graphql
@@ -315,28 +315,28 @@ fragment AllCertifyVEXStatement on CertifyVEXStatement {
   collector
 }
 
-fragment AllHasMetadata on HasMetadata {
-    id
-    subject {
-      __typename
-      ... on Package {
-        ...AllPkgTree
-      }
-      ... on Source {
-        ...AllSourceTree
-      }
-      ... on Artifact {
-        ...AllArtifactTree
-      }
-    }
+# fragment AllHasMetadata on HasMetadata {
+#     id
+#     subject {
+#       __typename
+#       ... on Package {
+#         ...AllPkgTree
+#       }
+#       ... on Source {
+#         ...AllSourceTree
+#       }
+#       ... on Artifact {
+#         ...AllArtifactTree
+#       }
+#     }
 
-    key
-    value
-    timestamp
-    justification
-    origin
-    collector
-}
+#     key
+#     value
+#     timestamp
+#     justification
+#     origin
+#     collector
+# }
 
 fragment AllPointOfContact on PointOfContact {
     id
--- a/pkg/assembler/clients/operations/trees.graphql
+++ b/pkg/assembler/clients/operations/trees.graphql
@@ -315,28 +315,28 @@ fragment AllCertifyVEXStatement on CertifyVEXStatement {
   collector
 }
 
-# fragment AllHasMetadata on HasMetadata {
-#     id
-#     subject {
-#       __typename
-#       ... on Package {
-#         ...AllPkgTree
-#       }
-#       ... on Source {
-#         ...AllSourceTree
-#       }
-#       ... on Artifact {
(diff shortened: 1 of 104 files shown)


pkg/assembler/clients/operations/trees.graphql
```diff
@@ -298,48 +298,48 @@ fragment AllCertifyVEXStatement on CertifyVEXStatement {
...
-        ...AllPkgTree
-      }
-      ... on Source {
-        ...AllSourceTree
-      }
-      ... on Artifact {
-        ...AllArtifactTree
-      }
-    }
+# fragment AllHasMetadata on HasMetadata {
```
mihaimaruseac, line 301: Why is this commented out?

pkg/assembler/graphql/resolvers/artifact.resolvers.go
```diff
@@ -12,13 +12,24 @@ import (
...
 
 // IngestArtifact is the resolver for the ingestArtifact field.
-func (r *mutationResolver) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (*model.Artifact, error) {
-	return r.Backend.IngestArtifact(ctx, artifact)
+func (r *mutationResolver) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (string, error) {
+	ingestedArtifact, err := r.Backend.IngestArtifact(ctx, artifact)
+	if err != nil {
+		return "", err
+	}
+	return ingestedArtifact.ID, err
```
mihaimaruseac, line 20: This is ok for now, but we still do a lot of work in the resolvers to retrieve the entire struct. We will need to push down and have all resolvers just return the id directly from the backend.

Could be done in separate PR(s).

Also, I think we should not merge this until we do another release, since this is breaking API compatibility.

mihaimaruseac (approved):

pxp928 (commented):
+1 to Mihai, all the backend interfaces will need to change in order for this to be complete: https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/backends.go.

pkg/assembler/graphql/resolvers/artifact.resolvers.go
```diff
@@ -12,13 +12,24 @@ import (
...
 }
 
 // IngestArtifacts is the resolver for the ingestArtifacts field.
-func (r *mutationResolver) IngestArtifacts(ctx context.Context, artifacts []*model.ArtifactInputSpec) ([]*model.Artifact, error) {
-	return r.Backend.IngestArtifacts(ctx, artifacts)
+func (r *mutationResolver) IngestArtifacts(ctx context.Context, artifacts []*model.ArtifactInputSpec) ([]string, error) {
+	ingestedArtifacts, err := r.Backend.IngestArtifacts(ctx, artifacts)
```
pxp928, line 25: Blocker comment to ensure that this does not auto merge. Will need to update the backends.

pkg/assembler/graphql/resolvers/artifact.resolvers.go
arorasoham9: Would you like me to change the backend so that the resolvers can just return the ID directly?

pxp928: We can break that into other PRs and I can help you with some of the backend implementations (such as arango) as they can be optimized.

pxp928: you can start with the `inmem` ones for now.

mihaimaruseac: Given this is huge, I'd say let's merge as it is and then work on the backends.

We will need to convert the resolver for the same functionality at the same time across all backends, given that these are implemented as an interface.

pkg/assembler/clients/operations/trees.graphql
pxp928: @arorasoham9 can you fix this comment and we should be ok to merge.

arorasoham9: fixed.

pxp928: ah so that is a bug. It is missing from the path queries here: https://github.com/guacsec/guac/blob/main/pkg/assembler/clients/operations/path.graphql

So it needs to be added like this:
```
    ... on HasMetadata {
      ...AllHasMetadata
    }
```

Added for the Path query, Neighbors query, Node query and Nodes query. That will resolve this error

pxp928: Fixed!

pkg/assembler/graphql/schema/certifyBad.graphql
```diff
@@ -127,7 +127,15 @@ extend type Query {
 
 extend type Mutation {
   "Adds a certification that a package, source or artifact is considered bad."
```
lumjjb, line 129: minor: if i recall from conversations - the ID is not necessarily a required return value right? in that case, would we set it as ID or at least document that ID can be empty "" (note may need to check if it may trigger a same bug as #1050).

lumjjb (approved):
1 minor qn, else LGTM - thanks so much @arorasoham9, this was really awesome work!

pkg/assembler/graphql/schema/certifyBad.graphql
pxp928: I've implemented some of these changes on the new PRs for bulk ingestion and it does not trigger an error when its set to an empty "". Agree that it should be documented that it can be an empty ""

pxp928: Added comment.

PR #1170, opened 2023-08-21
Feature/add vuln metadata schema

pxp928:
# Description of the PR

Schema for issue https://github.com/guacsec/guac/issues/1072

# PR Checklist


pkg/assembler/clients/operations/vulMetadata.graphql
```diff
@@ -0,0 +1,44 @@
...
+  $vulnMetadata: VulnerabilityMetadataInputSpec!
+  ) {
+  ingestVulnerabilityMetadata(
+    vulnerability: $vulnerability, 
+    score: $score, 
+    vulnerabilityMetadata: $vulnMetadata
+  )
+}
+
+# Defines the GraphQL operations to bulk ingest metadata for a vulnerability into GUAC
```
mihaimaruseac, line 32: Nit: Let's make this comment `"` and attached (no empty line) to the schema. Same for the comment on line 18. This way it gets into the documentation.

mihaimaruseac (approved):

pkg/assembler/graphql/schema/vulMetadata.graphql
```diff
@@ -0,0 +1,120 @@
...
+type: CVSSv2
+value: "5.0"
+
+type: CVSSv3
+value: "7.5"
+"""
+type VulnerabilityScore {
+  id: ID!
+  type: VulnerabilityScoreType!
+  value: String!
```
lumjjb, line 46: hmm thoughts about this being a float? would be easier to do comparisons.

pkg/assembler/graphql/schema/vulMetadata.graphql
```diff
@@ -0,0 +1,120 @@
...
+}
+
+"""
+VulnerabilityScoreSpec allows for filtering a vulnerability score.
+"""
+input VulnerabilityScoreSpec {
+  type: VulnerabilityScoreType
+  value: String
```
lumjjb, line 62: would this be greater than a certain score? or lower? I feel like query vulnerability scores, it would make sense that the queries would be around ranges.

lumjjb: my thoughts is having an enum comparator for >,>=, <,<=,

pxp928: yeah I agree, that would be more useful. Let me add that

pkg/assembler/graphql/schema/vulMetadata.graphql
jeffmendoza: Would a future "score" be something like "high, medium, low" ?

pxp928: Right based on that if it should be string...?

pxp928: Well thinking about this... that would be criticality that we will not be capturing. That will be determined based on the type and score

lumjjb: yea imo i think this is a place where we can be opiniated because it adds utility. anything that doesn't fall under can be put under criticality or any other generic metadata

mihaimaruseac: Could this be CVSS or similar?

pxp928: yes this is for CVSS, EPSS etc...

pkg/assembler/clients/operations/vulMetadata.graphql
pxp928: that seems to error @mihaimaruseac 
`operations/vulMetadata.graphql:18: invalid query-spec file operations/vulMetadata.graphql: Unexpected String "Defines the GraphQL operations to ingest metadata for a vulnerability into GUAC"`

pkg/assembler/graphql/schema/vulMetadata.graphql
pxp928: changed.

pkg/assembler/graphql/schema/vulMetadata.graphql
pxp928: added

pkg/assembler/clients/operations/vulMetadata.graphql
pxp928: We can fix this with all the others in another PR.

pkg/assembler/graphql/schema/vulMetadata.graphql
mihaimaruseac: In that case, it should be string, I think

PR #1184, opened 2023-08-25
[Feature] add ingestion (including bulk) and query for VEX in Arango and inmem

pxp928:
# Description of the PR

- update vex schema for bulk ingestion
- update inmem with bulk ingestion implementation with unit tests
- arango implementation for vex ingestion, bulk ingestion and query
- remove reused code
- fix a couple of typos in backend
- update bulk assembler to use new bulk ingestion for vex

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/arangodb/certifyBad.go
+++ b/pkg/assembler/backends/arangodb/certifyBad.go
@@ -938,8 +938,10 @@ func getCertifyBadFromCursor(ctx context.Context, cursor driver.Cursor) ([]*mode
 			certifyBad.Subject = pkg
 		} else if src != nil {
 			certifyBad.Subject = src
-		} else {
+		} else if createdValue.Artifact != nil {
 			certifyBad.Subject = createdValue.Artifact
+		} else {
+			return nil, fmt.Errorf("failed to get subject from cursor for certifyBad")
 		}
 		certifyBadList = append(certifyBadList, certifyBad)
 	}
--- a/pkg/assembler/backends/arangodb/certifyBad.go
+++ b/pkg/assembler/backends/arangodb/certifyBad.go
@@ -169,7 +169,7 @@ func getSrcCertifyBadForQuery(ctx context.Context, c *arangoClient, arangoQueryB
 	}
 	defer cursor.Close()
 
-	return getSourceCertifyBad(ctx, cursor)
+	return getCertifyBadFromCursor(ctx, cursor)
 }
 
 func getArtCertifyBadForQuery(ctx context.Context, c *arangoClient, arangoQueryBuilder *arangoQueryBuilder, values map[string]any) ([]*model.CertifyBad, error) {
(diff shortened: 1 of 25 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/arangodb/certifyBad.go
```diff
@@ -956,54 +917,29 @@ func getArtifactCertifyBad(ctx context.Context, cursor driver.Cursor) ([]*model.
...
-		src := generateModelSource(createdValue.SrcName.TypeID, createdValue.SrcName.SrcType, createdValue.SrcName.NamespaceID, createdValue.SrcName.Namespace,
-			createdValue.SrcName.NameID, createdValue.SrcName.Name, createdValue.SrcName.Commit, createdValue.SrcName.Tag)
-
-		certifyBad := &model.CertifyBad{
-			ID:            createdValue.CertifyBadID,
-			Subject:       src,
-			Justification: createdValue.Justification,
-			Origin:        createdValue.Collector,
-			Collector:     createdValue.Origin,
+			certifyBad.Subject = createdValue.Artifact
```
lumjjb, line 942: nit: wonder if it would be helpful to check if artifact is nil as well and if so throw an error

pkg/assembler/backends/arangodb/certifyGood.go
```diff
@@ -889,50 +889,11 @@ func (c *arangoClient) IngestCertifyGoods(ctx context.Context, subjects model.Pa
...
-	}
-	return certifyGoodList, nil
-}
-
-func getArtifactCertifyGood(ctx context.Context, cursor driver.Cursor) ([]*model.CertifyGood, error) {
+func getCertifyGoodFromCursor(ctx context.Context, cursor driver.Cursor) ([]*model.CertifyGood, error) {
 	type collectedData struct {
+		PkgVersion    *dbPkgVersion   `json:"pkgVersion"`
 		Artifact      *model.Artifact `json:"artifact"`
+		SrcName       *dbSrcName      `json:"srcName"`
```
lumjjb, line 896: nit: wonder if its possible to do an embedded struct here and so to dedup some of the matching code.

lumjjb (approved):
minor nits, lgtm

pkg/assembler/backends/arangodb/certifyGood.go
pxp928: hmmm, actually can you expand upon this more? I don't understand where to make the changes.

lumjjb: yea so im thinking that the matching against pkg, src or artifact is kind of a repeated pattern in the collectedData struct, so wondering we can embed it such as having
```
Artifact
SrcName
PackageName
PackageVersion
```
as an embedded struct and just including it in each collectedData struct, and then having the parsing function https://github.com/guacsec/guac/pull/1184/files#diff-d396bee1ffad665a5bbc2762f9391519ffda2dd538a5b222131d6b68b1731794R920-R929 be refactored to a function so we don't have to repeat it multiple times.


This would be something probably not within this PR but could be a nice QOL overall since there's quite some room for reuse.

PR #1186, opened 2023-08-28
Included docstrings for parser_csaf

nathannaveen:
# Description of the PR

I am in the process of implementing OpenVEX (Going to create an issue for that). To do this, I am trying to understand CSAF. 

When trying to understand `parser_csaf`, I started including docstrings for myself, but then realized that it could also be helpful for anyone else trying to understand `parser_casf`.

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/csaf/parser_csaf.go
+++ b/pkg/ingestor/parser/csaf/parser_csaf.go
@@ -91,6 +91,11 @@ func (c *csafParser) GetIdentifiers(ctx context.Context) (*common.IdentifierStri
 	return nil, fmt.Errorf("not yet implemented")
 }
 
+// findPurl searches the given CSAF product tree recursively to find the
+// purl for the specified product reference.
+//
+// It recursively calls itself on each child branch of the tree in a
+// depth first search manner if the nodes name isn't equal to product_ref.
 func findPurl(ctx context.Context, tree csaf.ProductBranch, product_ref string) *string {
 	if tree.Name == product_ref {
 		purl := tree.Product.IdentificationHelper["purl"]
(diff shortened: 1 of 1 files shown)


pkg/ingestor/parser/csaf/parser_csaf.go
```diff
@@ -221,43 +273,42 @@ func (c *csafParser) generateVexIngest(ctx context.Context, vulnInput *generated
...
+			for _, product := range products {
+				vi := c.generateVexIngest(ctx, vuln, &v, status, product)
+				if vi == nil {
+					continue
+				}
 
-			statuses := []string{"fixed", "known_not_affected", "known_affected", "first_affected", "first_fixed", "last_affected", "recommended", "under_investigation"}
-			for _, status := range statuses {
-				products := v.ProductStatus[status]
-				if len(products) > 0 {
```
nathannaveen, line 240: Removed redundant check. The `for` loop is accomplishing what the `if` statement is supposed to.

pkg/ingestor/parser/csaf/parser_csaf.go
```diff
@@ -221,43 +273,42 @@ func (c *csafParser) generateVexIngest(ctx context.Context, vulnInput *generated
...
-					for _, product := range products {
```
nathannaveen, line 241: Removed redundant check. The `for` loop accomplishes what the `if len(products) > 0` is supposed to.

pkg/ingestor/parser/csaf/parser_csaf.go
```diff
@@ -156,6 +191,17 @@ func findImpactStatement(tree *csaf.Vulnerability, product_id string) *string {
...
+// given ID in the CSAF document.
+//
+// It first calls findProductRef to get the product reference string for
+// the ID. If not found, returns an error.
+//
+// It then calls findPurl to get the purl for that product reference.
+// If not found, returns an error.
+//
+// Finally, it calls helpers.PurlToPkg to convert the purl to a package
+// spec to return.
```
lumjjb, line 204: Thanks for adding the function comments. 

One note about this though.. the comments for the function should include what the function does (+ any implicit contracts with arguments, state and returns), but not so much as how the function does it - in this case calling a function is a means to do it, but what's really happening is business logic - i.e. someone should be able to implement a different version of the function that adhere's to the spec but uses different underlying code. 

In this case, just the first line should be sufficient

nathannaveen: I have shortened and edited the docstrings so that each maintains an overview of what the functions do.

lumjjb (approved):

PR #1195, opened 2023-08-29
Refactor ingestor code

dejanb:
# Description of the PR

While looking into adding a compression, I noticed that ingestion code has a lot of repetitions in different collectors and services, so thought to try to consolidate things to make it easier to modify as we move forward.

This PR:
* Introduces new `ingestor` package that extracts all common logic into one place
* Introduces `ingestor.Ingest` function that can be used by different collectors to synchronously ingest `Document` using GraphQL
* Changes all existing collectors to use it
* Refactor `guacingest` tool to be able to use both sync and async ingestion (as async ingestion send data over nats multiple times which could be an issue with large files)

# PR Checklist

### DIFF
--- a/cmd/guacingest/cmd/ingest.go
+++ b/cmd/guacingest/cmd/ingest.go
@@ -19,20 +19,17 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"net/http"
 	"os"
 	"os/signal"
 	"sync"
 	"syscall"
 
-	"github.com/Khan/genqlient/graphql"
 	"github.com/guacsec/guac/pkg/assembler"
-	"github.com/guacsec/guac/pkg/assembler/clients/helpers"
 	csub_client "github.com/guacsec/guac/pkg/collectsub/client"
-	"github.com/guacsec/guac/pkg/collectsub/collectsub/input"
 	"github.com/guacsec/guac/pkg/emitter"
 	"github.com/guacsec/guac/pkg/handler/processor"
 	"github.com/guacsec/guac/pkg/handler/processor/process"
+	"github.com/guacsec/guac/pkg/ingestor"
 	"github.com/guacsec/guac/pkg/ingestor/parser"
 	parser_common "github.com/guacsec/guac/pkg/ingestor/parser/common"
 	"github.com/guacsec/guac/pkg/logging"
@@ -44,6 +41,7 @@ type options struct {
 	natsAddr        string
 	csubAddr        string
 	graphqlEndpoint string
+	async           bool
 }
 
 func ingest(cmd *cobra.Command, args []string) {
@@ -52,6 +50,7 @@ func ingest(cmd *cobra.Command, args []string) {
 		viper.GetString("nats-addr"),
 		viper.GetString("csub-addr"),
 		viper.GetString("gql-addr"),
+		viper.GetBool("async-ingest"),
 		args)
 	if err != nil {
 		fmt.Printf("unable to validate flags: %v\n", err)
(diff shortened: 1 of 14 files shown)


pxp928 (commented):
quick nits. Will look into detail. Thanks @dejanb!

cmd/guacingest/cmd/ingest.go
```diff
@@ -80,69 +79,65 @@ func ingest(cmd *cobra.Command, args []string) {
...
+		}
 	}
 
 	// Assuming that publisher and consumer are different processes.
 	var wg sync.WaitGroup
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		if err := processorFunc(); err != nil {
+		if err := process.Subscribe(ctx, emit); err != nil {
```
pxp928, line 112: in this case, even if it's not async, you will still need NATS running for this to work?

pxp928: meaning if `return ingestor.Ingest(ctx, d, opts.graphqlEndpoint, csubClient)` is running, why are we throwing it back into NATS here?

dejanb: Yes, this is the ingestor the receives documents via NATS. The only thing that async flag is doing is preventing it from sending it back via pub/sub via stages. 

So, sync scenario:

`doc -> nats -> ingestor.Ingest`

async scenario:

`doc -> nats -> processor.Process -> nats -> parser.Parse -> assembler`

I'm not sure if async scenario is having much sense any more, but I didn't want to remove it without further discussion.

lumjjb: yea i am in favor of removing async for now unless we run into the need to granularly scale the processor and parser aspects, which i doubt we would need to in the near future.

cmd/guacone/cmd/gcs.go
```diff
@@ -93,46 +93,17 @@ var gcsCmd = &cobra.Command{
...
-			err = collectSubEmitFunc(idstrings)
-			if err != nil {
-				logger.Infof("unable to create entries in collectsub server, but continuing: %v", err)
-			}
+			err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, csubClient)
 
-			err = assemblerFunc(predicates)
 			if err != nil {
 				gotErr = true
-				return fmt.Errorf("unable to assemble graphs: %v", err)
```
lumjjb, line 131: nit :%w (we missed it in the previous code) - multiple occurrences

docker-compose.yml
```diff
@@ -88,7 +88,7 @@ services:
   guac-ingestor:
     networks: [frontend]
     image: $GUAC_IMAGE
-    command: "/opt/guac/guacingest"
```
lumjjb, line 91: can we add this to guac.yaml instead?

lumjjb (approved):
Some nits, else LGTM

dejanb:
@pxp928 @lumjjb thanks for the review. I ended up removing async ingest option per previous comments. It simplify things considerebly.

pkg/cli/store.go
```diff
@@ -71,6 +71,7 @@ func init() {
 	set.Bool("service-poll", true, "sets the collector or certifier to polling mode")
 	set.BoolP("poll", "p", false, "sets the collector or certifier to polling mode")
 	set.StringP("interval", "i", "5m", "if polling set interval, m, h, s, etc.")
+	set.Bool("async-ingest", true, "sets the ingestor in async mode, meaning document will be sent via pub/sub between ingestion stages")
```
pxp928, line 74: this flag is no longer needed

pxp928 (approved):
couple of small nits else LGTM! Thanks @dejanb

dejanb:
> couple of small nits else LGTM! Thanks @dejanb

Thanks @pxp928 . I think all of them should be addressed now

pxp928:
@dejanb Looks like you missed a sign-off on the last commit and it will auto merge

dejanb:
> @dejanb Looks like you missed a sign-off on the last commit and it will auto merge

@pxp928 fixed

pxp928 (approved):

PR #1201, opened 2023-08-29
Remove 32-bit arm build as ent does not support it.

jeffmendoza:
https://github.com/ent/ent/issues/3125

Ent doesn't support 32 bit builds, so remove from goreleaser config.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -71,12 +71,9 @@ builds:
     goarch:
       - amd64
       - arm64
-      - arm
     ignore:
       - goos: windows
         goarch: arm64
-      - goos: windows
-        goarch: arm
   - main: ./cmd/guaccsub
     id: guaccsub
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -71,12 +71,9 @@ builds:
     goarch:
       - amd64
       - arm64
-      - arm
```
pxp928, line 74: there was discussion around this in the original PR and keeping the arm support: https://github.com/guacsec/guac/pull/910#discussion_r1303123801

jeffmendoza: Judging from the ent issue, it seems like it will not work. I'll create an issue to investigate.

jeffmendoza: #1206

jeffmendoza:
Closing in favor of #1226

Issue #1198, opened 2023-08-29
[feature - ENT] Change all backend mutation implementations to return only ID 


pxp928:
**Is your feature request related to a problem? Please describe.**
Followup to issue https://github.com/guacsec/guac/issues/1116. Now that the graphQL mutation  API has been updated, the [backend](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/backends.go) needs to be updated to only return the IDs based on the revised [resolvers](https://github.com/guacsec/guac/tree/main/pkg/assembler/graphql/resolvers)

**Describe the solution you'd like**

Multiple issues will be created for each backend update. This issue will focus on the updates to the [Ent](https://github.com/guacsec/guac/pull/910) backend. 

All mutations must be changed to return only ID (string). 

For example for artifacts:

```
IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (string, error)
IngestArtifacts(ctx context.Context, artifacts []*model.ArtifactInputSpec) ([]string, error) 
```

Must both be updated in the [backend interface](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/backends.go#L54-L55) and in the ENT Backend

Optimization to replace `UPSERT` with `INSERT` must also be done.

Unit tests must be updated as needed to reflect these changes

ivanvanderbyl:
This should be quite easy with Ent, but I think we'll need to continue using upserts unless we handle constraint errors as well, which would require doing a query to return the ID.

Basically 3 ways to do this:

```go
func ingestArtifactID(ctx context.Context, spec *model.ArtifactInputSpec) (string, error) {
	client := ent.TxFromContext(ctx)
	record, err := client.Artifact.Create().Save(ctx)
	if err != nil {
		if ent.IsConstraintError(err) {
			// Do a query to get the ID by converting the artifact spec to a predicate
			// and then querying for it.
			id, err := client.Artifact.Query().Where(artifactQueryInputPredicates(*spec)).OnlyID(ctx)
			if err != nil {
				return "", err
			}
			return nodeID(id), nil
		}

		return "", err
	}

	return nodeID(record.ID), nil
}

func ingestArtifactIDDoNothing(ctx context.Context, spec *model.ArtifactInputSpec) (string, error) {
	client := ent.TxFromContext(ctx)
	id, err := client.Artifact.Create().
		OnConflict(
			sql.ConflictColumns(artifact.FieldDigest),
		).
[...]

PR #1212, opened 2023-09-01
Ent - Vulnerabilities management

mrizzi:
# Description of the PR

Provide implementation for endpoints related to `Vulnerability`, `VulnEqual` and `CertifyVuln` management.

Implementation details:
* `DoNothing()` failed with `sql: no rows in result set` err when running the `TestVulnerability\Duplicates` test (ref.https://github.com/ent/ent/issues/1821). So I've opted for `Ignore()`

* optimized the `IngestVulnerability` output to just populate the `ID` (#1198). This explains also why the `TestIngestVulnerabilities\Multiple` has `exp` which is an array with 3 empty elements

* `NoVuln` type vulnerabilites inserts on purpose a `VulnerabilityID` with an empty `vulnerability_id` (ref. discussion https://openssf.slack.com/archives/C03U677QD46/p1693576695717769)

* tests are the same as `pkg/assembler/backends/inmem/vulnerability_test.go` tests

@ivanvanderbyl WDYT?

# PR Checklist


ivanvanderbyl (commented):
LGTM

jeffmendoza (approved):
Curious on the approach, why do certifyVuln and Vuln have indexes, but vulnEqual does not? Looks like the ingest strategy is different too, where vulnequal does a query first then an insert.

mrizzi:
> Curious on the approach, why do certifyVuln and Vuln have indexes, but vulnEqual does not? Looks like the ingest strategy is different too, where vulnequal does a query first then an insert.

Good question, thanks.

I tried to have a consistent approach between `VulnEqual` and `HashEqual` entities.

For the ingestion process: the `QueryVulnEquals()` func execution is meant to check if there's already an M2M relation between the input vulnerabilties and an `VulnEqual` entity which, AFAIU, it's the way for managing conflicts with an M2M relationship involving two types.
 `HashEqual` ingestion does the same with the `QuerySame` invocation.

The same consistency applies to the indexes.

But I'm not an expert with Ent so any suggestion is more than welcome :+1:

PR #1222, opened 2023-09-05
Add support for sending encoded documents

dejanb:
# Description of the PR

This PR builds on top of #1195 and adds support for sending encoded/compressed documents. It expands `Document` struct with `Encoding` filed.

```
type Document struct {
	Blob              []byte
	Type              DocumentType
	Format            FormatType
	Encoding          EncodingType
	SourceInformation SourceInformation
}
```

which can indicate if content was encoded. At the moment only `bz2` is supported and tested, but the code can be easily expanded to add more popular codecs, gzip, zstd, ...

I've tested these changes by ingesting a large sbom of 230MB (compressed 17MB) and it works fine. Ingestion takes 90sec which is something we can take a look next.

# PR Checklist

### DIFF
--- a/pkg/handler/processor/process/process.go
+++ b/pkg/handler/processor/process/process.go
@@ -38,6 +38,7 @@ import (
 	"github.com/guacsec/guac/pkg/handler/processor/scorecard"
 	"github.com/guacsec/guac/pkg/handler/processor/spdx"
 	"github.com/guacsec/guac/pkg/logging"
+	"github.com/klauspost/compress/zstd"
 )
 
 var (
@@ -211,11 +212,18 @@ func unpackDocument(i *processor.Document) ([]*processor.Document, error) {
 }
 
 func decodeDocument(ctx context.Context, i *processor.Document) error {
+	logger := logging.FromContext(ctx)
 	var reader io.Reader
+	var err error
+	logger.Infof("Decoding document with encoding:  %v", i.Encoding)
 	switch i.Encoding {
 	case processor.EncodingBzip2:
 		reader = bzip2.NewReader(bytes.NewReader(i.Blob))
-	// TODO Add more encoding options
+	case processor.EncodingZstd:
+		reader, err = zstd.NewReader(bytes.NewReader(i.Blob))
+		if err != nil {
+			return fmt.Errorf("unable to create zstd reader: %w", err)
+		}
 	case processor.EncodingUnknown:
 	}
 	if reader != nil {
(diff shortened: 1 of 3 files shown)


pkg/handler/processor/process/process.go
```diff
@@ -140,6 +138,10 @@ func processHelper(ctx context.Context, doc *processor.Document) (*processor.Doc
 }
 
 func processDocument(ctx context.Context, i *processor.Document) ([]*processor.Document, error) {
```
lumjjb, line 140: hmm QQ, was there a reason for the decision to have encoding as separate vs having it as another Format? I think i was thinking about it more like how mime types work

+json+gzip

The unpack function would be able to unwrap the gzip and also have a guesser to test the blob to see if its gzipped, bz2, etc.

lumjjb (commented):
Awesome stuff! Took a quick glance and have a quick qn - will take a more indetph review soon

pkg/handler/processor/process/process.go
dejanb: I was originally thinking the same way, but then it seemed it will complicate things trying to figure out format and encoding from one value. Also, HTTP spec does something similar, like
```
Content-Type:application/json
Content-Encoding: gzip
```
making it more explicit.

On the other hand, after a brief search I didn't find any good guessers for this case. I would love to have that in any case when folks don't provide it explicitly.

I added now zsdt encoding, just to be up to date with what I use in my PoC. But I'm happy to change the API according to the discussion here. If you have a desired way of defining the API and implementing it, can you please describe it here and I'll try to get it running

jeffmendoza:
@dejanb Can you rebase, looks like the new ingestor changes you made need to be merged in.

pkg/handler/processor/process/process.go
lumjjb: hmm if there aren't good guessers (which i'm a little surprised), then this should be fine for now, but its on the onus of the client to provide the encoding.. I think this is good for now, and if we find / develop guessers, perhaps we can streamline it back into the process pipeline again as a format. (if you can create an issue for developing guessers for the content encoding, that would be awesome :))..

dejanb: Thanks @lumjjb. Raised #1237. Even with the current implementation it would be valuable to have a guesser when user provide document with `unknown` encoding.

dejanb:
@jeffmendoza done.

cmd/guacingest/cmd/ingest.go
```diff
@@ -45,6 +46,7 @@ func ingest(cmd *cobra.Command, args []string) {
 		viper.GetString("nats-addr"),
 		viper.GetString("csub-addr"),
 		viper.GetString("gql-addr"),
+		viper.GetBool("async-ingest"),
```
jeffmendoza, line 49: Looks like this arg is not being used now?

pxp928: looks like a rebase issue?

dejanb: @jeffmendoza @pxp928 yes, sorry my mistake. I removed the problematic commit.

jeffmendoza (approved):

pxp928 (approved):

PR #1226, opened 2023-09-05
Disable ent on 32 bit

jeffmendoza:
# Description of the PR

Disable ent on 32 bit

supersedes #1201

# PR Checklist

### DIFF
--- a/cmd/guacgql/cmd/ent.go
+++ b/cmd/guacgql/cmd/ent.go
@@ -1,0 +1,46 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !(386 || arm || mips)
+
+package cmd
+
+import (
+	"context"
+
+	"github.com/guacsec/guac/pkg/assembler/backends"
+	entbackend "github.com/guacsec/guac/pkg/assembler/backends/ent/backend"
+)
+
+func init() {
+	if getBackend == nil {
+		getBackend = make(map[string]backendFunc)
+	}
+	getBackend[ent] = getEnt
+}
+
+func getEnt(ctx context.Context) (backends.Backend, error) {
+	client, err := entbackend.SetupBackend(ctx, entbackend.BackendOptions{
+		DriverName:  flags.dbDriver,
+		Address:     flags.dbAddress,
+		Debug:       flags.dbDebug,
+		AutoMigrate: flags.dbMigrate,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return entbackend.GetBackend(client)
+}
--- a/cmd/guacgql/cmd/ent.go
+++ b/cmd/guacgql/cmd/ent.go
@@ -18,29 +18,22 @@
 package cmd
 
 import (
-	"context"
-
 	"github.com/guacsec/guac/pkg/assembler/backends"
 	entbackend "github.com/guacsec/guac/pkg/assembler/backends/ent/backend"
 )
 
 func init() {
-	if getBackend == nil {
-		getBackend = make(map[string]backendFunc)
+	if getOpts == nil {
+		getOpts = make(map[string]optsFunc)
 	}
-	getBackend[ent] = getEnt
+	getOpts[ent] = getEnt
 }
 
-func getEnt(ctx context.Context) (backends.Backend, error) {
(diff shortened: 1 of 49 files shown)


pxp928 (approved):

cmd/guacgql/cmd/server.go
```diff
@@ -264,3 +204,45 @@ func getAWSRequestSigner() (*v4.Signer, error) {
...
+	return neo4j.GetBackend(args)
+}
+
+func getInMem(ctx context.Context) (backends.Backend, error) {
+	args := &inmem.DemoCredentials{}
+	return inmem.GetBackend(args)
+}
+
+func getNeptune(ctx context.Context) (backends.Backend, error) {
+	// TODO: rename the neo4j config to something more generic since it would be used by Neptune as well.
```
mihaimaruseac, line 233:
```suggestion
```

We no longer need this TODO

mihaimaruseac (approved):
I like this refactoring! :100:

cmd/guacgql/cmd/server.go
```diff
@@ -51,6 +53,20 @@ const (
...
+
+func init() {
+	if getBackend == nil {
+		getBackend = make(map[string]backendFunc)
+	}
+	getBackend[arango] = getArango
+	getBackend[neo4js] = getNeo4j
+	getBackend[inmems] = getInMem
+	getBackend[neptune] = getNeptune
+}
```
lumjjb, line 68: can we maybe move this to the backends pakage that will return this map and instead do that there? I think that may be another appropriate place to put it and we can dedup the code if other CLIs or unit tests need to do this as well.

cmd/guacgql/cmd/ent.go
```diff
@@ -0,0 +1,46 @@
...
+package cmd
+
+import (
+	"context"
+
+	"github.com/guacsec/guac/pkg/assembler/backends"
+	entbackend "github.com/guacsec/guac/pkg/assembler/backends/ent/backend"
+)
+
+func init() {
```
lumjjb, line 27: ahh i wasn't thinking of it this way, but this works as well - i actually thought of doing this in the ent backend package itself.

lumjjb (approved):
LGTM, some comments

cmd/guacgql/cmd/server.go
mihaimaruseac: +1 to this, if possible.

cmd/guacgql/cmd/ent.go
jeffmendoza: Yeah, even if the backend is ignored, the cmd can't have any code that pulls it in, so just do it here.

cmd/guacgql/cmd/server.go
jeffmendoza: Updated, ptal. Still need to setup `entbackend.BackendOptions` from cli flags in a 64-bit-only file.

pkg/assembler/backends/inmem/certifyLegal_test.go
```diff
@@ -485,7 +485,7 @@ func TestLegal(t *testing.T) {
 	ctx := context.Background()
 	for _, test := range tests {
 		t.Run(test.Name, func(t *testing.T) {
-			b, err := inmem.GetBackend(nil)
+			b, err := backends.Get("inmem", nil, nil)
```
pxp928, line 488: Looks good! Just need to update the arango tests with these also

pxp928 (approved):

pkg/assembler/backends/inmem/certifyLegal_test.go
jeffmendoza: done

cmd/guacgql/cmd/server.go
lumjjb: I was also thinking to do the registration within the package, kind of like we do for the parsers, and other pluggable things: 

https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/parser.go#L41-L51

jeffmendoza: Yes, I've added that as you requested in the third commit on this PR:

https://github.com/guacsec/guac/pull/1226/commits/144d04501776e54e6b06d70d4f01c42ab1cef0ff#diff-1aba4420a9017b9d486c0fab38a08b9eea2e9c22e2402039d7726f26b1ccf544R164-R167
https://github.com/guacsec/guac/pull/1226/commits/144d04501776e54e6b06d70d4f01c42ab1cef0ff#diff-84928fb53202e5f8faf758692f063b2919686aec75b8d6d22381bc6e612a3e81R22-R24
And then the call uses `backend.Get` here https://github.com/guacsec/guac/pull/1226/commits/144d04501776e54e6b06d70d4f01c42ab1cef0ff#diff-c5774aa38759e9db3bd03f13cbd58be3e05839da1c2c9b66e9fa2a45f33a5661R138

The cli still needs to prepare the backend-specific options struct, and also cannot import the entbackend options struct on 32-bit (as that would pull in the whole package).

It's a bit of a false encapsulation as the options structs need to be known, I'm not sure it is better than how it was previously.

lumjjb: Ah gotcha, i think that's good for now then. I do have an idea that could maybe help a bit but not blocking.

PR #1228, opened 2023-09-07
update SPDX parser to skip empty and 0 hashes

lumjjb:
# Description of the PR

When ingesting SBOMs, i noticed that the ingestion was slow (and was waiting on `IsOccurrences`). Experimented removing the hashes with all 0s in the digest field results in much faster times.

It looks like many of these files have this 0 hash and thus creates a node with very large number of edges out of it. 

This PR modifies the SPDX parser such that it ignores the hashes of all 0s (which Syft uses to represent symlinks), and empty file hashes.

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/spdx/parse_spdx.go
+++ b/pkg/ingestor/parser/spdx/parse_spdx.go
@@ -157,6 +157,9 @@ func (s *spdxParser) getFiles() error {
 
 		// if checksums exists create an artifact for each of them
 		for _, checksum := range file.Checksums {
+			if isEmptyChecksum(checksum.Value) {
+				continue
+			}
 			// for each file create a package for each of them so they can be referenced as a dependency
 			purl := asmhelpers.GuacFilePurl(strings.ToLower(string(checksum.Algorithm)), checksum.Value, &file.FileName)
 			pkg, err := asmhelpers.PurlToPkg(purl)
@@ -169,6 +172,7 @@ func (s *spdxParser) getFiles() error {
 				Algorithm: strings.ToLower(string(checksum.Algorithm)),
 				Digest:    checksum.Value,
 			}
+
 			s.fileArtifacts[string(file.FileSPDXIdentifier)] = append(s.fileArtifacts[string(file.FileSPDXIdentifier)], artifact)
 		}
 	}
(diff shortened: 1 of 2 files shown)


pxp928 (approved):

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -324,3 +328,15 @@ func getJustification(r *spdx.Relationship) string {
...
+
+func isEmptyChecksum(v string) bool {
+	return map[string]bool{
+		// all 0 hash
+		"0000000000000000000000000000000000000000":                         true,
+		"0000000000000000000000000000000000000000000000000000000000000000": true,
+		// sha1 empty file
+		"da39a3ee5e6b4b0d3255bfef95601890afd80709": true,
+		// sha256 empty file
+		"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855": true,
```
pxp928, line 340: qq: do we need to handle sha512 also?

lumjjb: yea and technically the whole list too :). so i just picked the common ones, but yea ill add sha512 for this one and then maybe the rest we can leave till when it comes up? or if something wants to help as a good first issue.

 [SHA1](https://tools.ietf.org/html/rfc3174), [SHA224](https://en.wikipedia.org/wiki/SHA-2), [SHA256](https://tools.ietf.org/html/rfc6234), [SHA384](https://en.wikipedia.org/wiki/SHA-2), [SHA512](https://en.wikipedia.org/wiki/SHA-2), [SHA3-256](https://en.wikipedia.org/wiki/SHA-3), [SHA3-384](https://en.wikipedia.org/wiki/SHA-3), [SHA3-512](https://en.wikipedia.org/wiki/SHA-3), [BLAKE2b-256](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2), [BLAKE2b-384](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2), [BLAKE2b-512](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE2), [BLAKE3](https://en.wikipedia.org/wiki/BLAKE_(hash_function)#BLAKE3), [MD2](https://tools.ietf.org/html/rfc1319), [MD4](https://tools.ietf.org/html/rfc1320), [MD5](https://tools.ietf.org/html/rfc1321), [MD6](https://groups.csail.mit.edu/cis/md6/), [ADLER32](https://en.wikipedia.org/wiki/Adler-32)

lumjjb: added that and created an issue for tracking other algs (https://github.com/guacsec/guac/issues/1229)

jeffmendoza (approved):

PR #1230, opened 2023-09-07
fix: increase gprc max message size

dejanb:
# Description of the PR

Another issue while trying to ingest large SBOM is that a collector-sub message can get larger than default 4MB. I made a quick fix to increase it to 16MB which double than what I saw in largest SBOMs so far. If you think we can also make it configurable?

# PR Checklist


pkg/collectsub/server/server.go
```diff
@@ -126,6 +127,7 @@ func (s *server) Serve(ctx context.Context) error {
 				grpc_zap.UnaryServerInterceptor(logger.Desugar()),
 				contextToZapFieldsUnaryServerInterceptor(),
 			)),
+		grpc.MaxRecvMsgSize(16777216),
```
pxp928, line 130: depending on how often we see this changing, we could make it an env variable for now with a default value.

pxp928 (approved):
LGTM!

jeffmendoza (approved):

PR #1231, opened 2023-09-07
Inmem mutation API fix to only return ID(string or  []string)

### LINKED ISSUE #1196: [feature - inmem] Change all backend mutation implementations to return only ID
Followup to issue https://github.com/guacsec/guac/issues/1116. Now that the graphQL mutation  API has been updated, the [backend](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/backends.go) needs to be updated to only return the IDs based on the revised [resolvers](https://github.com/guacsec/guac/tree/main/pkg/assembler/graphql/resolvers)

Multiple issues will be created for each backend update. This issue will focus on the updates to the [inmem](https://github.com/guacsec/guac/tree/main/pkg/assembler/backends/inmem) backend. 

All mutations must be changed to return only ID (string). 

For example for artifacts:

```
IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (string, error)
IngestArtifacts(ctx context.Context, artifacts []*model.ArtifactInputSpec) ([]string, error) 
```

Must both be updated in the [backend interface](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/backends.go#L54-L55) and in the [inmem backend](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/inmem/artifact.go#L102-L116)

Unit tests must be updated as needed to reflect these changes

arorasoham9:
# Description of the PR

Fixes #1196

# PR Checklist


pkg/assembler/backends/inmem/artifact.go
```diff
@@ -99,20 +99,24 @@ func (n *artStruct) setPointOfContactLinks(id uint32) {
...
+		modelArtifactsIDS = append(modelArtifactsIDS, modelArt.ID)
 	}
-	return modelArtifacts, nil
+	return modelArtifactsIDS, nil
 }
 
-func (c *demoClient) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (*model.Artifact, error) {
-	return c.ingestArtifact(ctx, artifact, true)
+func (c *demoClient) IngestArtifact(ctx context.Context, artifact *model.ArtifactInputSpec) (string, error) {
+	model, err := c.ingestArtifact(ctx, artifact, true)
```
pxp928, line 115: Thanks @arorasoham9 for the PR. This could be improved even more where it does not need to call `return c.convArtifact(a), nil` as its unnecessary and can just return the `nodeID(a.id)` (same for others in the inmem backend).

pkg/assembler/backends/inmem/certifyBad.go
```diff
@@ -57,38 +57,42 @@ func (n *badLink) BuildModelNode(c *demoClient) (model.Node, error) {
...
+		modelCertifyBadsIDS = append(modelCertifyBadsIDS, certifyBad.ID)
 	}
-	return modelCertifyBads, nil
+	return modelCertifyBadsIDS, nil
 }
 
-func (c *demoClient) IngestCertifyBad(ctx context.Context, subject model.PackageSourceOrArtifactInput, pkgMatchType *model.MatchFlags, certifyBad model.CertifyBadInputSpec) (*model.CertifyBad, error) {
-	return c.ingestCertifyBad(ctx, subject, pkgMatchType, certifyBad, true)
+func (c *demoClient) IngestCertifyBad(ctx context.Context, subject model.PackageSourceOrArtifactInput, pkgMatchType *model.MatchFlags, certifyBad model.CertifyBadInputSpec) (string, error) {
+	model, err := c.ingestCertifyBad(ctx, subject, pkgMatchType, certifyBad, true)
```
pxp928, line 91: same here we do not need to call `buildCertifyBad` as its a wasted operation. Can instead just call `nodeID(link.id)` and return that. (similar for others)

pxp928 (commented):
Thanks @arorasoham9 for working on this! Couple of improvements that can be made. Also the some of the unit tests might have to be updated. For example when `IngestArtifacts` is being tested, we can call `Artifacts(ctx, tt.artifactSpec)` to query for the values so we are not comparing just IDs.

lumjjb:
Looks like there's some merge editor artifacts still around, could you help resolve them? Thanks @arorasoham9 !

pkg/assembler/backends/backends.go
```diff
@@ -53,6 +53,7 @@ type Backend interface {
 	VulnerabilityMetadata(ctx context.Context, vulnerabilityMetadataSpec *model.VulnerabilityMetadataSpec) ([]*model.VulnerabilityMetadata, error)
 
 	// Mutations for software trees (read-write queries)
+<<<<<<< Updated upstream
```
mihaimaruseac, line 56: This looks like a merge conflict marker

pkg/assembler/backends/inmem/artifact.go
pxp928: based on the conversation on the issue, we can do this in other PRs.

pkg/assembler/backends/inmem/certifyBad.go
pxp928: based on the conversation on the issue, we can do this in other PRs.

pkg/assembler/backends/backends.go
arorasoham9: I'll take care of this. Thank you for pointing out!

PR #1232, opened 2023-09-07
Ent - Dependency management

mrizzi:
# Description of the PR

Provide implementation for endpoints related to `Dependency` management.

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/dependency.go
+++ b/pkg/assembler/backends/ent/dependency.go
@@ -20,8 +20,10 @@ type Dependency struct {
 	ID int `json:"id,omitempty"`
 	// PackageID holds the value of the "package_id" field.
 	PackageID int `json:"package_id,omitempty"`
-	// DependentPackageID holds the value of the "dependent_package_id" field.
-	DependentPackageID int `json:"dependent_package_id,omitempty"`
+	// DependentPackageNameID holds the value of the "dependent_package_name_id" field.
+	DependentPackageNameID int `json:"dependent_package_name_id,omitempty"`
+	// DependentPackageVersionID holds the value of the "dependent_package_version_id" field.
+	DependentPackageVersionID int `json:"dependent_package_version_id,omitempty"`
 	// VersionRange holds the value of the "version_range" field.
 	VersionRange string `json:"version_range,omitempty"`
 	// DependencyType holds the value of the "dependency_type" field.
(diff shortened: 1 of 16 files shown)


pkg/assembler/backends/ent/dependency.go
```diff
@@ -42,13 +44,15 @@ type Dependency struct {
...
-	// DependentPackage holds the value of the dependent_package edge.
-	DependentPackage *PackageName `json:"dependent_package,omitempty"`
+	// DependentPackageName holds the value of the dependent_package_name edge.
+	DependentPackageName *PackageName `json:"dependent_package_name,omitempty"`
+	// DependentPackageVersion holds the value of the dependent_package_version edge.
+	DependentPackageVersion *PackageVersion `json:"dependent_package_version,omitempty"`
 	// loadedTypes holds the information for reporting if a
 	// type was loaded (or requested) in eager-loading or not.
-	loadedTypes [2]bool
+	loadedTypes [3]bool
```
mihaimaruseac, line 53: Question for my understanding: What is the reason for the change from 2 to 3 here?

mihaimaruseac: I see line 87 below. Could we instead make these bools separate variables with clear names?

pkg/assembler/backends/ent/dependency.go
```diff
@@ -64,25 +68,38 @@ func (e DependencyEdges) PackageOrErr() (*PackageVersion, error) {
...
+		return e.DependentPackageName, nil
 	}
-	return nil, &NotLoadedError{edge: "dependent_package"}
+	return nil, &NotLoadedError{edge: "dependent_package_name"}
+}
+
+// DependentPackageVersionOrErr returns the DependentPackageVersion value or an error if the edge
+// was not loaded in eager-loading, or loaded but was not found.
+func (e DependencyEdges) DependentPackageVersionOrErr() (*PackageVersion, error) {
+	if e.loadedTypes[2] {
```
mihaimaruseac, line 87: This seems very likely to cause issues on future refactoring as it is unclear why the element at index 2 is the one to be used here

mihaimaruseac (changes requested):

pkg/assembler/backends/ent/dependency.go
mrizzi: This code is generated by Ent so I'm not so sure we have room for customization.
At the same time, the fact that they're generated from Ent could also mean we won't have to work on them directly so having unclear names isn't a big problem.

WDYT?

pkg/assembler/backends/ent/dependency.go
mrizzi: The same consideration as above applies here.

pkg/assembler/backends/ent/dependency.go
mihaimaruseac: Makes sense, I didn't realzie that. Resolving

mihaimaruseac (approved):

jeffmendoza (approved):

PR #1239, opened 2023-09-08
Feature/pkg equals bulk ingestion and arango updates

pxp928:
# Description of the PR

- bulk ingestion for pkg Equal
- update resolver checks and add missing unit tests
- update inmem backend with unit tests
- update arango backend with unit tests

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [x] All new changes are covered by tests
- [x] If GraphQL schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [x] All CI checks are passing (tests and formatting)
- [x] All dependent PRs have already been merged

PR #1243, opened 2023-09-08
 Ent - IngestPkgEquals with tests

mrizzi:
# Description of the PR

Follow up on https://github.com/guacsec/guac/pull/1239

The only difference is in the `Query on pkg algo and pkg` where I added `Subpath: ptrfrom.String("")` to the specs otherwise the query would have returned both the ingested `pkgEqual`, basically having Ent backend to consider the lack of a field in the query's spec meaning "don't filter by this field"

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/backend/pkgequal_test.go
+++ b/pkg/assembler/backends/ent/backend/pkgequal_test.go
@@ -544,6 +544,209 @@ func (s *Suite) TestPkgEqual() {
 	}
 }
 
+func (s *Suite) TestIngestPkgEquals() {
+	type call struct {
+		P1 []*model.PkgInputSpec
+		P2 []*model.PkgInputSpec
+		PE []*model.PkgEqualInputSpec
+	}
+	tests := []struct {
+		Name         string
+		InPkg        []*model.PkgInputSpec
+		Calls        []call
+		Query        *model.PkgEqualSpec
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/ent/backend/pkgequal_test.go
```diff
@@ -544,6 +544,209 @@ func (s *Suite) TestPkgEqual() {
...
+						{
+							Justification: "test justification",
+						},
+						{
+							Justification: "test justification",
+						},
+					},
+				},
+			},
+			Query: &model.PkgEqualSpec{
```
pxp928, line 663: hmm is this a bug on the `inmem` backend that it does not return all if the `subpath` is not set?

mrizzi: I think so based on consideration that lacking a field in the query's spec should mean "don't filter by this field"

mrizzi: Draft PR to reproduce the issue https://github.com/guacsec/guac/pull/1249

pxp928: yeah let's discuss this there. No need to block this PR.

jeffmendoza (approved):

PR #1244, opened 2023-09-08
Ingestor/Assembler and SPDX Parser for Legal information.

jeffmendoza:
# Description of the PR

Not sure if we should create "License" nodes for [SPDX exceptions](https://spdx.org/licenses/exceptions-index.html), or just leave them in the expression only.

Suggestions on default Justification ""Found in SPDX document." appreciated.

# PR Checklist

### DIFF
--- a/pkg/assembler/assembler.go
+++ b/pkg/assembler/assembler.go
@@ -47,6 +47,7 @@ type IngestPredicates struct {
 	PointOfContact   []PointOfContactIngest   `json:"contact,omitempty"`
 	VulnMetadata     []VulnMetadataIngest     `json:"vulnMetadata,omitempty"`
 	HasMetadata      []HasMetadataIngest      `json:"hasMetadata,omitempty"`
+	CertifyLegal     []CertifyLegalIngest     `json:"certifyLegal,omitempty"`
 }
 
 type CertifyScorecardIngest struct {
@@ -184,6 +185,16 @@ type PkgEqualIngest struct {
 	PkgEqual *generated.PkgEqualInputSpec `json:"pkgEqual,omitempty"`
 }
 
+type CertifyLegalIngest struct {
+	Pkg *generated.PkgInputSpec    `json:"pkg,omitempty"`
+	Src *generated.SourceInputSpec `json:"src,omitempty"`
+
+	Declared   []generated.LicenseInputSpec `json:"declared,omitempty"`
+	Discovered []generated.LicenseInputSpec `json:"discovered,omitempty"`
+
+	CertifyLegal *generated.CertifyLegalInputSpec `json:"certifyLegal,omitempty"`
+}
+
 func (i IngestPredicates) GetPackages(ctx context.Context) []*generated.PkgInputSpec {
 	packageMap := make(map[string]*generated.PkgInputSpec)
 	for _, dep := range i.IsDependency {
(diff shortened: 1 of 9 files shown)


pkg/assembler/assembler.go
```diff
@@ -493,6 +520,29 @@ func (i IngestPredicates) GetVulnerabilities(ctx context.Context) []*generated.V
 	return vulns
 }
 
+func (i IngestPredicates) GetLicenses(ctx context.Context) []generated.LicenseInputSpec {
```
pxp928, line 523: can you add a quick unit test for this similar to the others

pkg/assembler/clients/helpers/bulk.go
```diff
@@ -615,3 +634,48 @@ func ingestIsOccurrences(ctx context.Context, client graphql.Client, v []assembl
...
+	var sources []model.SourceInputSpec
+	var pkgDec [][]model.LicenseInputSpec
+	var pkgDis [][]model.LicenseInputSpec
+	var pkgCL []model.CertifyLegalInputSpec
+	var srcDec [][]model.LicenseInputSpec
+	var srcDis [][]model.LicenseInputSpec
+	var srcCL []model.CertifyLegalInputSpec
+	for _, ingest := range v {
+
+		if ingest.Pkg != nil && ingest.Src != nil {
```
pxp928, line 649: we can add these checks to the resolver but you already have that as part of your issue TODOs so all good.

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -65,6 +70,14 @@ func (s *spdxParser) Parse(ctx context.Context, doc *processor.Document) error {
 		return fmt.Errorf("failed to parse SPDX document: %w", err)
 	}
 	s.spdxDoc = spdxDoc
+	if spdxDoc.CreationInfo == nil {
```
pxp928, line 73: qq: is this a required field by SPDX?

pxp928 (approved):
LGTM. minor comments and a question

pkg/assembler/assembler.go
jeffmendoza: Will do, missed those

pkg/ingestor/parser/spdx/parse_spdx.go
jeffmendoza: yes, I was hesitant to add, but looks required: https://spdx.github.io/spdx-spec/v2-draft/document-creation-information/#69-created-field

pkg/assembler/assembler.go
jeffmendoza: done

pkg/ingestor/parser/common/license.go
```diff
@@ -0,0 +1,95 @@
...
+	"golang.org/x/exp/slices"
+)
+
+var ignore = []string{
+	"AND",
+	"OR",
+	"WITH",
+}
+
+// Could add exceptions to ignore list, as they are not licenses:
```
mihaimaruseac, line 33: Is there a way to query OSI approved licenses and only use those? Do we want to do that?

mihaimaruseac (approved):
LGTM, just one question

pkg/ingestor/parser/common/license.go
jeffmendoza: The plan is to stick to the SPDX license list, which both SPDX and CDX use: https://spdx.org/licenses/
It includes both OSI approved and some non OSI approved licenses, reasoning here: https://github.com/spdx/license-list-XML/blob/main/DOCS/license-inclusion-principles.md

jeffmendoza: A license expression can include both `license-id` which points to an actual license, or also `license-exception` (https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/#d44-exception-with-operator). I copied the exception list here, as we maybe don't want to create GUAC Licnese nodes for exceptions, as they are not licenses. But maybe GUAC users want to easily query for packages with a particular exception, so that would be a reason to not ignore them.

PR #1249, opened 2023-09-11
inmem - PkgEqual: Subpath query filtering

mrizzi:
# Description of the PR

Draft PR to further discuss about https://github.com/guacsec/guac/pull/1243#discussion_r1320122295:

Added both `PkgEqual` instance to the expected outcomes for `Query on pkg algo and pkg` test based on the assumption that not having provided a `Subpath` value in the query `PkgEqualSpec` should not filter by `Subpath`.

The "reasonable" way I found to have it working has been adding the `filter.Subpath == nil` in the `exactPackageVersion` func rather than changing the `noMatchInput` because changing the latter made much more tests to fail even if this brings me the doubt that *maybe* the filtering could be broken somewhere else.

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/inmem/pkg.go
+++ b/pkg/assembler/backends/inmem/pkg.go
@@ -743,21 +743,6 @@ func noMatchQualifiers(filter *model.PkgSpec, v map[string]string) bool {
 	return false
 }
 
-func (c *demoClient) exactPackageVersion(filter *model.PkgSpec) (*pkgVersionNode, error) {
-	pkgs, err := c.findPackageVersion(filter)
-	if err != nil {
-		return nil, gqlerror.Errorf("No package version matches spec %v", err)
-	}
-	switch len(pkgs) {
-	case 0:
-		return nil, nil
-	case 1:
-		return pkgs[0], nil
-	default:
-		return nil, gqlerror.Errorf("More than one package version matches spec")
-	}
-}
-
 func (c *demoClient) findPackageVersion(filter *model.PkgSpec) ([]*pkgVersionNode, error) {
 	if filter == nil {
 		return nil, nil
@@ -790,7 +775,6 @@ func (c *demoClient) findPackageVersion(filter *model.PkgSpec) ([]*pkgVersionNod
 		}
 		for _, v := range nm.versions {
 			if *filter.Version != v.version ||
-				//filter.Subpath == nil ||
 				noMatch(filter.Subpath, v.subpath) ||
 				noMatchQualifiers(filter, v.qualifiers) {
 				continue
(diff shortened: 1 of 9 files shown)


mrizzi:
@pxp928 @mihaimaruseac @jeffmendoza what do you think about this issue?

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -774,6 +774,7 @@ func (c *demoClient) exactPackageVersion(filter *model.PkgSpec) (*pkgVersionNode
 		}
 		for _, v := range nm.versions {
 			if *filter.Version != v.version ||
+				filter.Subpath == nil ||
 				noMatchInput(filter.Subpath, v.subpath) ||
```
pxp928, line 778: first this should be `noMatch` and not `noMatchInput`. Also, there is a logic issue here. `nm.versions` does contain the version that contains the subpath but that is never evaluated and it only returns before it can be evaluated.

pxp928: Seems like `exactPackageVersion` should return a list of matching `pkgVersionNode` that can be evaluated

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -744,6 +744,21 @@ func noMatchQualifiers(filter *model.PkgSpec, v map[string]string) bool {
...
+	if err != nil {
+		return nil, gqlerror.Errorf("No package version matches spec %v", err)
+	}
+	switch len(pkgs) {
+	case 0:
+		return nil, nil
+	case 1:
+		return pkgs[0], nil
+	default:
+		return nil, gqlerror.Errorf("More than one package version matches spec")
```
pxp928, line 757: Thanks for doing this @mrizzi, so this would be a valid return. In that case, it should return all found and let the verb filters figure out which ones are valid per the query.

pkg/assembler/backends/inmem/pkg.go
mrizzi: > Seems like `exactPackageVersion` should return a list of matching `pkgVersionNode` that can be evaluated as it currently returns on the first matching, when in reality it should return all matching and let the downstream sort out the ones that are matching.

I've introduced `findPackageVersion` func to return the list of matching `pkgVersionNode` and refactored the `exactPackageVersion` func for consistency with its usages done in other components (and managing the `exact` constraint accordingly)

pkg/assembler/backends/inmem/pkg.go
pxp928: hmm I think it should always be `findPackageVersion` as the `exactPackageVersion` is not valid. Meaning we need to search all returned matching pkgVersions backlinks for matching verbs.

pxp928: else we would have a similar issue in other verbs as we did with `pkgEquals`

mrizzi: > hmm I think it should always be `findPackageVersion` as the `exactPackageVersion` is not valid. Meaning we need to search all returned matching pkgVersions backlinks for matching verbs.

Done :wink:

jeffmendoza (approved):
Awesome. thanks for digging into this.

PR #1279, opened 2023-09-15
Add OCI registry collector

### LINKED ISSUE #298: Registry collector to expand OCI collector
Implementation of a registry collector, where the collector would query for all repositories and pull all metadata stored for each, was put on hold as `_catalog` API is not implemented the same across all registries. This would result in fixes and bugs depending on which registry (cloud or private) the user utilized. Each cloud-hosted registry adds its own custom API for listing things within a single org. We can revisit this later if needed but oci collector is currently setup such that it can be called by an upstream "registry collector" to collect from all repos in a registry in the future if needed. This can be done by the following using regclient API:
```
   repoTags := map[string][]string{}
   r, err := ref.New(o.registry)
   if err != nil {
     return fmt.Errorf("failed to parse ref %s: %v", r, err)
   }
   defer rc.Close(ctx, r)

   rl, err := rc.RepoList(ctx, o.registry)
   if err != nil && errors.Is(err, types.ErrNotImplemented) {
     return fmt.Errorf("registry %s does not support underlying _catalog API: %w", o.registry, err)
   }

   for _, repo := range rl.Repositories {
     r, err := ref.New(repo)
     if err != nil {
       return err
     }
     collectedTags, err := rc.TagList(ctx, r)
     if err != nil {
       return err
     }
     repoTags[repo] = collectedTags
   }

   ociRepoCollector := NewOCIRepoCollector(ctx, repoTags, false, time.Second)
   ociRepoCollector.RetrieveArtifacts(ctx, docChannel)
   o.checkedDigest[repo] = ociRepoCollector.checkedDigest
   return nil
```

ridhoq:
# Description of the PR

Support collecting all OCI artifacts in an OCI registry that has the `/v2/_catalog` endpoint enabled.

TODO:
- [ ] Add unit tests
- [ ] Should we be duplicating the logic in both `guacone` and `guaccollect`?
- [ ] Should we be doing the collection in parallel? Huge registries will take a long time?
- [ ] How can we ensure that we don't collect something that hasn't been collected before? IE, if I run a collection twice against a registry, can we ensure that we don't collect everything all over again?

Fixes #298 

# PR Checklist

### DIFF
--- a/pkg/handler/collector/oci/oci_registry.go
+++ b/pkg/handler/collector/oci/oci_registry.go
@@ -1,0 +1,99 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package oci
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/guacsec/guac/pkg/collectsub/datasource"
+	"github.com/guacsec/guac/pkg/collectsub/datasource/inmemsource"
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/pkg/errors"
+	"github.com/regclient/regclient"
+	"github.com/regclient/regclient/types"
+	"github.com/regclient/regclient/types/ref"
+)
+
+const (
+	OCIRegistryCollector = "OCIRegistryCollector"
+)
+
+type ociRegistryCollector struct {
+	collectDataSource datasource.CollectSource
+	checkedDigest     map[string][]string
+	registry          string
+	poll              bool
+	interval          time.Duration
+}
+
+func NewOCIRegistryCollector(ctx context.Context, registry string, poll bool, interval time.Duration) *ociRegistryCollector {
+	return &ociRegistryCollector{
+		checkedDigest: map[string][]string{},
+		registry:      registry,
+		poll:          poll,
+		interval:      interval,
+	}
+}
+
+func (o *ociRegistryCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	r, err := ref.New(o.registry)
+	if err != nil {
+		return fmt.Errorf("failed to parse ref %s: %v", r, err)
+	}
+
+	rcOpts := getRegClientOptions()
+	rc := regclient.New(rcOpts...)
+	defer rc.Close(ctx, r)
+
+	rl, err := rc.RepoList(ctx, o.registry)
+	if err != nil && errors.Is(err, types.ErrNotImplemented) {
+		return fmt.Errorf("registry %s does not support underlying _catalog API: %w", o.registry, err)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to list repositories in registry %s: %w", o.registry, err)
+	}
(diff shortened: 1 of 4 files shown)


pkg/handler/collector/oci/oci_registry.go
```diff
@@ -0,0 +1,99 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("unable to create datasource: %w", err)
+	}
+	ociCollector := NewOCICollector(ctx, o.collectDataSource, o.poll, o.interval)
+	err = ociCollector.RetrieveArtifacts(ctx, docChannel)
+	if err != nil {
+		return fmt.Errorf("unable to retrieve artifacts from OCI collector: %w", err)
+	}
+	o.checkedDigest = ociCollector.checkedDigest
```
pxp928, line 92: hmm the oci collector could skip all those that are duplicates. So no checks have to be done at the OCI registry collector.

pkg/handler/collector/oci/oci_registry.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (o *ociRegistryCollector) RetrieveArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	r, err := ref.New(o.registry)
+	if err != nil {
+		return fmt.Errorf("failed to parse ref %s: %v", r, err)
+	}
+
+	rcOpts := getRegClientOptions()
+	rc := regclient.New(rcOpts...)
+	defer rc.Close(ctx, r)
```
naveensrinivasan, line 61: The error returned from `ref.New` is not checked before passing r to `rc.Close`. 

This could lead to a panic if ref.New returns an error.

pkg/handler/collector/oci/oci_registry.go
```diff
@@ -0,0 +1,99 @@
...
+
+	rl, err := rc.RepoList(ctx, o.registry)
+	if err != nil && errors.Is(err, types.ErrNotImplemented) {
+		return fmt.Errorf("registry %s does not support underlying _catalog API: %w", o.registry, err)
+	}
+	if err != nil {
+		return fmt.Errorf("failed to list repositories in registry %s: %w", o.registry, err)
+	}
+	if len(rl.Repositories) == 0 {
```
naveensrinivasan, line 70: `rl.Repositories` is not checked for nil. This could cause a panic.

Suggestion: Add nil check on rl.Repositories.

pkg/handler/collector/oci/oci_registry.go
naveensrinivasan, line 61: Close can return an error. This would not be noticed if someone is trying to debug.

A recommendation would be to log this.

```golang
defer func(rc *regclient.RegClient, ctx context.Context, r ref.Ref) {
		err := rc.Close(ctx, r)
		if err != nil {
			//log error
		}
	}(rc, ctx, r)
```

naveensrinivasan (commented):
A few suggestions 👍 . Thanks

pxp928:
@ridhoq did you have any cycles to complete this PR? Would be a great feature to add in.

ridhoq:
> @ridhoq did you have any cycles to complete this PR? Would be a great feature to add in.

@pxp928 Agreed, thanks for the ping - let's try to get this one merged. If anything, I will spin up follow up issues for this one to reduce the scope of this PR to get it merged.

ridhoq:
Pinging this one so it doesn't close

ridhoq:
@robert-cronin will be taking on this work in #2185 so we can close this one out.

pkg/handler/collector/oci/oci_registry.go
robert-cronin: This has been handled in the new PR

pkg/handler/collector/oci/oci_registry.go
robert-cronin: This has been handled in the new PR

pkg/handler/collector/oci/oci_registry.go
robert-cronin: This has been handled in the new PR

PR #1292, opened 2023-09-18
Improve build process with dirt check

naveensrinivasan:
# Description of the PR

- Update Makefile to include `dirty` target
- The reason for this check is that if generate produces a different file from what is the source tree, this will catch it.

# PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ LOCAL_IMAGE_NAME ?= local-organic-guac
 .DEFAULT_GOAL := build
 
 .PHONY: all
-all: test cover fmt lint build generate dirty
+all: test cover fmt lint build generate
 
 # Run the unit tests
 .PHONY: test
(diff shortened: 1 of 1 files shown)


lumjjb:
if that is the case i think this check should be part of generate check, if not i'm not sure it would guarantee the ordering of generate then dirty.

naveensrinivasan:
> if that is the case i think this check should be part of generate check, if not i'm not sure it would guarantee the ordering of generate then dirty.

It is usually done before the build. I can move it to that.

jeffmendoza:
`make generated_up_to-date` does a similar thing
https://github.com/guacsec/guac/blob/main/Makefile#L50-L53
It is used a few times in workflows, ex: https://github.com/guacsec/guac/blob/main/.github/workflows/ci.yaml#L113-L114

Looks like that one will fail on any extra files, while this diff one will only fail on changes to existing files. I'm not sure we need both, but also don't know which way to go.

lumjjb:
yea i have a dirty branch almost always, so i think only for the CI case are my thoughts. In that case +1 to @jeffmendoza 's comment where i think we already have that check.

naveensrinivasan:
> `make generated_up_to-date` does a similar thing https://github.com/guacsec/guac/blob/main/Makefile#L50-L53 It is used a few times in workflows, ex: https://github.com/guacsec/guac/blob/main/.github/workflows/ci.yaml#L113-L114
> 
> Looks like that one will fail on any extra files, while this diff one will only fail on changes to existing files. I'm not sure we need both, but also don't know which way to go.

I'll close this.

PR #1299, opened 2023-09-20
Ent - IngestHasSourceAts implementation

mrizzi:
# Description of the PR

Ent backend:

- `IngestHasSourceAts` implementation with tests (as usual tests reflect the inmem ones)
- fix typo introduced in https://github.com/guacsec/guac/pull/1295

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/backend/source.go
+++ b/pkg/assembler/backends/ent/backend/source.go
@@ -27,6 +27,7 @@ import (
 	"github.com/guacsec/guac/pkg/assembler/backends/ent/sourcenamespace"
 	"github.com/guacsec/guac/pkg/assembler/backends/ent/sourcetype"
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
+	"github.com/vektah/gqlparser/v2/gqlerror"
 )
 
 func (b *EntBackend) HasSourceAt(ctx context.Context, filter *model.HasSourceAtSpec) ([]*model.HasSourceAt, error) {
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/ent/backend/source.go
```diff
@@ -272,6 +285,8 @@ func toModelHasSourceAt(record *ent.HasSourceAt) *model.HasSourceAt {
 		pkg = toModelPackage(backReferencePackageVersion(record.Edges.PackageVersion))
 	} else {
 		pkg = toModelPackage(backReferencePackageName(record.Edges.AllVersions))
+		// in this case, the expected response is package name with an empty package version array
```
pxp928, line 288: If I am understanding this right: if it is set to AllVersions, the edge should be to the package name - `pkg.Namespaces[0].Names[0]`. We might have miscommunicated when you asked me before.

pxp928: Meaning that all versions under the `pkgName`

mrizzi: That's a good point, let me try to explain.

I had to add `pkg.Namespaces[0].Names[0].Versions = []*model.PackageVersion{}` in order to have the `TestIngestHasSourceAts/HappyPath_All_Versions` test to work because it's expecting `p1outName` which has `Versions: []*model.PackageVersion{},` but, without this line, the Ent output is `Versions: nil,`

So it's a match type `ALL_VERSIONS` that it's expected to return an empty `model.PackageVersion` array in that test.
If this expectation is right, then I feel like I have to set that empty array, otherwise the test's expected result has to be changed.

Does this make sense? Am i missing something?

pxp928: ah ok yes that makes sense.

pxp928 (approved):

PR #1302, opened 2023-09-20
update docker manifest name in nightly releases

mdeicas:
# Description of the PR

The docker manifest in the nightly releases are named [`nightly`](https://github.com/guacsec/guac/pkgs/container/guac/129561709?tag=nightly) instead of `v0-nightly` which is the release tag. This is inconsistent with `.goreleaser.yaml ` and with the expected manifest name of `v0-nightly` in the `.env` file in the nightly release compose tarball.

### DIFF
--- a/.goreleaser-nightly.yaml
+++ b/.goreleaser-nightly.yaml
@@ -41,7 +41,7 @@ dockers:
       - "--builder={{ .Env.DOCKER_CONTEXT }}"
 
 docker_manifests:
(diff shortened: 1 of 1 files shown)


lumjjb (approved):
Lgtm! Good catch!

sunnyyip:
Let's also remove this comment if we were to align the tag between bins and container image - https://github.com/guacsec/guac/blob/main/.goreleaser-nightly.yaml#L217.

pxp928 (approved):

Issue #1020, opened 2023-07-05
[feature] Add S3 Collector


dejanb:
This is a holder issues for implementing/integrating S3 collector into Guac.

**Describe the solution you'd like**

There's a Rust implementation of Collector available [here](https://github.com/trustification/trustification/tree/main/exporter), which works in combination with SBOM/VEX archival services.

It can also be used directly from [guac-rs library](https://github.com/dejanb/guac-rs#s3).

I'm still not sure what would be the best way for people to use this. I can create docker image with just this S3 exporter. We can use existing trustification images and services. And so on. 

It'd be great if more people would give it a try and provide a feedback how they would this service without or in combination with trustification services.

lumjjb:
awesome yea i think in the near term, we can create like a community-projects MD or something similar and point to those repos. And I think it should be good to have docs that point to other community stuff as well. Thoughts @guacsec/guac-maint

mihaimaruseac:
+1 on having a repository for community-projects. We should also have a process for graduating a community project from the community repo to its standalone repo

lumjjb:
Implemented in #1308

Issue #1177, opened 2023-08-23
[feature] Ingestor document upload/ingest API


lumjjb:
This is an attempt to provide a solution to #731 . And alternatives to this solution are described more within the design document.

Inspired by @dejanb during our community call, the proposal is as follows:

# Ingestor API for ingesting blobs

The following is a proposed solution for the ingestor to expose a network accessible API (e.g. gRPC) to allow the streaming of a large blob. 

This document metadata can be part of the proto definition (like in this file upload example), with the following golang reference for chunk upload.

```
message DocumentMetaData {
  string collectorInformation = 1;
  string sourceInformation = 2;
}

message Document {
  bytes content = 1;
}

enum Status {
  PENDING = 0;
  IN_PROGRESS = 1;
  SUCCESS = 2;
  FAILED = 3;
}

message DocumentIngestRequest {
  oneof request {
    DocumentMetaData metadata = 1;
    Document doc = 2;
  }
}

message DocumentIngestResponse {
  Status status = 1;
}

service DocumentIngestService {
  rpc ingestDocument(stream DocumentIngestRequest) returns(DocumentIngestResponse);
}
```

Nikhil-cp1905:
I would like to take up this issue

PR #1308, opened 2023-09-22
S3 collector implementation

jmle:
# Description of the PR

This PR creates a new `guacone collect` command which connects to a source of S3 "upload" events (sqs of kafka), downloads a document from a bucket whenever a message is received, and ingests it into guac.

- docker-compose files for leveraging s3/sqs/kafka infrastructure can be found [here](https://github.com/jmle/trustification-compose)
- Related to [this issue](https://github.com/guacsec/guac/issues/1020).

# PR Checklist

### DIFF
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -35,15 +35,15 @@ import (
 
 // s3Options flags for configuring the command
 type s3Options struct {
-	s3url      string // base url of the s3 to collect from
-	s3bucket   string // name of bucket to collect from
-	s3item     string // s3 item (only for non-polling behaviour)
-	region     string // AWS region, for s3/sqs configuration (defaults to us-east-1)
-	queues     string // comma-separated list of queues/topics (only for polling behaviour)
-	mp         string // message provider name (sqs or kafka, will default to kafka)
-	mpEndpoint string // endpoint for the message provider (only for polling behaviour)
-	poll       bool   // polling or non-polling behaviour? (defaults to non-polling)
-	graphqlEndpoint   string // endpoint for the graphql server
+	s3url             string                        // base url of the s3 to collect from
+	s3bucket          string                        // name of bucket to collect from
+	s3item            string                        // s3 item (only for non-polling behaviour)
+	region            string                        // AWS region, for s3/sqs configuration (defaults to us-east-1)
+	queues            string                        // comma-separated list of queues/topics (only for polling behaviour)
+	mp                string                        // message provider name (sqs or kafka, will default to kafka)
+	mpEndpoint        string                        // endpoint for the message provider (only for polling behaviour)
+	poll              bool                          // polling or non-polling behaviour? (defaults to non-polling)
+	graphqlEndpoint   string                        // endpoint for the graphql server
 	csubClientOptions csub_client.CsubClientOptions // options for the collectsub client
 }
 
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -177,7 +177,7 @@ func validateS3Opts(args []string, s3bucket string, s3item string, mp string, mp
 }
 
 func init() {
-	set, err := cli.BuildFlags([]string{"s3-bucket", "s3-item", "s3-mp", "s3-mp-host", "s3-mp-port", "s3-queues", "s3-region"})
+	set, err := cli.BuildFlags([]string{"s3-bucket", "s3-item", "s3-mp", "s3-mp-host", "s3-mp-port", "s3-queues", "s3-region", "poll"})
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to setup flag: %s", err)
 		os.Exit(1)
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -86,7 +86,6 @@ var s3Cmd = &cobra.Command{
 			MessageProvider:         s3Opts.mp,
 			MessageProviderEndpoint: s3Opts.mpEndpoint,
 			Queues:                  s3Opts.queues,
-			SigChan:                 signals,
 			Poll:                    s3Opts.poll,
 		})
 
@@ -124,7 +123,15 @@ var s3Cmd = &cobra.Command{
 			return false
 		}
 
-		if err := collector.Collect(ctx, emit, errHandler); err != nil {
+		cancelCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
+		// Send cancellation in case of receiving SIGINT/SIGTERM
+		go func(cancel context.CancelFunc) {
+			cancel()
+		}(cancel)
+
+		if err := collector.Collect(cancelCtx, emit, errHandler); err != nil {
 			logger.Fatal(err)
 		}
 
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -19,7 +19,10 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/signal"
+	"syscall"
 
+	"github.com/guacsec/guac/pkg/cli"
 	csub_client "github.com/guacsec/guac/pkg/collectsub/client"
 	"github.com/guacsec/guac/pkg/handler/collector"
 	"github.com/guacsec/guac/pkg/handler/collector/s3"
@@ -28,7 +31,6 @@ import (
 	"github.com/guacsec/guac/pkg/logging"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"golang.org/x/sync/errgroup"
 )
 
 // s3Options flags for configuring the command
@@ -52,12 +54,26 @@ var s3Cmd = &cobra.Command{
 		ctx := logging.WithLogger(context.Background())
 		logger := logging.FromContext(ctx)
 
-		s3Opts, err := validateS3Opts(s3Opts.s3hostname, s3Opts.s3port, s3Opts.mp, s3Opts.mpHostname, s3Opts.mpPort, s3Opts.queues, s3Opts.region)
+		//s3Opts, err := validateS3Opts(s3Opts.s3hostname, s3Opts.s3port, s3Opts.mp, s3Opts.mpHostname, s3Opts.mpPort, s3Opts.queues, s3Opts.region)
+		s3Opts, err := validateS3Opts(
+			viper.GetString("s3-host"),
+			viper.GetString("s3-port"),
+			viper.GetString("s3-mp"),
+			viper.GetString("s3-mp-host"),
+			viper.GetString("s3-mp-port"),
+			viper.GetString("s3-queues"),
+			viper.GetString("s3-region"),
+		)
 		if err != nil {
+			logger.Errorf("failed to validate flags: %v", err)
+			_ = cmd.Help()
 			os.Exit(1)
 		}
 
-		s3Collector, err := s3.NewS3Collector(s3.S3CollectorConfig{
+		signals := make(chan os.Signal, 1)
+		signal.Notify(signals, syscall.SIGINT)
+
+		s3Collector := s3.NewS3Collector(s3.S3CollectorConfig{
 			S3Host:              s3Opts.s3hostname,
 			S3Port:              s3Opts.s3port,
 			MessageProvider:     s3Opts.mp,
@@ -65,14 +81,10 @@ var s3Cmd = &cobra.Command{
 			MessageProviderPort: s3Opts.mpPort,
 			Queues:              s3Opts.queues,
 			Region:              s3Opts.region,
+			SigChan:             signals,
 		})
-		if err != nil {
-			logger.Errorf("unable to create s3 collector: %v", err)
-			os.Exit(1)
-		}
 
-		err = collector.RegisterDocumentCollector(s3Collector, s3.S3CollectorType)
-		if err != nil {
+		if err := collector.RegisterDocumentCollector(s3Collector, s3.S3CollectorType); err != nil {
 			logger.Errorf("unable to register s3 collector: %v\n", err)
 			os.Exit(1)
 		}
@@ -87,9 +99,8 @@ var s3Cmd = &cobra.Command{
 
 		errFound := false
 
-		_, s3ctx := errgroup.WithContext(ctx)
 		emit := func(d *processor.Document) error {
-			err := ingestor.Ingest(s3ctx, d, viper.GetString("gql-addr"), csubClient)
+			err := ingestor.Ingest(ctx, d, viper.GetString("gql-addr"), csubClient)
 
 			if err != nil {
 				errFound = true
@@ -143,12 +154,16 @@ func validateS3Opts(s3hostname string, s3port string, mp string, mpHostname stri
 }
 
 func init() {
-	s3Cmd.Flags().StringVarP(&s3Opts.s3hostname, "s3-host", "", "", "hostname for s3")
-	s3Cmd.Flags().StringVarP(&s3Opts.s3port, "s3-port", "", "", "port for s3")
-	s3Cmd.Flags().StringVarP(&s3Opts.mp, "mp", "m", "", "message provider (sqs or kafka, defaults to kafka)")
-	s3Cmd.Flags().StringVarP(&s3Opts.mpHostname, "mp-host", "", "", "hostname for the message provider (sqs/kafka)")
-	s3Cmd.Flags().StringVarP(&s3Opts.mpPort, "mp-port", "", "", "port for the message provider (sqs/kafka)")
-	s3Cmd.Flags().StringVarP(&s3Opts.queues, "queues", "", "", "comma-separated list of queue/topic names")
-	s3Cmd.Flags().StringVarP(&s3Opts.region, "region", "", "us-east-1", "aws region, defaults to us-east-1")
+	set, err := cli.BuildFlags([]string{"s3-host", "s3-port", "s3-mp", "s3-mp-host", "s3-mp-port", "s3-queues", "s3-region"})
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "failed to setup flag: %s", err)
+		os.Exit(1)
+	}
+	s3Cmd.Flags().AddFlagSet(set)
+	if err := viper.BindPFlags(s3Cmd.Flags()); err != nil {
+		fmt.Fprintf(os.Stderr, "failed to bind flags: %s", err)
+		os.Exit(1)
+	}
+
 	collectCmd.AddCommand(s3Cmd)
 }
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -43,6 +43,8 @@ type s3Options struct {
 	mp         string // message provider name (sqs or kafka, will default to kafka)
 	mpEndpoint string // endpoint for the message provider (only for polling behaviour)
 	poll       bool   // polling or non-polling behaviour? (defaults to non-polling)
+	graphqlEndpoint   string // endpoint for the graphql server
+	csubClientOptions csub_client.CsubClientOptions // options for the collectsub client
 }
 
 var s3Cmd = &cobra.Command{
@@ -54,6 +56,10 @@ var s3Cmd = &cobra.Command{
 		logger := logging.FromContext(ctx)
 
 		s3Opts, err := validateS3Opts(
+			viper.GetString("gql-addr"),
+			viper.GetString("csub-addr"),
+			viper.GetBool("csub-tls"),
+			viper.GetBool("csub-tls-skip-verify"),
 			viper.GetString("s3-url"),
 			viper.GetString("s3-bucket"),
 			viper.GetString("s3-region"),
@@ -89,7 +95,7 @@ var s3Cmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		csubClient, err := csub_client.NewClient(viper.GetString("csub-addr"))
+		csubClient, err := csub_client.NewClient(s3Opts.csubClientOptions)
 		if err != nil {
 			logger.Infof("collectsub client initialization failed, this ingestion will not pull in any additional data through the collectsub service: %v", err)
 			csubClient = nil
@@ -100,7 +106,7 @@ var s3Cmd = &cobra.Command{
 		errFound := false
 
 		emit := func(d *processor.Document) error {
-			err := ingestor.Ingest(ctx, d, viper.GetString("gql-addr"), csubClient)
+			err := ingestor.Ingest(ctx, d, s3Opts.graphqlEndpoint, csubClient)
 
 			if err != nil {
 				errFound = true
@@ -130,7 +136,7 @@ var s3Cmd = &cobra.Command{
 	},
 }
 
-func validateS3Opts(s3url string, s3bucket string, region string, s3item string, mp string, mpEndpoint string, queues string, poll bool) (s3Options, error) {
+func validateS3Opts(graphqlEndpoint string, csubAddr string, csubTls bool, csubTlsSkipVerify bool, s3url string, s3bucket string, region string, s3item string, mp string, mpEndpoint string, queues string, poll bool) (s3Options, error) {
 	var opts s3Options
 
 	if poll {
@@ -152,7 +158,12 @@ func validateS3Opts(s3url string, s3bucket string, region string, s3item string,
 		return opts, fmt.Errorf("expected s3 bucket")
 	}
 
-	opts = s3Options{s3url, s3bucket, region, s3item, queues, mp, mpEndpoint, poll}
+	csubClientOptions, err := csub_client.ValidateCsubClientFlags(csubAddr, csubTls, csubTlsSkipVerify)
+	if err != nil {
+		return opts, fmt.Errorf("unable to validate csub client flags: %w", err)
+	}
+
+	opts = s3Options{s3url, s3bucket, region, s3item, queues, mp, mpEndpoint, poll, graphqlEndpoint, csubClientOptions}
 
 	return opts, nil
 }
--- a/cmd/guacone/cmd/s3.go
+++ b/cmd/guacone/cmd/s3.go
@@ -1,0 +1,154 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	csub_client "github.com/guacsec/guac/pkg/collectsub/client"
+	"github.com/guacsec/guac/pkg/handler/collector"
+	"github.com/guacsec/guac/pkg/handler/collector/s3"
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/guacsec/guac/pkg/ingestor"
+	"github.com/guacsec/guac/pkg/logging"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"golang.org/x/sync/errgroup"
+)
+
+// s3Options flags for configuring the command
+type s3Options struct {
+	s3hostname string // hostname of the s3 provider
+	s3port     string // port of the s3 provider
+	region     string // AWS region, for s3/sqs configuration (defaults to us-east-1)
+	queues     string // comma-separated list of queues/topics
+	mp         string // message provider name (sqs or kafka, will default to kafka)
+	mpHostname string // hostname for the message provider
+	mpPort     string // port for the message provider
+}
+
+var s3Opts s3Options
+
+var s3Cmd = &cobra.Command{
+	Use:   "s3 [flags]",
+	Short: "listens to kafka/sqs s3 events to download documents and add them to the GUAC graph",
+	Args:  cobra.MinimumNArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := logging.WithLogger(context.Background())
+		logger := logging.FromContext(ctx)
+
+		s3Opts, err := validateS3Opts(s3Opts.s3hostname, s3Opts.s3port, s3Opts.mp, s3Opts.mpHostname, s3Opts.mpPort, s3Opts.queues, s3Opts.region)
+		if err != nil {
+			os.Exit(1)
+		}
+
+		s3Collector, err := s3.NewS3Collector(s3.S3CollectorConfig{
+			S3Host:              s3Opts.s3hostname,
+			S3Port:              s3Opts.s3port,
+			MessageProvider:     s3Opts.mp,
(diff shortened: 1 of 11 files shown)


cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func init() {
+	s3Cmd.Flags().StringVarP(&s3Opts.s3hostname, "s3-host", "", "", "hostname for s3")
+	s3Cmd.Flags().StringVarP(&s3Opts.s3port, "s3-port", "", "", "port for s3")
+	s3Cmd.Flags().StringVarP(&s3Opts.mp, "mp", "m", "", "message provider (sqs or kafka, defaults to kafka)")
+	s3Cmd.Flags().StringVarP(&s3Opts.mpHostname, "mp-host", "", "", "hostname for the message provider (sqs/kafka)")
+	s3Cmd.Flags().StringVarP(&s3Opts.mpPort, "mp-port", "", "", "port for the message provider (sqs/kafka)")
+	s3Cmd.Flags().StringVarP(&s3Opts.queues, "queues", "", "", "comma-separated list of queue/topic names")
+	s3Cmd.Flags().StringVarP(&s3Opts.region, "region", "", "us-east-1", "aws region, defaults to us-east-1")
```
jeffmendoza, line 152: Please add these to `pkg/cli/store.go` and access them the same way as the other commands. Also add "s3-" to the front of all of them.

cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,154 @@
...
+	Use:   "s3 [flags]",
+	Short: "listens to kafka/sqs s3 events to download documents and add them to the GUAC graph",
+	Args:  cobra.MinimumNArgs(0),
+	Run: func(cmd *cobra.Command, args []string) {
+		ctx := logging.WithLogger(context.Background())
+		logger := logging.FromContext(ctx)
+
+		s3Opts, err := validateS3Opts(s3Opts.s3hostname, s3Opts.s3port, s3Opts.mp, s3Opts.mpHostname, s3Opts.mpPort, s3Opts.queues, s3Opts.region)
+		if err != nil {
+			os.Exit(1)
```
jeffmendoza, line 57: Print error and use ` cmd.Help()`

cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,154 @@
...
+			Queues:              s3Opts.queues,
+			Region:              s3Opts.region,
+		})
+		if err != nil {
+			logger.Errorf("unable to create s3 collector: %v", err)
+			os.Exit(1)
+		}
+
+		err = collector.RegisterDocumentCollector(s3Collector, s3.S3CollectorType)
+		if err != nil {
```
jeffmendoza, line 75: use `if err := collector.RegisterDocumentCollector(s3Collector, s3.S3CollectorType); err != nil`
Whenever there is just err as the return value.

This way you don't have to keep track if err is defined or not and use `=` vs `:=` based on that.

cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,154 @@
...
+			logger.Infof("collectsub client initialization failed, this ingestion will not pull in any additional data through the collectsub service: %v", err)
+			csubClient = nil
+		} else {
+			defer csubClient.Close()
+		}
+
+		errFound := false
+
+		_, s3ctx := errgroup.WithContext(ctx)
```
jeffmendoza, line 90: Doesn't look like this errgroup is being used.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+	S3Host              string
+	S3Port              string
+	Region              string // optional (defaults to us-east-1, assumes same region for s3 and sqs)
+	Queues              string
+	MpBuilder           messaging.MessageProviderBuilder // optional
+	BucketBuilder       bucket.BuildBucket               // optional
+	SigChan             chan os.Signal                   // optional
+}
+
+func NewS3Collector(cfg S3CollectorConfig) (*S3Collector, error) {
```
jeffmendoza, line 52: No need to have error return value if not used.

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+}
+
+func GetDefaultBucket(hostname string, port string, region string) Bucket {
+	return S3Bucket{hostname, port, region}
+}
+
+func (d S3Bucket) DownloadFile(ctx context.Context, bucket string, item string) ([]byte, error) {
+	cfg, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("error loading AWS SDK config: %v", err)
```
jeffmendoza, line 62: Change all `fmt.Errorf` that include errors to use `%w` to wrap the error.

pkg/handler/collector/s3/messaging/message.go
```diff
@@ -0,0 +1,67 @@
...
+func (mb MpBuilder) GetMessageProvider(config MessageProviderConfig) (MessageProvider, error) {
+	switch config.Provider {
+	case "sqs":
+		return NewSqsProvider(config)
+	default:
+		return NewKafkaProvider(config)
+	}
+}
+
+func GetDefaultMessageProviderBuilder() (MessageProviderBuilder, error) {
```
jeffmendoza, line 65: Looks like you don't need error here as well.

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,157 @@
...
+
+		messages := receiveOutput.Messages
+		if len(messages) > 0 {
+			message := receiveOutput.Messages[0]
+			logger.Debugf("Received message: %v\n", *message.Body)
+
+			var msg SqsMessage
+			err := json.Unmarshal([]byte(*message.Body), &msg)
+			if err != nil {
+				fmt.Println("Error:", err)
```
jeffmendoza, line 135: Change to logger or just wrap and return.

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,157 @@
...
+	addr := fmt.Sprintf("http://%s:%s/000000000000/%s", s.hostname, s.port, s.queue)
+
+	// Receive messages from the queue
+	receiveInput := &sqs.ReceiveMessageInput{
+		QueueUrl:            &addr,
+		MaxNumberOfMessages: 1,
+		WaitTimeSeconds:     10,
+	}
+
+	for {
```
jeffmendoza, line 120: If not messages are ever received, will this just loop forever?

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+			if err != nil {
+				logger.Errorf("error getting message provider for queue %v: %v", queue, err)
+				return
+			}
+
+			for {
+				select {
+				case <-sigChan:
+					logger.Infof("Shutting down collector for queue %s...\n", queue)
+					break
```
jeffmendoza, line 80: It's good to have this here, but it needs to be setup for `Ctrl-C`. The collector should cleanly shut down then.

Also, Ideally there are two versions of each collector `guacone collect <name>` is meant to be a user run collector that runs once and collects the files in the bucket/source, then exits. The `guaccollect <name>` command is meant to be run as a daemon/service and stay running and watch the bucket/source for new files.

jeffmendoza (commented):
Thanks for this! Some comments below.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+
+	var wg sync.WaitGroup
+	for _, queue := range queues {
+		wg.Add(1)
+
+		go func(queue string) {
+			mp, err := getMessageProvider(s, queue)
+			if err != nil {
+				logger.Errorf("error getting message provider for queue %v: %v", queue, err)
+				return
```
naveensrinivasan, line 73: The code has a never-ending loop in the goroutine that takes in messages. 
But, when sigChan gets a signal, the code just logs it and exits the select block. 

This triggers another loop cycle.

The code should exit the goroutine instead. After the code logs the shutdown message, use the return statement to leave the goroutine.

Here is a code example that can be demonstrated in a simple example.
``` golang
package main

import (
	"fmt"
	"time"
)

func main() {
	sigChan := make(chan struct{})

	go func() {
		for {
			select {
			case <-sigChan:
				fmt.Println("Received signal. Exiting goroutine!")
				return
				//break
				// instead if we use break it will continue to print "Working..."
			default:
				fmt.Println("Working...")
				time.Sleep(500 * time.Millisecond)
			}
		}
	}()

	// Simulate some work.
	time.Sleep(2 * time.Second)

	// Send signal to terminate.
	sigChan <- struct{}{}

	// Wait to observe the behavior.
	time.Sleep(5 * time.Second)
}
```

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+	queues := strings.Split(s.config.Queues, ",")
+	downloader := getDownloader(s)
+	sigChan := s.config.SigChan
+
+	var wg sync.WaitGroup
+	for _, queue := range queues {
+		wg.Add(1)
+
+		go func(queue string) {
+			mp, err := getMessageProvider(s, queue)
```
naveensrinivasan, line 70: The `defer wg.Done()` is missing.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+						continue
+					}
+					bucketName, err := m.GetBucket()
+					if err != nil {
+						logger.Errorf("skipping message: %v\n", err)
+						continue
+					}
+					item, err := m.GetItem()
+					if err != nil {
+						logger.Errorf("skipping message: %v\n", err)
```
naveensrinivasan, line 101: I think we need a `continue` to prevent processing with potentially invalid data.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,176 @@
...
+	logger := logging.FromContext(ctx)
+	queues := strings.Split(s.config.Queues, ",")
+	downloader := getDownloader(s)
+	sigChan := s.config.SigChan
+
+	var wg sync.WaitGroup
+	for _, queue := range queues {
+		wg.Add(1)
+
+		go func(queue string) {
```
naveensrinivasan, line 69: When working with channels, especially in a multi-goroutine environment, there's the potential risk of sending data on a closed channel. Sending data on a closed channel in Go will result in a runtime panic.

To avoid this panic, one common strategy is to use a recover() function within a defer statement at the beginning of a goroutine. This allows the program to gracefully handle the panic rather than crashing.

``` golang
package main

import (
	"fmt"
	"time"
)

func main() {
	ch := make(chan int)

	go func() {
		defer func() {
			if r := recover(); r != nil {
				fmt.Println("Recovered from panic:", r)
			}
		}()

		ch <- 1 // attempt to send data to the channel
	}()

	close(ch) // close the channel
	time.Sleep(2 * time.Second)
}
```

naveensrinivasan (commented):
A few suggestions 👍 . Thank you!

naveensrinivasan (commented):
A few more comments 👍

pkg/handler/collector/s3/messaging/sqs.go
jmle: This client has a polling behaviour, but yes, the idea is for this to keep running and listening.

pkg/handler/collector/s3/s3.go
jmle: Thank you, good catch

pkg/handler/collector/s3/s3.go
dejanb: @jeffmendoza I wasn't aware that difference between `guacone collect` and `guaccollect` was in the daemon and user use case. In my mind it was difference between direct gql ingest vs nats+ingestor pipeline. For example osv certifier is also used as a running daemon with guacone.
In any case, is it then better to move this command to guaccollect and with current gql ingestion? And then in a subsequent PR implement a `guacone collect s3` command that collects directly from the bucket (without sqs/kafka notifications)?

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+
+	resp, err := client.GetObject(ctx, input)
+	if err != nil {
+		return nil, fmt.Errorf("unable to download file: %v", err)
+	}
+	defer resp.Body.Close()
+
+	buf := new(bytes.Buffer)
+	n, err := io.Copy(buf, resp.Body)
+	if err != nil || n == 0 {
```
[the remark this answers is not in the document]
jmle: After testing locally, it was a potential error scenario that I encountered, therefore the decision

jmle:
@jeffmendoza @naveensrinivasan thank you for your reviews - I'm fairly new to Go and there are conventions and best practices that I'm still not very familiar with.

pkg/handler/collector/s3/s3.go
jeffmendoza: @dejanb You are right, the fundamental difference between the two is direct gql vs nats+ingestor pipeline, and it should stay like that for now.

This expanded into default poll behavior though, assuming the collector that uses nats (guaccollect) should be part of a deployment and run as a service, it should default poll (with optional one-time), and conversely the collector that is direct gql (guacone collect) could be run anywhere, and likely by a user, it should default to run-once (with optional poll). The two cli options are:
https://github.com/guacsec/guac/blob/main/pkg/cli/store.go#L71-L72

In general "guacone" should be the do-all cli for user oriented commands, and service-oriented binaries should be more unixy (do one thing well), and seperate.

Certifiers don't have this yet, but I think they should.

This is covered a bit in these issues: https://github.com/guacsec/guac/issues/719 https://github.com/guacsec/guac/issues/809 I need to work on converting these to developer docs.

dejanb: Thanks @jeffmendoza. So are you OK with leaving things as is in this PR for now? For me it would be ideal that `guaccollect` have options of nats, graphql or ingestor API (#1177) when it's ready. And then we can refactor all collectors to use any of the configured options.
I wouldn't like to refactor this collector to use nats as we are doing this to avoid that exact scenario. So we would like to have option for a long running collector that uses gql. Please let us know what is the best place for this at the moment.

jeffmendoza: Just opened up #1324 to cover some of this.

Yes, we should do just this `guacone` collector first that uses direct gql ingest. No need to also do `guaccollect` at this time, that can be done later if needed.

I do think we need this default to non-polling. Adding the polling option is great, and agree that this should be able to do so. When polling is enabled, make sure to respect `SIGINT/SIGTERM`.

jmle: @jeffmendoza I have added non-polling behaviour and set it as a default. Non-polling behaviour will download a single file from a specified bucket from S3 and ingest it into guac, then return. I guess in the future this chunk will stay here and the polling logic will be moved to guaccollect.

jmle:
Hi @jeffmendoza, I implemented the requested changes, is there anything else I can do to get this merged?

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+type BuildBucket interface {
+	GetDownloader(hostname string, port string, region string) Bucket
+}
+
+type BucketBuilder struct {
+}
+
+func (bd BucketBuilder) GetBucket(hostname string, port string, region string) Bucket {
```
lumjjb, line 36: nit: For consistency we use pointer receivers `(bd *BucketBuilder)`

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+	hostname string
+	port     string
+	region   string
+}
+
+func GetDefaultBucket(hostname string, port string, region string) Bucket {
+	return S3Bucket{hostname, port, region}
+}
+
+func (d S3Bucket) DownloadFile(ctx context.Context, bucket string, item string) ([]byte, error) {
```
lumjjb, line 59: pointer receiver

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+		region,
+	}
+}
+
+type Bucket interface {
+	DownloadFile(ctx context.Context, bucket string, item string) ([]byte, error)
+	GetEncoding(ctx context.Context, bucket string, item string) (string, error)
+}
+
+type S3Bucket struct {
```
lumjjb, line 49: it seems like this struct can be private since its public interface since its type is never returned

pkg/handler/collector/s3/bucket/bucket.go
```diff
@@ -0,0 +1,116 @@
...
+		o.BaseEndpoint = aws.String(addr)
+	})
+
+	headObject, err := client.HeadObject(context.Background(), &s3.HeadObjectInput{Bucket: aws.String(bucket), Key: aws.String(item)})
+	if err != nil {
+		return "", fmt.Errorf("could not get head object: %w", err)
+	}
+
+	if headObject.ContentEncoding == nil {
+		return "", nil
```
lumjjb, line 112: should this be an error? or can this be handled?

pkg/handler/collector/s3/bucket/encoding.go
```diff
@@ -0,0 +1,52 @@
...
+const (
+	BZIP2   = "BZIP2"
+	ZSTD    = "ZSTD"
+	UNKNOWN = "UNKNOWN"
+)
+
+func ExtractEncoding(encoding string, filename string) processor.EncodingType {
+	switch encoding {
+	case BZIP2:
+		return BZIP2
```
lumjjb, line 33: use the enums in the [processor](https://github.com/guacsec/guac/blob/main/pkg/handler/processor/processor.go#L86-L87), and if not present, add them to the file

pkg/handler/collector/s3/messaging/kafka.go
```diff
@@ -0,0 +1,108 @@
...
+	// Kafka-specific configuration fields
+	reader *kafka.Reader
+}
+
+type KafkaMessage struct {
+	EventName string `json:"EventName"`
+	Key       string `json:"Key"`
+}
+
+func (m KafkaMessage) GetEvent() (EventName, error) {
```
lumjjb, line 38: nit: pointer receiver

pkg/handler/collector/s3/messaging/kafka.go
```diff
@@ -0,0 +1,108 @@
...
+	kafkaHostname := mpConfig.Host
+	kafkaPort := mpConfig.Port
+	kafkaTopic := mpConfig.Queue
+
+	kafkaProvider := KafkaProvider{}
+	kafkaProvider.reader = kafka.NewReader(kafka.ReaderConfig{
+		Brokers:   []string{fmt.Sprintf("%s:%s", kafkaHostname, kafkaPort)},
+		Topic:     kafkaTopic,
+		Partition: 0,
+		MaxBytes:  10e6,
```
lumjjb, line 71: can you add a comment for the magic number?

pkg/handler/collector/s3/messaging/message.go
```diff
@@ -0,0 +1,67 @@
...
+	Provider string
+	Region   string
+}
+
+// MessageProviderBuilder Returns a builder for a MessageProvider
+type MessageProviderBuilder interface {
+	GetMessageProvider(config MessageProviderConfig) (MessageProvider, error)
+}
+
+type MpBuilder struct {
```
lumjjb, line 52: make private

pkg/handler/collector/s3/messaging/message.go
```diff
@@ -0,0 +1,67 @@
...
+	GetItem() (string, error)
+}
+
+// MessageProvider Reads and returns messages from a given queue (or topic)
+type MessageProvider interface {
+	ReceiveMessage(ctx context.Context) (Message, error)
+	Close(ctx context.Context) error
+}
+
+type MessageProviderConfig struct {
```
lumjjb, line 39: make private

pkg/handler/collector/s3/s3_test.go
```diff
@@ -0,0 +1,219 @@
...
+	// wait for a while to get some messages
+	time.Sleep(5 * time.Second)
+	// shut down collector
+	signal.Notify(sigChan, syscall.SIGINT)
+
+	w.Close()
+	os.Stdout = oldStdout // restoring the real stdout
+	out := <-outC
+
+	fmt.Println(out)
```
lumjjb, line 166: remove print statement of output, error messages should be sufficient

lumjjb (commented):
Awesome work! Thanks for the great contribution @jmle! Looks good overall, just a couple nits and requested changes, else LGTM!

pkg/cli/store.go
```diff
@@ -87,6 +87,18 @@ func init() {
...
+	set.String("s3-host", "", "hostname for the s3 provider")
+	set.String("s3-port", "", "port for the s3 provider")
+	set.String("s3-bucket", "", "bucket in the s3 provider")
+	set.String("s3-item", "", "item in the s3 provider")
+	set.String("s3-mp", "kafka", "message provider (sqs or kafka)")
+	set.String("s3-mp-host", "", "hostname for the message provider")
+	set.String("s3-mp-port", "", "port for the message provider")
+	set.String("s3-queues", "", "comma-separated list of queue/topic names")
+	set.String("s3-region", "us-east-1", "aws region")
+	set.Bool("s3-poll", false, "polling or non-polling behaviour")
```
jeffmendoza, line 100: Use the existing "poll" option, no need to add an s3 specific one.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,239 @@
...
+	logger := logging.FromContext(ctx)
+	downloader := getDownloader(s)
+
+	bckt := s.config.S3Bucket
+	item := s.config.S3Item
+
+	blob, err := downloader.DownloadFile(ctx, bckt, item)
+	if err != nil {
+		logger.Errorf("could not download item %v: %v", item, err)
+		return
```
jeffmendoza, line 83: This function should return this error, which RetrieveArtifacts should then return. Same with the one a couple lines down.

cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func validateS3Opts(s3hostname string, s3port string, s3bucket string, s3item string, mp string, mpHostname string, mpPort string, queues string, region string, poll bool) (s3Options, error) {
+	var opts s3Options
+	if len(s3hostname) == 0 {
+		return opts, fmt.Errorf("expected s3 hostname")
+	}
+	if len(s3port) == 0 {
+		return opts, fmt.Errorf("expected s3 port")
+	}
```
jeffmendoza, line 148: Options that are non-optional shouln't be options. If hostname and port are always required, make them positional arguments. 
ex: https://github.com/guacsec/guac/blob/main/cmd/guacone/cmd/vulnerability.go#L503-L507
https://github.com/guacsec/guac/blob/main/cmd/guacone/cmd/vulnerability.go#L52
https://github.com/guacsec/guac/blob/main/cmd/guacone/cmd/vulnerability.go#L63

cmd/guacone/cmd/s3.go
```diff
@@ -0,0 +1,186 @@
...
+		if len(queues) == 0 {
+			return opts, fmt.Errorf("expected at least one queue")
+		}
+	} else {
+		if len(s3bucket) == 0 {
+			return opts, fmt.Errorf("expected s3 bucket")
+		}
+		if len(s3item) == 0 {
+			return opts, fmt.Errorf("expected s3 item")
+		}
```
jeffmendoza, line 165: Up to you, but if item is not provided is there a way to collect every file in the bucket?

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,162 @@
...
+		select {
+		case <-ctx.Done():
+			return SqsMessage{}, nil
```
jeffmendoza, line 123: Return `ctx.Err()` as well.

pkg/handler/collector/s3/messaging/sqs.go
jeffmendoza: There is polling happening at https://github.com/guacsec/guac/pull/1308/files#diff-0b3e316624485f7c06c51b3fad631283726dde094f05f3369777eafe4df3c1b4R140 which will repeatedly call `ReceiveMessage()` here. I don't see the need for the for loop here.

pkg/handler/collector/s3/messaging/kafka.go
```diff
@@ -0,0 +1,108 @@
...
+	msg := KafkaMessage{}
+	err = json.Unmarshal(m.Value, &msg)
+	if err != nil {
+		return msg, fmt.Errorf("error parsing JSON: %w", err)
+	}
+
+	return msg, err
+}
+
+func (k KafkaProvider) Close(ctx context.Context) error {
```
jeffmendoza, line 99: I don't see s3.go calling this close anywhere.

pkg/handler/collector/s3/bucket/bucket.go
jmle: It's not an error, I have found in my tests `ContentEncoding` to be nil sometimes, so I added this to avoid a null pointer dereference. The empty encoding is dealt with properly afterwards.

pkg/handler/collector/s3/messaging/kafka.go
jmle: Removing it - default is 1MB, which I think is good enough.

cmd/guacone/cmd/s3.go
jmle: It could be a good addition, but I would leave it for another PR, what do you think @dejanb ?

dejanb: Yeah, IMHO we can address this in a follow-up PR. Let's open an issue once this one is merged.

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	select {
+	case <-ctx.Done():
+		return &SqsMessage{}, ctx.Err()
+	default:
+		receiveOutput, err := s.client.ReceiveMessage(ctx, receiveInput)
+		if err != nil {
+			fmt.Printf("error receiving message, skipping: %v\n", err)
+			//continue
```
jeffmendoza, line 127: Should return an error here.

pkg/handler/collector/s3/messaging/kafka.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return kafkaProvider, nil
+}
+
+func (k *KafkaProvider) ReceiveMessage(ctx context.Context) (Message, error) {
+	logger := logging.FromContext(ctx)
+
+	m, err := k.reader.ReadMessage(ctx)
+	if err != nil {
+		fmt.Println(err.Error())
```
jeffmendoza, line 85: return error here.

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,236 @@
...
+						Blob:     blob,
+						Type:     processor.DocumentUnknown,
+						Format:   processor.FormatUnknown,
+						Encoding: bucket.ExtractEncoding(enc, item),
+						SourceInformation: processor.SourceInformation{
+							Collector: S3CollectorType,
+							Source:    "S3",
+						},
+					}
+					docChannel <- doc
```
naveensrinivasan, line 190: A suggestion
```golang 
select {
case docChannel <- doc:
case <-cncCtx.Done():
    logger.Infof("Shutting down collector for queue %s...\n", queue)
    return
}
```

lulf:
This should be ready now. The e2e check failure seems unrelated to this PR, but the other things should be sorted out. I'm not the creator of the PR so I'm not able to resolve the remaining threads.

dejanb:
+1 from my side as well ... We tested it now extensively within our project

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,170 @@
...
+		WaitTimeSeconds:     10,
+	}
+
+	select {
+	case <-ctx.Done():
+		return &SqsMessage{}, ctx.Err()
+	default:
+		receiveOutput, err := s.client.ReceiveMessage(ctx, receiveInput)
+		if err != nil {
+			return &SqsMessage{}, fmt.Errorf("error receiving message, skipping: %w\n", err)
```
lumjjb, line 136: nit: err messages probably shouldn't have new lines?

lumjjb (approved):
very minor nit around error and log messages having new lines, otherwise lgtm!

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,170 @@
...
+	gQInput := &sqs.GetQueueUrlInput{
+		QueueName: &s.queue,
+	}
+
+	// Get URL of queue
+	urlResult, err := s.client.GetQueueUrl(ctx, gQInput)
+	if err != nil {
+		fmt.Println("Got an error getting the queue URL:")
+		fmt.Println(err)
+		return nil, err
```
jeffmendoza, line 118: No fmt.Print
`return nil, fmt.Errorf("Got an error getting the queue URL : %w", err)`

pkg/handler/collector/s3/s3.go
```diff
@@ -0,0 +1,229 @@
...
+	var wg sync.WaitGroup
+	for _, queue := range queues {
+		wg.Add(1)
+
+		go func(wg *sync.WaitGroup, cncCtx context.Context, queue string) {
+			defer wg.Done()
+
+			defer func() {
+				if r := recover(); r != nil {
+					fmt.Println("recovered from panic:", r)
```
jeffmendoza, line 116: `logger.Errorf("recovered from panic: %v", r)`

pkg/handler/collector/s3/s3_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// spawn collector
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	cancelCtx, cancel := context.WithCancel(ctx)
+	go func() {
+		err := collector.Collect(cancelCtx, em, eh)
+		if err != nil {
+			fmt.Printf("error collecting: %v", err)
```
jeffmendoza, line 140: `t.Errorf("error collecting: %v", err)`

pkg/handler/collector/s3/s3_test.go
```diff
@@ -0,0 +1,203 @@
...
+		s = append(s, d)
+		return nil
+	}
+	eh := func(err error) bool {
+		return true
+	}
+
+	err := collector.Collect(ctx, em, eh)
+	if err != nil {
+		fmt.Printf("error collecting: %v", err)
```
jeffmendoza, line 193: `t.Errorf("error collecting: %v", err)`

pkg/handler/collector/s3/messaging/sqs.go
```diff
@@ -0,0 +1,170 @@
...
+			}
+
+			// Delete the received message from the queue (stardard sqs procedure)
+			deleteInput := &sqs.DeleteMessageInput{
+				QueueUrl:      addr,
+				ReceiptHandle: message.ReceiptHandle,
+			}
+			_, err = s.client.DeleteMessage(context.TODO(), deleteInput)
+			if err != nil {
+				logger.Errorf("error deleting message: %v\n", err)
```
jeffmendoza, line 158: `return nil, fmt.Errorf("error deleting message: %w", err)`

jeffmendoza (commented):
Some fmt.Print to clean up and one dropped error.

jeffmendoza (approved):

PR #1309, opened 2023-09-22
OSV certifier: bulk ingest

mrizzi:
# Description of the PR

Currently the OSV certifier ingests one `CertifuVuln` (with corresponding `Package` and `Vulnerability`) one by one without fully leveraging the guac's bulk ingestion capabilities.
It looks related to invoking `ingestor.Ingest` for each `processor.Document`
I've tried to collect all the `processor.Document`s first and then create a new `ingestor.BulkIngest` func that collects all of the `assembler.IngestPredicates` (generated by `ingestorFunc`) into the 0th element of `predicates` array.

In this way, the call to `assemblerFunc` \ `GetBulkAssembler` will send all the data invoking each mutation just once.

Since I've never worked on this part of guac, I've added only `CertifyVuln` and `VulnEqual` on purpose in order to get feedback and if the approach is fine, I'll extend to other arrays within `IngestPredicates` struct

# PR Checklist

### DIFF
--- a/pkg/ingestor/ingestor.go
+++ b/pkg/ingestor/ingestor.go
@@ -69,6 +69,50 @@ func Ingest(ctx context.Context, d *processor.Document, graphqlEndpoint string, 
 	return nil
 }
 
+func BulkIngest(ctx context.Context, docs []*processor.Document, graphqlEndpoint string, csubClient csub_client.Client) error {
+	logger := logging.FromContext(ctx)
+	// Get pipeline of components
+	processorFunc := GetProcessor(ctx)
+	ingestorFunc := GetIngestor(ctx)
+	collectSubEmitFunc := GetCollectSubEmit(ctx, csubClient)
+	assemblerFunc := GetAssembler(ctx, graphqlEndpoint)
+
+	start := time.Now()
+
+	var predicates = make([]assembler.IngestPredicates, 1)
+	var idstrings []*parser_common.IdentifierStrings
+	for _, d := range docs {
+		docTree, err := processorFunc(d)
+		if err != nil {
+			return fmt.Errorf("unable to process doc: %v, format: %v, document: %v", err, d.Format, d.Type)
+		}
+
+		preds, idstrs, err := ingestorFunc(docTree)
+		if err != nil {
+			return fmt.Errorf("unable to ingest doc tree: %v", err)
+		}
+		for i := range preds {
+			predicates[0].CertifyVuln = append(predicates[0].CertifyVuln, preds[i].CertifyVuln...)
+			predicates[0].VulnEqual = append(predicates[0].VulnEqual, preds[i].VulnEqual...)
+		}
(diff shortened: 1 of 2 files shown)


pkg/ingestor/ingestor.go
```diff
@@ -69,6 +69,50 @@ func Ingest(ctx context.Context, d *processor.Document, graphqlEndpoint string,
...
+			return fmt.Errorf("unable to process doc: %v, format: %v, document: %v", err, d.Format, d.Type)
+		}
+
+		preds, idstrs, err := ingestorFunc(docTree)
+		if err != nil {
+			return fmt.Errorf("unable to ingest doc tree: %v", err)
+		}
+		for i := range preds {
+			predicates[0].CertifyVuln = append(predicates[0].CertifyVuln, preds[i].CertifyVuln...)
+			predicates[0].VulnEqual = append(predicates[0].VulnEqual, preds[i].VulnEqual...)
```
jeffmendoza, line 96: Is the idea here that we want to merge the predicates because each doc will only have 1 certify/vuln?

The assemblerFunc we are using is already called "bulk" because it does the bulk ingest per "predicates" Maybe we should call this "MergedIngest"?

Also, we should merge all the predicates here, or name this MergedVulnIngest or similar.

mrizzi: > Is the idea here that we want to merge the predicates because each doc will only have 1 certify/vuln?

Each `processor.Document` could potentially generate multiple `assembler.CertifyVulnIngest` because the attestation's statement has an array of subjects.
This is the reason for having the `for` over `preds`.

> The assemblerFunc we are using is already called "bulk" because it does the bulk ingest per "predicates" Maybe we should call this "MergedIngest"?

Yep, sure, makes sense.

> Also, we should merge all the predicates here, or name this MergedVulnIngest or similar.

Fine for merging all predicates here. I didn't do because I wanted to have a feedback but I'm going to add the missing predicates.

pkg/ingestor/ingestor.go
```diff
@@ -92,8 +92,23 @@ func BulkIngest(ctx context.Context, docs []*processor.Document, graphqlEndpoint
 			return fmt.Errorf("unable to ingest doc tree: %v", err)
 		}
 		for i := range preds {
+			predicates[0].CertifyScorecard = append(predicates[0].CertifyScorecard, preds[i].CertifyScorecard...)
```
pxp928, line 95: so we might have to put a limit here on how many we transfer to the graphQL server. After a certain limit, send to the graphQL server and start collecting again.

mrizzi: I put a limit at 5000 `assembler.IngestPredicates` merged but I have no strong opinion on how much it's a good value so feel free about suggesting a more appropriate value in case.

jeffmendoza (approved):

mlieberman85 (approved):
LGTM

PR #1332, opened 2023-09-28
Change version list to map

### LINKED ISSUE #1234: [Optimization] In-mem getPackageIDFromInput
There's a function for retrieval of packages in [getPackageIDFromInput](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/inmem/pkg.go#L656), which I hypothesize is causing some slow down in the ingestion for in-mem backend.

I suspect that the issue is within the [version loop](https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/inmem/pkg.go#L674-L688). If this can be optimized into a hash map, that would be great.

mdeicas:
# Description of the PR

Changes the representation of versions in the inmem database to a map instead of a list, for better performance. 

The keys of this map are created by joining the version, subpath, and sorted qualifiers of a `pkgVersionNode`, and hashing the result. 

There is no longer an order to the versions, so tests in `pkg_test.go` and `pkgEqual_test.go` didn't pass consistently. As a result I wrote some helper functions to sort all of the nested lists in the nodes. `pkg_test.go` and `pkgEqual_test.go` are in different packages (`inmem` and `inmem_test`), and I don't think that `pkg_test.go` can be easily changed to be in the `inmem_test` package, so I added the helper functions in two duplicate files. This isn't ideal, so let me know if there is a better solution to this. 

Makes progress on #1234

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/inmem/pkg.go
+++ b/pkg/assembler/backends/inmem/pkg.go
@@ -17,10 +17,13 @@ package inmem
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
 	"strconv"
+	"strings"
 
 	"github.com/vektah/gqlparser/v2/gqlerror"
 
@@ -98,7 +101,7 @@ type pkgVersionStruct struct {
 	id                  uint32
 	parent              uint32
 	name                string
-	versions            pkgVersionList
+	versions            pkgVersionMap
 	srcMapLinks         []uint32
 	isDependencyLinks   []uint32
 	badLinks            []uint32
@@ -106,7 +109,8 @@ type pkgVersionStruct struct {
 	hasMetadataLinks    []uint32
 	pointOfContactLinks []uint32
 }
-type pkgVersionList []*pkgVersionNode
+type pkgVersionNodeHash string
+type pkgVersionMap map[pkgVersionNodeHash]*pkgVersionNode
 type pkgVersionNode struct {
 	id                  uint32
 	parent              uint32
@@ -386,20 +390,21 @@ func (c *demoClient) IngestPackage(ctx context.Context, input model.PkgInputSpec
 				id:       c.getNextID(),
 				parent:   namesStruct.id,
 				name:     input.Name,
-				versions: pkgVersionList{},
+				versions: pkgVersionMap{},
 			}
 			c.index[versionStruct.id] = versionStruct
 			names[input.Name] = versionStruct
 		}
 		c.m.Unlock()
 	}
+	versions := versionStruct.versions
 
 	c.m.RLock()
-	duplicate, collectedVersion := duplicatePkgVer(versionStruct.versions, input)
+	duplicate, collectedVersion := duplicatePkgVer(versions, input)
 	c.m.RUnlock()
 	if !duplicate {
 		c.m.Lock()
-		duplicate, collectedVersion = duplicatePkgVer(versionStruct.versions, input)
+		duplicate, collectedVersion = duplicatePkgVer(versions, input)
 		if !duplicate {
 			collectedVersion = &pkgVersionNode{
 				id:         c.getNextID(),
@@ -409,8 +414,13 @@ func (c *demoClient) IngestPackage(ctx context.Context, input model.PkgInputSpec
 				qualifiers: getQualifiersFromInput(input.Qualifiers),
 			}
 			c.index[collectedVersion.id] = collectedVersion
-			// Need to append to version and replace field in versionStruct
-			versionStruct.versions = append(versionStruct.versions, collectedVersion)
+
+			versionDigest, err := hashPkgVersionNode(collectedVersion)
+			if err != nil {
+				c.m.Unlock()
+				return nil, err
+			}
+			versions[versionDigest] = collectedVersion
 		}
 		c.m.Unlock()
 	}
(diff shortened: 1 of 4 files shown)


pkg/assembler/backends/inmem/pkg.go
```diff
@@ -421,20 +436,58 @@ func (c *demoClient) IngestPackage(ctx context.Context, input model.PkgInputSpec
...
+	for _, qualifier := range qualifierSlice {
+		_, err = hash.Write([]byte(qualifier))
+		if err != nil {
+			return "", fmt.Errorf("Error writing when computing the hash of a version: %v", err)
 		}
-		return true, v
 	}
-	return false, nil
+
+	digest := hash.Sum([]byte{})
```
lumjjb, line 477: it seems like the key that is being hashed may not be unique. For example, if we have:

version: "v1"
subpath: ".2-abc"

vs 

version: "v1.2-abc"
subpath: ""

pkg/assembler/backends/inmem/pkgEqual_test.go
```diff
@@ -657,6 +661,17 @@ func TestIngestPkgEquals(t *testing.T) {
...
+
+			// make the expected input and output canonical
+			for _, pkgEquals := range test.ExpHE {
+				makeCanonicalPkgEquals(pkgEquals)
+			}
+			for _, pkgEquals := range got {
+				makeCanonicalPkgEquals(pkgEquals)
+			}
+			slices.SortFunc(test.ExpHE, lessPkgEquals)
+			slices.SortFunc(got, lessPkgEquals)
```
lumjjb, line 673: hmm.. can we fold this into a cmpopt?

pkg/assembler/backends/inmem/testhelpers.go
```diff
@@ -0,0 +1,147 @@
...
+// the inmem package, so it cannot use the definitions in testhelpers_test.go
+
+////////////////////// comparison functions //////////////////////
+
+func lessPackageQualifier(a, b *model.PackageQualifier) int {
+	return cmp.Compare(a.Key+a.Value, b.Key+b.Value)
+}
+
+func lessPackageVersion(a, b *model.PackageVersion) int {
+	if cmp.Compare(a.Version, b.Version) != 0 {
```
lumjjb, line 22: nit: can store the result and return. example:

```
if res := cmp.Compare(...); res != 0 {
    return res
}
```

pkg/assembler/backends/inmem/testhelpers.go
```diff
@@ -0,0 +1,147 @@
+package inmem
```
lumjjb, line 1: wonder if this can be renamed to a `_test.go` file which will only be compiled if running a test.

lumjjb: o wait this already exists, is this a duplicate file?

pkg/assembler/backends/inmem/pkg.go
lumjjb: May consider using `fmt.Sprintf` to construct the string first and then hash it, will be a little less efficient but more readable

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -98,15 +100,15 @@ type pkgVersionStruct struct {
...
+	versions            pkgVersionMap
 	srcMapLinks         []uint32
 	isDependencyLinks   []uint32
 	badLinks            []uint32
 	goodLinks           []uint32
 	hasMetadataLinks    []uint32
 	pointOfContactLinks []uint32
 }
-type pkgVersionList []*pkgVersionNode
+type pkgVersionMap map[string]*pkgVersionNode
```
lumjjb, line 111: to make things a bit more clearer, could define a type for `pkgVersionNodeHash` and then have `hashPkgVersionNode` return that type

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -421,20 +436,58 @@ func (c *demoClient) IngestPackage(ctx context.Context, input model.PkgInputSpec
...
+		return false, nil, err
+	}
+
+	if version, ok := versions[digest]; ok {
+		return true, version, nil
+	}
+	return false, nil, nil
 }
 
 // Query Package
```
lumjjb, line 493: I think we can also leverage the map in the query in the query flow (i.e. in `Packages` flow in `buildPkgName`), we may want to check the flow in which `IsDependency` and `IsOccurrence` and other evidence trees lookup packages as well..

If you'd like we can do this as a separate PR as well. Just note it in the original issue.

lumjjb (commented):
Awesome work @mdeicas - needs the golang update and a couple comments and a follow up on applying the change to the query/lookup flows as well (potentially as a follow up PR)

pkg/assembler/backends/inmem/pkgEqual_test.go
mdeicas: Yes that would work, but taking order into account this way isn't consistent with how the makeCanonical functions do it. Instead, I [factored this section of code out](https://github.com/mdeicas/guac/blob/e05bdf2ed900d77dfaae1eec29993309cc2f17cf/pkg/assembler/backends/inmem/pkg_test.go#L758) to a helper function. 

Is this OK or do you still prefer to use cmpopts to sort the top level list of PkgEqual nodes?

lumjjb: that's fine - maybe we can add a TODO there to see if we make a cmpopts to it. I think the `MakeCanonicalPackageSlice`  function probably should be private.

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -421,18 +430,38 @@ func (c *demoClient) IngestPackage(ctx context.Context, input model.PkgInputSpec
...
+func hashPkgInputSpecVersion(input model.PkgInputSpec) pkgVersionNodeHash {
+	qualifiers := getQualifiersFromInput(input.Qualifiers)
+	return hashVersionHelper(nilToEmpty(input.Version), nilToEmpty(input.Subpath), qualifiers)
+}
+
+func hashVersionHelper(version string, subpath string, qualifiers map[string]string) pkgVersionNodeHash {
+	// first sort the qualifiers
+	qualifierSlice := []string{}
+	for key, value := range qualifiers {
+		qualifierSlice = append(qualifierSlice, fmt.Sprintf("%s:%s,", key, value))
```
naveensrinivasan, line 451: In the current implementation, each key-value pair in the qualifiers map is formatted as `%s:%s,` appending a comma after each pair.

This leads to a trailing comma after the last qualifier when all the qualifiers are concatenated. This trailing comma could cause inconsistencies or issues when parsing or comparing the hashed string later.

``` golang
func hashVersionHelper(version string, subpath string, qualifiers map[string]string) pkgVersionNodeHash {
    // First sort the qualifiers.
    qualifierSlice := make([]string, 0, len(qualifiers))
    for key, value := range qualifiers {
        qualifier := fmt.Sprintf("%s:%s", key, value)
        qualifierSlice = append(qualifierSlice, qualifier)
    }
    slices.Sort(qualifierSlice)
    
    // Join the sorted qualifiers with a comma.
    qualifiersStr := strings.Join(qualifierSlice, ",")
    
    // Create the canonical version.
    canonicalVersion := fmt.Sprintf("%s,%s,%s", version, subpath, qualifiersStr)
    
    // Hash the canonical version.
    digest := sha256.Sum256([]byte(canonicalVersion))
    return pkgVersionNodeHash(fmt.Sprintf("%x", digest))
}
```

naveensrinivasan (commented):
Thanks,👍 

A suggestion that could be an issue.

pkg/assembler/backends/inmem/pkg.go
```diff
@@ -17,9 +17,11 @@ package inmem
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"reflect"
+	"slices"
```
mihaimaruseac, line 24: CI seems to fail here with 

```
package slices is not in GOROOT (/opt/hostedtoolcache/go/1.19.13/x64/src/slices)
```

Might need to first see how to fix that (separate PR) before we can land this PR

pkg/assembler/backends/inmem/pkg.go
mihaimaruseac: +1 to this, we need to remove the last comma

mihaimaruseac (approved):

pkg/assembler/backends/inmem/pkg.go
mdeicas: I added this in, thanks!

pkg/assembler/backends/inmem/pkg.go
mdeicas: Sounds good, I will follow up with another PR

lumjjb (approved):

PR #1333, opened 2023-09-28
update go version

mdeicas:
# Description of the PR

Update the go version to 1.21. #1332 uses the packages "slices" and "cmp" which were introduced to the standard library in this release ([See the release notes](https://go.dev/doc/go1.21#slices). )

# PR Checklist


lumjjb:
I think you may have to run `go mod tidy` with the new version and push the changes as well

pkg/collectsub/datasource/csubsource/csubsource.go
```diff
@@ -57,7 +57,6 @@ func (d *csubDataSources) GetDataSources(ctx context.Context) (*datasource.DataS
 		return nil, err
 	}
 	ds := entriesToSources(ctx, entries)
-	d.lastEntries = ds
```
pxp928, line 60: what is the reason for this change?

pxp928 (approved):

pkg/collectsub/datasource/csubsource/csubsource.go
mdeicas: This is what causes the data race (thanks to @lumjjb for finding it). To my understanding, this is an optimization, to write to `lastEntries` before returning. I think it's a small optimization, so I removed it instead of adding locking to the csubDataSources struct.

mdeicas: For reference, the failing test can be seen in the unit tests of [this](https://github.com/guacsec/guac/actions/runs/6344018675) ci job (I force pushed, so the CI run isn't visible in this PR thread anymore).

jeffmendoza:
`golang.org/x/exp/slices` is used in quite a few tests. I assume those should all be changed to use the standard library ones now?

mdeicas:
> `golang.org/x/exp/slices` is used in quite a few tests. I assume those should all be changed to use the standard library ones now?

True, good point. One difference is that the standard library [version](https://pkg.go.dev/slices) uses comparison functions that return integers, while the [version used of golang.org/x/exp/slices](https://pkg.go.dev/golang.org/x/exp@v0.0.0-20230713183714-613f0c0eb8a1/slices) uses comparison functions that return booleans. The code would have to be updated in roughly [these places](https://github.com/search?q=repo%3Aguacsec%2Fguac+%2Fslices.*Func%2F&type=code). 

I'm happy to do it if you think we should.

lumjjb:
Can probably open up as an issue as a good first issue!

lumjjb (approved):

pxp928 (approved):

mdeicas:
> Can probably open up as an issue as a good first issue!

Opened in #1350, but I don't think I have permission to apply a label @lumjjb

pxp928:
> > Can probably open up as an issue as a good first issue!
> 
> Opened in #1350, but I don't think I have permission to apply a label @lumjjb

added the labels

Issue #1336, opened 2023-10-01
[bug] Places where race conditions in testing fails 


neilnaveen:
**Describe the bug**
Based on https://github.com/guacsec/guac/issues/1335, I enabled race conditions on the test and ran into these issues 

```
Test_SearchSubgraphFromVuln (0.00s)  
panic: http: multiple registrations for /query [recovered]  
    panic: http: multiple registrations for /query 

Test_gitCol_RetrieveArtifacts/get_repo  
    git_test.go:124: number of files collected does not match test = 10, want 9 
   FAIL: Test_gitCol_RetrieveArtifacts (1.89s)  
    --- FAIL: Test_gitCol_RetrieveArtifacts/get_repo (0.35s)

Test_DsseParser/testing  
{"level":"warn","ts":1696190765.0445309,"caller":"dsse/parser_dsse.go:74","msg":"DSSE verification currently not implemented in this release. Continuing without DSSE verification"}  
--- FAIL: Test_DsseParser (0.00s)

TestVerifyIdentity  
    verifier_test.go:132: RegisterVerifier() failed with error: the verification provider is being overwritten: sigstore  
=== RUN   TestVerifyIdentity/verify  
=== RUN   TestVerifyIdentity/not_found  
--- FAIL: TestVerifyIdentity (0.00s)

Test_prom_NewCounter/default  
--- FAIL: Test_prom_NewCounter (0.00s)  
    --- FAIL: Test_prom_NewCounter/default (0.00s)  
panic: duplicate metrics collector registration attempted [recovered]  
    panic: duplicate metrics collector registration attempted
```
    

**To Reproduce**

Add `-count=2` to the `make test` target in the Makefile.

**Expected behavior**
No errors should pop up 

**GUAC version**
latest commit 

### Race Condition Failures 

- [ ] Test_SearchSubgraphFromVuln 
- [ ] Test_gitCol_RetrieveArtifacts
- [ ] Test_DsseParser
- [ ] TestVerifyIdentity  
- [ ] Test_prom_NewCounter

neilnaveen:
> good find @neilnaveen. Are you working to fix these?

Yes I am working to fix these.

lumjjb:
Quick note - many of these are not race conditions but that the tests are not idempotent. I.e. running them twice will not yield the same result. The fix would be to do ignore errors that are not a big deal or do proper setup and tear down of tests.

PR #1345, opened 2023-10-02
Fixed Race Condition in ingestor/verifier Test

neilnaveen:
# Description of the PR

- Fixed the race condition bug in test
- For https://github.com/guacsec/guac/issues/1336

# PR Checklist

### DIFF
--- a/pkg/ingestor/verifier/verifier_test.go
+++ b/pkg/ingestor/verifier/verifier_test.go
@@ -20,6 +20,7 @@ import (
 	"encoding/base64"
 	"reflect"
 	"strings"
+	"sync"
 	"testing"
 
 	jsoniter "github.com/json-iterator/go"
@@ -33,6 +34,7 @@ import (
 
 var (
 	json = jsoniter.ConfigCompatibleWithStandardLibrary
+	once sync.Once
 	// Taken from: https://slsa.dev/provenance/v0.1#example
 	ite6SLSA = `
 	{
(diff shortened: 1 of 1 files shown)


jeffmendoza (approved):

pxp928 (approved):

pkg/ingestor/verifier/verifier_test.go
```diff
@@ -127,7 +129,10 @@ func (m *mockSigstoreVerifier) Type() VerifierType {
 
 func TestVerifyIdentity(t *testing.T) {
 	ctx := logging.WithLogger(context.Background())
-	err := RegisterVerifier(newMockSigstoreVerifier(), "sigstore")
+	var err error
+	once.Do(func() {
+		err = RegisterVerifier(newMockSigstoreVerifier(), "sigstore")
+	})
```
lumjjb, line 135: synchronization is not the the issue. i don't think this is the right way to solve the issue here see my comment in https://github.com/guacsec/guac/issues/1336

lumjjb: i see in this case its telling it to do it once, but I think it'll be nicer to have it be proper startup/teardown. Because it doesn't quite make sense in terms of the test itself

neilnaveen: @lumjjb How would we handle this? What would be the setup and teardown process for this? Please let me know.

PR #1347, opened 2023-10-03
SPDX Parser: ingest CPE from externalRefs

### LINKED ISSUE #1346: [ingestion/data-quality issue] SPDX/SBOM externalRefs with cpe
**Describe the bug**

SPDX SBOM can have in the `externalRefs` section, beside the currently managed `PACKAGE-MANAGER`/`purl` information, also `SECURITY`/`CPE` identifiers (ref. [Annex F: External Repository Identifiers - Security ](https://spdx.github.io/spdx-spec/v2.3/external-repository-identifiers/#f2-security))

The CPE information is used in CSAF2.0/VEX files for referencing packages leveraging the `product_identification_helper` field (ref. [Full Product Name Type - Product Identification Helper - CPE](https://docs.oasis-open.org/csaf/csaf/v2.0/os/csaf-v2.0-os.html#31331-full-product-name-type---product-identification-helper---cpe))

Hence the CPE information is key to correlate some SPDX/SBOM files with CSAF/VEX files but CPEs information is currently discarded during SPDX/SBOM parsing.

**To Reproduce**

Ingesting `alpine-small-spdx.json` test data discards `externalRefs` CPE information, i.e. https://github.com/guacsec/guac/blob/67371d73d815e57973e983e532d27173e5599558/internal/testing/testdata/exampledata/alpine-small-spdx.json#L23-L32

**Expected behavior**
Having CPE information ingested just like the `purl` is.

**GUAC version**
`main` branch

**Ingested document(s)**

[`alpine-small-spdx.json`](https://github.com/guacsec/guac/blob/67371d73d815e57973e983e532d27173e5599558/internal/testing/testdata/exampledata/alpine-small-spdx.json)

**Additional context**

mrizzi:
# Description of the PR

### 2023-10-09 update

After the discussion on how to best manage the CPE information, the new `HasMetadata` based approach has been pushed.

### Initial proposal

The changes in this PR are meant to read the `referenceLocator` values for the `externalRefs` with `referenceCategory: SECURITY` and `referenceType` equals `cpe22Type` or `cpe23Type`.

In order to properly sustain the correlation's requirement,  CPE found is translated into a `purl` using an approach similar to the `GuacPkgPurl` func generated `purl`.

The CPE-based purl will be created following this algorithm:

- `cpe` as the hardcoded purl type
- cpe vendor as purl namespace
- cpe product as purl name
- cpe version as purl version
- cpe part as purl qualifier `cpe-part`
- cpe update as purl qualifier `cpe-update`
- cpe edition as purl qualifier `cpe-edition`
- cpe lang as purl qualifier `cpe-lang`
- cpe sw_edition as purl qualifier `cpe-sw-edition`
- cpe target_sw as purl qualifier `cpe-target-sw`
- cpe target_hw as purl qualifier `cpe-target-hw`
- cpe other as purl qualifier `cpe-other`

So, for example, the CPE
```
cpe:/o:redhat:enterprise_linux:7::server
```
will generate the purl
```
pkg:cpe/redhat/enterprise_linux@7?cpe-part=o&cpe-update=ANY&cpe-edition=server&cpe-lang=ANY&cpe-other=ANY&cpe-sw-edition=ANY&cpe-target-hw=ANY&cpe-target-sw=ANY
```

The SPDX parser also generates `PkgEqual` predicate between the current `purl` package and the CPE-based purl in order to have into guac also this information.

~A follow-up PR must be opened to change consistently https://github.com/guacsec/guac-docs/blob/main/guac-graphql.md~

Fixes #1346 

# PR Checklist


internal/testing/testdata/testdata.go
```diff
@@ -700,6 +700,420 @@ var (
 				Justification:  isDepJustifyContainedByJustification,
 			},
 		},
+		{
+			Pkg: &model.PkgInputSpec{
```
pxp928, line 704: is there a reason to duplicate this information if it is already being captured by the original purl? Meaning I understand the need for the `pkgEqual` but seems like it's redundant information to do `isDependency`. We could traverse `pkgEqual` to reach a similar conclusion for the "CPE Purls". Let me know if I am misunderstanding this.

pxp928 (commented):
Thanks @mrizzi this makes sense. Questions about if we need the redundant verbs for the CPEs

internal/testing/testdata/testdata.go
mrizzi: You're right, this information is duplicated because the same information could be retrieved using the `isDependency` + `pkgEqual` graph traversal you mentioned.

Let's say, it would be "shortcut" in the graph from CPE purl to dependency but I'm happy to get rid of it to avoid duplication and "normalize" the underlying graph.

mrizzi: @pxp928 sorry, I've checked the code since I didn't touch the `IsDependencyIngest` management in this PR.

It looks to me the way it works, it's the way that currently guac expects it to work, i.e. in
https://github.com/guacsec/guac/blob/67371d73d815e57973e983e532d27173e5599558/pkg/ingestor/parser/spdx/parse_spdx.go#L270-L286

`getPackageElement` is meant to retrieve an array of `PkgInputSpec` (not just one) and then iterates over `foundPackNodes`.

Do you think `PkgInputSpec` from CPEs should be filtered out in `getPackageElement` func?

pkg/assembler/helpers/purl.go
```diff
@@ -23,11 +23,13 @@ import (
 	"strings"
 
 	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	cpecommon "github.com/knqyf263/go-cpe/common"
 	purl "github.com/package-url/packageurl-go"
 )
 
 const (
 	PurlTypeGuac  = "guac"
+	PurlTypeCPE   = "cpe"
```
lumjjb, line 32: I'm hesitant to have CPE as a pkg trie since it is a bit more like a regex expression than a descriptor of a package, my initial thought is to perhaps have this as a `MetadataFor` package node

lumjjb: @pxp928 @mihaimaruseac @mlieberman85 @jeffmendoza thoughts?

pxp928: Having it as a `MetadataFor` would reduce the overall redundant information and make traversal for information easier

mrizzi: > Having it as a `MetadataFor` would reduce the overall redundant information and make traversal for information easier

Does `MetadataFor` mean `HasMetadata`?
Or are suggesting to add a new `MetadataFor` node?

pxp928: sorry yes [HasMetadata](https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/metadata.graphql#L36)

mrizzi: > sorry yes [HasMetadata](https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/metadata.graphql#L36)

Cool, thanks for confirming :+1:

mihaimaruseac: I also think it makes sense to do this via `HasMetadata`. If we discover that this is not easily searchable and things would improve when using CPE trie, we can switch and take in the cost of duplicating the information

mrizzi: Sorry guys, regarding this duplication topic, let me try to summarize the 2 approaches.

"CPE Package" approach in this PR:

- ingests 1 package (CPE based) and 2 edges (`PkgEqual`, `IsDependency`)
- retrieves the relation between CPE based pkg and dependency pkg with 1 `IsDependency` request

"HasMetadata" approach:

- ingests 1 Metadata (CPE based) and 1 edge (`HasMetadata`)
- retrieves the relation between CPE "metadated" pkg and dependency pkg with 2 requests: `HasMetadata` and `IsDependency`

So, hoping I didn't miss anything, both approaches add 1 Node but the former add 1 more edge and the former requires one more traversal to retrieve the same information.

TBH, I feel like adding one more edge is a better approach rather than having to execute 2 traversals for getting the same information.

WDYT?

pxp928: The main difference is there would be no need to create the pkg trie for the CPE (as they will be heuristic-based). The `HasMetadata` would just be one node attached to the package

mrizzi:
@pxp928 @lumjjb @mihaimaruseac the `HasMetadata` approach has been implemented

pxp928 (approved):
Awesome! Thanks

lumjjb (approved):
LGTM!

Issue #1350, opened 2023-10-04
Use package `slices` instead of  `golang.org/x/exp/slices`


mdeicas:
The standard library package `slices`, added in go 1.21, should be used instead of `golang.org/x/exp/slices`. 

**Describe the solution you'd like**
`slices` uses comparison functions that return integers, while (the version used of) `golang.org/x/exp/slices` uses comparison functions that return booleans. The comparison functions in roughly [these](https://github.com/search?q=repo%3Aguacsec%2Fguac+%2Fslices.*Func%2F&type=code) places would need to be updated. 

**Describe alternatives you've considered**
Keep using `golang.org/x/exp/slices`. This is not ideal because the `exp` module is technically experimental, and the latest version of this package has breaking changes.

PR #1352, opened 2023-10-04
Bump github.com/nats-io/nats-server

jeffmendoza:
This also bumped golang.org/x/exp which included some breaking changes. We will move off golang.org/x/exp/slices onto the stdlib one soon.

# Description of the PR

# PR Checklist

- [ ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [ ] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #1356, opened 2023-10-05
Update packages for slices import

naveensrinivasan:
# Description of the PR
- Replace `golang.org/x/exp/slices` with `slices` import
- Fix https://github.com/guacsec/guac/issues/1350

# PR Checklist

### DIFF
--- a/cmd/guacgql/cmd/server.go
+++ b/cmd/guacgql/cmd/server.go
@@ -21,16 +21,13 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"slices"
 	"syscall"
 	"time"
 
 	"github.com/99designs/gqlgen/graphql/handler"
 	"github.com/99designs/gqlgen/graphql/handler/debug"
 	"github.com/99designs/gqlgen/graphql/playground"
-	"github.com/spf13/cobra"
-	"golang.org/x/exp/maps"
-	"golang.org/x/exp/slices"
-
 	"github.com/guacsec/guac/pkg/assembler/backends"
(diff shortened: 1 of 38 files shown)


mihaimaruseac (changes requested):
This is just syntactic replace, but does not take into account that the API of the two packages is different.

See the following from the linked issue:

>  `slices` uses comparison functions that return integers, while (the version used of) `golang.org/x/exp/slices` uses comparison functions that return booleans. The comparison functions in roughly [these](https://github.com/search?q=repo%3Aguacsec%2Fguac+%2Fslices.*Func%2F&type=code) places would need to be updated.

naveensrinivasan:
> This is just syntactic replace, but does not take into account that the API of the two packages is different.
> 
> See the following from the linked issue:
> 
> > `slices` uses comparison functions that return integers, while (the version used of) `golang.org/x/exp/slices` uses comparison functions that return booleans. The comparison functions in roughly [these](https://github.com/search?q=repo%3Aguacsec%2Fguac+%2Fslices.*Func%2F&type=code) places would need to be updated.

I agree. Thanks.

If I am not wrong, only these must be addressed `sort.SliceStable`.  

Let me now.

jeffmendoza:
I changed a bunch of the compare funcs from `bool` to `int` in #1352

mihaimaruseac (approved):
In that case, approving, as I see the CI passed. But it doesn't harm to do an actual check and make sure the new API works as intended.

jeffmendoza (approved):

Issue #966, opened 2023-06-22
[Discussion] Principles of IsDepedency data model and beyond


lumjjb:
In response to some issues brought up around IsDepedency #594 and #965, the following design doc is written to help address those issues. 

https://docs.google.com/document/d/1A2Fz0TLTsYAsJnTlztCdH05XnMM9llcpVqZtQa5Wt6M/edit

During the use and development of GUAC, one predicate that has been tricky has been IsDepedency. This is due to multiple factors, including the data quality contributing to the predicates, the slightly differing intent and meaning of the predicates (declared build dependencies vs statically linked vs dynamically linked dependencies), and general variance in behavior over ecosystems.

This document talks about how to reason about such issues using IsDependency as an example, and then propose some potential solutions to the IsDepedency problem.

lumjjb:
@knrc

knrc:
@lumjjb Thanks very much, I'll take a look at this later today

knrc:
@lumjjb It looks as if this could address the issue, do you want me to give it a try?  My concern would be the increased verbosity when ingesting.

pxp928:
@knrc With the separation of "noun" and "verb" ingestion and batch ingestion of "nouns" and "verbs", we should be ok in terms of ingesting.

knrc:
@pxp928 I haven't looked at the code in a couple of weeks, but from what I remember there is a lot of duplication in what is sent to the server.  I'll update and see what I've missed

pxp928:
Duplication is still an issue to resolve but in terms of ingestion, we have been trying to make improvements as we work with a graph database backend

knrc:
@pxp928 okay, in that case it will still be an issue and this would add much more data to what is ingested

pxp928:
How so? You would just be adding links to the existing `isDependencies`, `isOccurrences` and `packages or artifacts`

knrc:
@pxp928 true, but that is resolved on the backend so the information still has to be transmitted.

lumjjb:
We went ahead with the solution from "Efficient retrieval with HasSBOM evidence tree edges", implemented in https://github.com/guacsec/guac/pull/1367. 

[...]

Issue #1261, opened 2023-09-12
[feature] Adding optional IDs as part of mutation API


lumjjb:
From the discussion in [slack](https://openssf.slack.com/archives/C03U677QD46/p1694201301936279) around #594 , we discussed what the interface should be in creating links between multiple evidence tree nodes.

This brought up the question of how node inputs are done today, for example in [IsDependency](https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/isDependency.graphql#L86-L87), PkgInputSpec is used to determine which pkg and depPkg to link. 

This is usually implemented in the backend through making a look up to the node, and getting its ID to form the edge. Lookups during ingestion generally slow down the process (e.g. thus introducing the [guacKey](https://github.com/guacsec/guac/blob/349527bff4c62a44d3a0c2ab9709e7437142c73b/pkg/assembler/backends/arangodb/pkg.go#L279)). 

This proposal is to add an optional ID to each InputSpec so that, if provided the backend can use it to enable quick retrieval of the nodes that it needs to form edges around. 

```
"Adds a dependency between two packages. The returned ID can be empty string."
  ingestDependency(
    pkg: PkgInputSpec!
    depPkg: PkgInputSpec!
    depPkgMatchType: MatchFlags!
    dependency: IsDependencyInputSpec!
  ): ID!
```

Would use the IDs provided in `pkg` and `depPkg` in:
```
"""
PkgInputSpec specifies a package for mutations.

This is different than PkgSpec because we want to encode mandatory fields:
type and name. All optional fields are given empty default values.
"""
input PkgInputSpec {
  ID: String               //Added optional field
  type: String!
  namespace: String = ""
  name: String!
  version: String = ""
  qualifiers: [PackageQualifierInputSpec!] = []
  subpath: String = ""
}
```

pxp928:
Based on the discussion in slack, the `pkgInputSpec` (and the other software tries: source and vulnerability) would need to contain all the IDs (for pkgType, pkgNamespace...etc) and not just the pkgVersion ID that the current `pkgInputSpec` returns.

So it would have to be:

```
input PkgInputSpec {
  typeID: ID               //Added optional field
  type: String!
  nameSpaceID: ID               //Added optional field
  namespace: String = ""
  nameID: ID               //Added optional field
  name: String!
  versionID: ID               //Added optional field
  version: String = ""
  qualifiers: [PackageQualifierInputSpec!] = []
  subpath: String = ""
}
```
[...]

PR #1367, opened 2023-10-09
Issue 966: Extend HasSBOM to include references to included software …

knrc:
…(Package and Artifact), dependencies and occurrences

# Description of the PR

Extends HasSBOM to include related packages/artifacts, dependencies and occurrences.  Requires IDs to be sent for included nodes.  Addresses #966, but still needs work done on backends other than `inmem`.

Consider this a work in progress for now

# PR Checklist

- [X] If GraphQL schema is changed, `make generate` has been run
- [X] All CI checks are passing (tests and formatting)


knrc:
Note this PR needs discussion and further work so should not be merged at present

knrc:
One question that needs answering relates to the edges used to identify `Neighbors`, as these are currently unidirectional (sbom -> PackageVersion/Artifact, IsDependency, IsOccurrence).  Do we want those to be bidirectional?  If so, how should we handle PackageVersion -> sbom?

mrizzi:
Cherry-pick'd https://github.com/guacsec/guac/pull/1367/commits/890fe58057796046c2b56f2ee4e03de16ac1cdf7 because other PRs failing and it would fix them without having to wait for this PR to get merged.

pkg/assembler/graphql/schema/package.graphql
```diff
@@ -181,8 +181,8 @@ extend type Query {
 }
 
 extend type Mutation {
-  "Ingests a new package and returns the corresponding package trie path. The returned ID can be empty string."
-  ingestPackage(pkg: PkgInputSpec!): ID!
-  "Bulk ingests packages and returns the list of corresponding package trie path. The returned array of IDs can be a an array of empty string."
-  ingestPackages(pkgs: [PkgInputSpec!]!): [ID!]!
+  "Ingests a new package and returns a corresponding package hierarchy containing only the IDs. The returned ID can be empty string."
+  ingestPackage(pkg: PkgInputSpec!): Package!
```
pxp928, line 185: based on the discussion in the slack, we should change this to return:
```
struct PackageIDs {
  packageTypeID: ID!
  packageNamespaceID: ID!
  packageNameID: ID!
  packageVersionID: ID!
}
```
so we are not capturing other fields unnecessarily during the ingestion. We only care about the IDs.

knrc: Ah good catch, I had intended to go back and tidy this up.  We are only returning the IDs at present, albeit in the same hierarchy as Package.

Do you want this flattened or should we keep the hierarchy?  I was thinking the hierarchy could possibly be useful for merging responses at some future time.

pxp928: so we had that discussion here in Slack: https://openssf.slack.com/archives/C03U677QD46/p1697044972489989?thread_ts=1696877458.535239&cid=C03U677QD46. Based on that, we decided to go the flattened route as its simpler to implement.

knrc: Strange, I didn't get notifications of that part of the discussion so had missed it.  I'll rework it

knrc:
@pxp928 updated to use a flattened structure.  There's also an additional commit for one e2e tests, it seemed to be non-deterministic in that the versions being diffed could be in different orders.

pkg/assembler/backends/inmem/hasSBOM.go
```diff
@@ -192,6 +258,29 @@ func (c *demoClient) convHasSBOM(in *hasSBOMStruct) (*model.HasSbom, error) {
 		}
 		out.Subject = c.convArtifact(art)
 	}
+	// TODO(knrc) - do we want to return full objects for the following?
```
pxp928, line 261: Based on how `model.HasSBOM` is defined:

```
type HasSBOM {
  id: ID!
  "SBOM subject"
  subject: PackageOrArtifact!
  "Identifier for the SBOM document"
  uri: String!
  "Algorithm by which SBOMs digest was computed"
  algorithm: String!
  "Digest of SBOM"
  digest: String!
  "Location from which the SBOM can be downloaded"
  downloadLocation: String!
  "Document from which this attestation is generated from"
  origin: String!
  collector: String!
  "Timestamp for SBOM creation"
  knownSince: Time!
  "Included packages and artifacts"
  includedSoftware: [PackageOrArtifact!]!
  "Included dependencies"
  includedDependencies: [IsDependency!]!
  "Included occurrences"
  includedOccurrences: [IsOccurrence!]!
}
```
we will have to return the objects

pkg/assembler/backends/inmem/hasSBOM.go
```diff
@@ -273,6 +362,8 @@ func (c *demoClient) addHasSBOMIfMatch(out []*model.HasSbom,
 	filter *model.HasSBOMSpec, link *hasSBOMStruct) (
 	[]*model.HasSbom, error) {
 
+	// TODO(knrc) - Should we include the included software, dependencies and occurrences?
+	// Need a better definition of what is required for this.
```
pxp928, line 366: we could do something similar to `hasSLSA`, which also contains an array of artifacts (`builtFrom`): https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/inmem/hasSLSA.go#L328

pkg/assembler/graphql/resolvers/package.resolvers.go
```diff
@@ -6,29 +6,39 @@ package resolvers
...
 
+	"github.com/guacsec/guac/pkg/assembler/graphql/helpers"
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 )
 
 // IngestPackage is the resolver for the ingestPackage field.
-func (r *mutationResolver) IngestPackage(ctx context.Context, pkg model.PkgInputSpec) (string, error) {
+func (r *mutationResolver) IngestPackage(ctx context.Context, pkg model.PkgInputSpec) (*model.PackageIDs, error) {
+	// Return the id of the PackageVersion which has been ingested
+	// TODO (knrc) - check why this was originally returning the ID for the Package, since that is not specific enough
```
pxp928, line 18: Returning only IDs was changed because of this issue: https://github.com/guacsec/guac/issues/1116

pkg/assembler/graphql/helpers/package.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/guacsec/guac/pkg/assembler/graphql/model"
+)
+
+// Return a package structure containing only IDs
+func GetPackageAsIds(packages []*model.Package) []*model.PackageIDs {
+	results := []*model.PackageIDs{}
+	for _, pkg := range packages {
+		resultPackage := model.PackageIDs{PackageTypeID: pkg.ID}
+		for _, namespace := range pkg.Namespaces {
+			resultNamespace := resultPackage
```
pxp928, line 28: nit: why create new variables here? Just keep using `resultPackage`

pxp928 (commented):
Overall this looks great! Thanks @knrc for all the work on this!

pkg/assembler/backends/inmem/hasSBOM.go
knrc: Thanks, will take a look

pkg/assembler/graphql/resolvers/package.resolvers.go
knrc: +1, I was wondering specifically why the ID of the outer Package was returned.  I couldn't think of any useful purpose for that ID but wasn't sure if I was missing something.

pkg/assembler/graphql/helpers/package.go
knrc: I was writing a generic function to flatten the tree structure, hence the new variables to handle multiple namespaces, names,  versions.  I'll add a test case to highlight/validate the behaviour, this may make it more obvious.

pkg/assembler/graphql/resolvers/package.resolvers.go
pxp928: yeah, that was a mistake that we did not catch. Originally it should have been the pkg version ID.

pkg/assembler/graphql/helpers/package.go
pxp928: ah ok

knrc: Test case has been added and pushed up

pkg/assembler/backends/inmem/hasSBOM.go
knrc: This has been updated and pushed

mihaimaruseac (approved):

pkg/assembler/backends/inmem/hasSBOM.go
knrc: This is now included

pkg/assembler/graphql/resolvers/package.resolvers.go
knrc: Note in the latest commit I've updated the Source ingestion to return multiple IDs, as with package ingestion

knrc:
@pxp928 @mihaimaruseac I've pushed the latest version, now also including the sbom includes in the filtering.  I have also updated the source ingestion so it returns all IDs, rather than just the top level, non-specific ID.

I believe this is all the required functionality, let me know what you think and I can squash and remove the WIP tag if you are happy

pxp928 (commented):
This looks good to me! Thanks for all the work on this!

mihaimaruseac (approved):
Looks great! Thank you

knrc:
Thanks very much.  I'll squash the PR into two commits, the first for these changes and the second for the small e2e test that's also in this PR.

pkg/assembler/backends/arangodb/hasSBOM.go
```diff
@@ -401,7 +402,8 @@ func (c *arangoClient) IngestHasSBOMs(ctx context.Context, subjects model.Packag
 	}
 }
 
-func (c *arangoClient) IngestHasSbom(ctx context.Context, subject model.PackageOrArtifactInput, hasSbom model.HasSBOMInputSpec) (*model.HasSbom, error) {
+func (c *arangoClient) IngestHasSbom(ctx context.Context, subject model.PackageOrArtifactInput, hasSbom model.HasSBOMInputSpec, includes model.HasSBOMIncludesInputSpec) (*model.HasSbom, error) {
+	// TODO(knrc) - handle includes
```
pxp928, line 406: do we want to create issues to make these changes on the ENT and arango backends?

knrc: I can add those one this is merged

pkg/assembler/backends/inmem/hasSBOM.go
```diff
@@ -82,11 +98,57 @@ func (c *demoClient) IngestHasSBOMs(ctx context.Context, subjects model.PackageO
...
+	var err error
+
+	if includes.Software != nil {
+		if softwareIDs, err = parseIDs(includes.Software); err != nil {
+			return nil, gqlerror.Errorf("%v ::  %s", funcName, err)
+		}
+		for _, id := range softwareIDs {
+			node, ok := c.index[id]
+			if ok {
+				if _, ok = node.(*pkgVersionNode); !ok {
```
lumjjb, line 116: is there an alternative to this where it returns just the bool, the ok logic is a bit hard to read

pkg/assembler/backends/inmem/hasSBOM.go
```diff
@@ -82,11 +98,57 @@ func (c *demoClient) IngestHasSBOMs(ctx context.Context, subjects model.PackageO
...
+					_, ok = node.(*artStruct)
+				}
+			}
+			if !ok {
+				return nil, gqlerror.Errorf("%v :: software id %d is neither an ingested Package nor an ingested Artifact", funcName, id)
```
lumjjb, line 121: error message is not quite right always , it could be the first !ok case where `node, ok := c.index[id]` is not OK

pkg/assembler/backends/inmem/hasSBOM.go
```diff
@@ -82,11 +98,57 @@ func (c *demoClient) IngestHasSBOMs(ctx context.Context, subjects model.PackageO
...
+			}
+		}
+		softwareIDs = sortAndRemoveDups(softwareIDs)
+	}
+	if includes.Dependencies != nil {
+		if dependencyIDs, err = parseIDs(includes.Dependencies); err != nil {
+			return nil, gqlerror.Errorf("%v ::  %s", funcName, err)
+		}
+		for _, id := range dependencyIDs {
+			if isDependency, err := byID[*isDependencyLink](id, c); isDependency == nil || err != nil {
```
lumjjb, line 131: hmm qq: why the extra check for `isDependency==nil` seems like this shouldn't be the case? if so there's probably a bug somewhere else?

pkg/assembler/backends/inmem/isDependency.go
```diff
@@ -322,3 +310,55 @@ func (c *demoClient) addDepIfMatch(out []*model.IsDependency,
...
+			if filter == nil {
+				continue
+			}
+			if filter.ID != nil {
+				// Check by ID if present
+				if !c.isIDPresent(*filter.ID, depLinkIDs) {
+					return false
+				}
+			} else {
+				// Otherwwise match spec information
```
lumjjb, line 347: nit: typo

lumjjb (approved):
Wanted to recall what the decision was around having the ID list vs having the ID be part of the DependencyInputSpec, OccurrenceInputSpec, etc. , else lgtm

pkg/assembler/graphql/schema/hasSBOM.graphql
```diff
@@ -55,9 +61,18 @@ input HasSBOMSpec {
...
   collector: String
   knownSince: Time
+  includedSoftware: [PackageOrArtifactSpec]!
+  includedDependencies: [IsDependencySpec]!
+  includedOccurrences: [IsOccurrenceSpec]!
 }
 
-"HasSBOMInputSpec is the same as HasSBOM but for mutation input."
+input HasSBOMIncludesInputSpec {
+  software: [ID!]!
```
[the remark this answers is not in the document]
pxp928: this is related to the issue: https://github.com/guacsec/guac/issues/1261. Based on the discussion in slack, the `pkgInputSpec` would need to contain all the IDs (for pkgType, pkgNamespace...etc) and not just the pkgVersion ID that the current `pkgInputSpec` returns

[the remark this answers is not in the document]
pxp928: Once #1261 is resolved, this could be modified to be:
```
input HasSBOMIncludesInputSpec {
  software: [PackageOrArtifactInputSpec!]!
  dependencies: [IsDependencyInputSpec!]!
  occurrences: [IsOccurrenceInputSpec!]!
}
```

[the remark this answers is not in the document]
pxp928: where only the IDs are populated for the various `InputSpec`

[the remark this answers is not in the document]
knrc: I believe this has been an orthogonal discussion, but changing to the input spec would force this to either search via filtering or require only the ID to be present.  I did have another version of this code which worked solely on the input spec without the ID, however that was more involved than this one.

pkg/assembler/backends/inmem/hasSBOM.go
knrc: I'll change the first to raise a different error

pkg/assembler/backends/inmem/hasSBOM.go
knrc: +1, I can remove it

pkg/assembler/backends/inmem/isDependency.go
knrc: will fix :)

pkg/assembler/backends/inmem/hasSBOM.go
knrc: I'll extract this logic into a function

knrc: done

pkg/assembler/backends/inmem/hasSBOM.go
knrc: done

pkg/assembler/backends/inmem/hasSBOM.go
knrc: done

pkg/assembler/backends/inmem/isDependency.go
knrc: done

knrc:
Looks like I need to rebase again, I'll update to the latest main branch and push

pkg/assembler/graphql/schema/hasSBOM.graphql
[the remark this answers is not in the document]
pxp928: yeah, it would require the ID to be present only. Not search via filtering in this case.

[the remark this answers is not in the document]
pxp928: I started work on issue #1261 and can update this as part of that.

PR #1263, opened 2023-09-12
Performance improvements for depsdev API

naveensrinivasan:
# Description of the PR

- Improved the performance of depsdev API fetching using concurrent calls.

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [x] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #1382, opened 2023-10-10
add flag to toggle getting deps.dev dependencies

### LINKED ISSUE #1359: Add a flag to disable pulling in dependencies from deps.dev
This flag will reduce the number of dependency false positives in Guac instances. For the full motivation see #1357.

Some notes: 
- The flag should probably be added to guaccollect. 
- The deps.dev collector should still collect source information (scorecards, license, security advisories). 
- Should the flag enable or disable this behavior?
- A related issue is #1358, which aims to allow labeling certain packages as suitable for ingesting deps.dev dependencies. Should those labels override this flag?

mdeicas:
# Description of the PR

Adds the flag `retrieve-deps` to guaccollect. When set to false, the deps.dev collector only queries for metadata (scorecard and source) and not for dependencies. The default setting is true. 

Also, a log message (level info) was added to log when the dependencies for a package are retrieved. 

Fixes #1359. 

## Behavior 
After starting Guac and ingesting an SBOM, running `guaccollect deps_dev --retrieve-dependencies=false` does not lead to any new IsDependency nodes appearing and results in logs look like 

> {"level":"info","ts":1696975519.9959376,"caller":"deps_dev/deps_dev.go:197","msg":"obtained additional metadata for package: pkg:golang/4d63.com/gocheckcompilerdirectives@v1.2.1"}
{"level":"info","ts":1696975520.0867465,"caller":"deps_dev/deps_dev.go:197","msg":"obtained additional metadata for package: pkg:golang/4d63.com/gochecknoglobals@v0.2.1"}
{"level":"info","ts":1696975520.166465,"caller":"deps_dev/deps_dev.go:197","msg":"obtained additional metadata for package: pkg:golang/github.com/4meepo/tagalign@v1.3.2"}
{"level":"info","ts":1696975520.2557125,"caller":"deps_dev/deps_dev.go:197","msg":"obtained additional metadata for package: pkg:golang/github.com/Abirdcfly/dupword@v0.0.12"}
...

Running  `guaccollect deps_dev` instead results in the normal expected behavior: dependencies are ingested and the logs indicate both metadata and dependency retrievals. 

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/root.go
+++ b/cmd/guaccollect/cmd/root.go
@@ -30,7 +30,7 @@ import (
 func init() {
 	cobra.OnInitialize(cli.InitConfig)
 
-	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll", "retrieve-dependencies"})
+	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll"})
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to setup flag: %v", err)
 		os.Exit(1)
--- a/cmd/guaccollect/cmd/root.go
+++ b/cmd/guaccollect/cmd/root.go
@@ -30,7 +30,7 @@ import (
 func init() {
 	cobra.OnInitialize(cli.InitConfig)
 
-	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll"})
+	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll", "retrieve-dependencies"})
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "failed to setup flag: %v", err)
 		os.Exit(1)
--- a/pkg/handler/collector/deps_dev/deps_dev.go
+++ b/pkg/handler/collector/deps_dev/deps_dev.go
@@ -62,18 +62,19 @@ type PackageComponent struct {
 }
 
 type depsCollector struct {
-	collectDataSource datasource.CollectSource
-	client            pb.InsightsClient
-	poll              bool
-	interval          time.Duration
-	checkedPurls      map[string]*PackageComponent
-	ingestedSource    map[string]*model.SourceInputSpec
-	projectInfoMap    map[string]*pb.Project
-	versions          map[string]*pb.Version
-	dependencies      map[string]*pb.Dependencies
+	collectDataSource    datasource.CollectSource
+	client               pb.InsightsClient
+	poll                 bool
+	retrieveDependencies bool
+	interval             time.Duration
+	checkedPurls         map[string]*PackageComponent
+	ingestedSource       map[string]*model.SourceInputSpec
+	projectInfoMap       map[string]*pb.Project
+	versions             map[string]*pb.Version
+	dependencies         map[string]*pb.Dependencies
 }
 
-func NewDepsCollector(ctx context.Context, collectDataSource datasource.CollectSource, poll bool, interval time.Duration) (*depsCollector, error) {
+func NewDepsCollector(ctx context.Context, collectDataSource datasource.CollectSource, poll bool, retrieveDependencies bool, interval time.Duration) (*depsCollector, error) {
 	// Get the system certificates.
 	sysPool, err := x509.SystemCertPool()
 	if err != nil {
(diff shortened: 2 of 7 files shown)


pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -133,6 +135,17 @@ func (d *depsCollector) populatePurls(ctx context.Context, docChannel chan<- *pr
 	if err != nil {
 		return fmt.Errorf("unable to retrieve datasource: %w", err)
 	}
+
+	if !d.retrieveDependencies {
+		for _, purl := range ds.PurlDataSources {
+			err = d.retrieveMetadataWithoutDeps(ctx, purl, docChannel)
```
pxp928, line 141: could this be made into a non-blocking go routine? That would speed up the fetching as they should all be independent of each other.

pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -150,6 +163,58 @@ func (d *depsCollector) populatePurls(ctx context.Context, docChannel chan<- *pr
...
+	if err != nil {
+		logger.Debugf("failed to get additional metadata for package: %s, err: %v", purl, err)
+		return nil
+	}
+	logger.Infof("obtained additional metadata for package: %s", purl)
+
+	d.checkedPurls[purl] = component
+	blob, err := json.Marshal(component)
+	if err != nil {
+		return err
```
pxp928, line 202: nit: `fmt.Errorf`

pxp928 (commented):
Overall looks good! Just a comment to maybe speed up the retrieval process.

pkg/handler/collector/deps_dev/deps_dev.go
mdeicas: Making each query in a separate goroutine, like this

``` 
for _, purl := range ds.PurlDataSources {
	go func(p datasource.Source) {
		d.retrieveMetadataWithoutDeps(ctx, p, docChannel)
	}(purl)
}
```
will cause the call to `populatePurls`, and then `RetrieveArtifacts` (at least in non-poll mode), to return without blocking. I think this breaks the contract of `RetrieveArtifacts` [here](https://github.com/guacsec/guac/blob/e48e534763f7d70750141e2a9d1047b8caea812c/pkg/handler/collector/collector.go#L42), that the function should block until all of the artifacts are collected.

pxp928: create wait groups and block until it is all collected?

pkg/handler/collector/deps_dev/deps_dev.go
mdeicas: I logged this error instead of returning it, in order to avoid creating a channel just for this marshaling error. Is that OK?

pkg/handler/collector/deps_dev/deps_dev.go
mdeicas: Added this, thanks!

cmd/guaccollect/cmd/root.go
```diff
@@ -30,7 +30,7 @@ import (
 func init() {
 	cobra.OnInitialize(cli.InitConfig)
 
-	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll"})
+	set, err := cli.BuildFlags([]string{"nats-addr", "csub-addr", "use-csub", "service-poll", "retrieve-dependencies"})
```
jeffmendoza, line 33: If this is only used in the deps dev command, just add it there and not here.

jeffmendoza (approved):

pkg/handler/collector/deps_dev/deps_dev.go
```diff
@@ -133,6 +135,22 @@ func (d *depsCollector) populatePurls(ctx context.Context, docChannel chan<- *pr
...
 	}
+
+	if !d.retrieveDependencies {
+		var wg sync.WaitGroup
+		wg.Add(len(ds.PurlDataSources))
+
+		for _, purl := range ds.PurlDataSources {
+			go func(p datasource.Source) {
+				defer wg.Done()
+				d.retrieveMetadataWithoutDeps(ctx, p, docChannel)
```
naveensrinivasan, line 146: Calling this `func` concurrently via `go func` will cause race conditions as the state is manipulated by the calling funcs. 

This needs a larger refactor than just calling concurrently.

pkg/handler/collector/deps_dev/deps_dev.go
naveensrinivasan: The nested calls mutate state, and it isn't thread-safe.

mdeicas: Ah true, thanks @naveensrinivasan. Should I revert this change or add synchronization? Making these calls concurrently was much faster, but making the calls sequentially is similar to how dependencies are fetched sequentially in [getAllDependencies](https://github.com/guacsec/guac/blob/51e8fc6212ac88a8d53d30922520690b8fb97a74/pkg/handler/collector/deps_dev/deps_dev.go#L154), for reference.

naveensrinivasan: The previous depsdev PR(https://github.com/guacsec/guac/pull/1263) was essentially to speed up the fetching of blocking calls. 

If we go with sequential calls, things will be slow.

Adding sync is not recommended in most conditions to avoid future deadlocks and race conditions. 

I recommend refactoring with channels, avoiding the global state returning values from these calls, and then mutating the global state.

mdeicas: I changed this to first concurrently retrieve all of the project and version information and once finished, generate the metadata documents with that information. The goroutines don't mutate any shared state, and they work in the same way as in `getAllDependencies()`

pxp928: Thanks @naveensrinivasan and @mdeicas! This looks great now!

naveensrinivasan (commented):
@mdeicas This is cool! Thanks for taking the feedback and reusing the existing code.

pxp928 (approved):

pxp928:
@mdeicas just need a rebase

PR #1395, opened 2023-10-12
Enable Guacone to ingest json.bz2  files 

desmax74:
# Description of the PR
Enable Guacone to ingest compressed json files with bz2 format

Currently Guacone can't ingest compressed json files with bz2 format

`guacone collect files yourfile.json.bz2`

This PR solve this limitation

# PR Checklist


pkg/handler/processor/processor.go
```diff
@@ -76,6 +76,7 @@ const (
 	FormatJSON      FormatType = "JSON"
 	FormatJSONLines FormatType = "JSON_LINES"
 	FormatXML       FormatType = "XML"
+	FormatBzip2     FormatType = "json.bz2"
```
dejanb, line 79: I wouldn't mix file suffix with the existing constants (and it's related to encoding, not format). Maybe introduce a map that will do this transformations

bz2-> Encoding.BZIP
zst -> Encoding.ZSTD

Then a single lookup could do the trick.

desmax74: Sounds good, refactored with this change

dejanb: Thanks. LGTM now

pkg/handler/processor/process/process.go
```diff
@@ -228,14 +237,14 @@ func decodeDocument(ctx context.Context, i *processor.Document) error {
 	case processor.EncodingUnknown:
```
pxp928, line 237: should this case processor.EncodingUnknown:` as the document is not processable by GUAC (or log that it is being skipped because it does not understand the encoding?)

desmax74: @pxp928  Before this pr one of the error messages (more than one in the stacktrace) with `guacone collect files yourfile.json.bz2` file was` "msg":"emit error: unable to ingest document: unable to process doc: no document processor registered for type: UNKNOWN, format: UNKNOWN, document: UNKNOWN"`

pxp928: so in that case, if the case is not being used, we can just remove it.

desmax74: > so in that case, if the case is not being used, we can just remove it.

Ok

pxp928 (approved):

mihaimaruseac (approved):

PR #1396, opened 2023-10-12
Certifier OSV: fixed emit func when polling

### LINKED ISSUE #1387: [bug] guacone certifier osv -p not working
**Describe the bug**
`guacone certifier osv -p` not working

**To Reproduce**

1. Setup GUAC with https://docs.guac.sh/setup/ or by hand and run `guacone certifier osv -p`
2. See that no vulns are ingested
3. Try `guacone certifier osv` without poll and see that it works

**Expected behavior**
Vulns ingested

**GUAC version**
v0.2.0

**Additional context**

Adding another bug to this: `guacone certifier osv -p` will not exit with `Ctrl-C`, it is hanging.

mrizzi:
# Description of the PR

Fixes #1387 

New Goroutine to "flush" the documents collected in the `totalDocs` based upon one condition being true:

- a 30 seconds ticker
- 1000 documents appended to `totalDocs` 

This also fixes the other bug, i.e. not exiting with `CTRL-C` 

# PR Checklist

### DIFF
--- a/cmd/guacone/cmd/osv.go
+++ b/cmd/guacone/cmd/osv.go
@@ -84,11 +84,55 @@ var osvCmd = &cobra.Command{
 
 		totalNum := 0
 		var totalDocs []*processor.Document
+		docChan := make(chan *processor.Document)
+		tickInterval := 30 * time.Second
+		ticker := time.NewTicker(tickInterval)
 		gotErr := false
+
+		go func() {
+			for !gotErr {
+				select {
+				case <-ticker.C:
+					if len(totalDocs) > 0 {
+						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
+						if err != nil {
+							gotErr = true
+							logger.Errorf("unable to ingest documents: %v", err)
+						}
+						totalDocs = []*processor.Document{}
+					}
+					ticker.Reset(tickInterval)
+				case d := <-docChan:
+					totalNum += 1
+					totalDocs = append(totalDocs, d)
+					if len(totalDocs) >= 1000 {
+						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
+						if err != nil {
+							gotErr = true
+							logger.Errorf("unable to ingest documents: %v", err)
+						}
+						totalDocs = []*processor.Document{}
+						ticker.Reset(tickInterval)
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
+			for len(docChan) > 0 {
+				totalNum += 1
+				totalDocs = append(totalDocs, <-docChan)
+				if len(totalDocs) >= 1000 {
+					err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
+					if err != nil {
+						gotErr = true
+						logger.Errorf("unable to ingest documents: %v", err)
+					}
+				}
+			}
+		}()
 		// Set emit function to go through the entire pipeline
 		emit := func(d *processor.Document) error {
-			totalNum += 1
-			totalDocs = append(totalDocs, d)
+			docChan <- d
 			return nil
 		}
 
@@ -120,15 +164,18 @@ var osvCmd = &cobra.Command{
 		select {
 		case s := <-sigs:
 			logger.Infof("Signal received: %s, shutting down gracefully\n", s.String())
+			cf()
 		case <-done:
 			logger.Infof("All certifiers completed")
 		}
 		wg.Wait()
 
-		err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
-		if err != nil {
-			gotErr = true
-			logger.Errorf("unable to ingest documents: %v", err)
+		if ctx.Err() == nil {
+			err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
+			if err != nil {
+				gotErr = true
+				logger.Errorf("unable to ingest documents: %v", err)
+			}
 		}
 		cf()
 
--- a/cmd/guacone/cmd/osv.go
+++ b/cmd/guacone/cmd/osv.go
@@ -192,7 +192,7 @@ var osvCmd = &cobra.Command{
 		wg.Wait()
 		cf()
 
-		if gotErr == 1 {
+		if atomic.LoadInt32(&gotErr) == 1 {
 			logger.Errorf("completed ingestion with errors")
 		} else {
 			logger.Infof("completed ingesting %v documents", totalNum)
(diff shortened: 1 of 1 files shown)


cmd/guacone/cmd/osv.go
```diff
@@ -84,11 +84,55 @@ var osvCmd = &cobra.Command{
...
+						ticker.Reset(tickInterval)
+					}
+				case <-ctx.Done():
+					return
+				}
+			}
+			for len(docChan) > 0 {
+				totalNum += 1
+				totalDocs = append(totalDocs, <-docChan)
+				if len(totalDocs) >= 1000 {
```
pxp928, line 124: if there are any documents left, we should `MergedIngest` them. To capture any stragglers

mrizzi: good catch, fixed

cmd/guacone/cmd/osv.go
```diff
@@ -84,11 +84,62 @@ var osvCmd = &cobra.Command{
 
 		totalNum := 0
 		var totalDocs []*processor.Document
+		docChan := make(chan *processor.Document)
+		tickInterval := 30 * time.Second
+		ticker := time.NewTicker(tickInterval)
 		gotErr := false
+
+		go func() {
+			for !gotErr {
```
naveensrinivasan, line 93: The `gotErr` is shared and isn't thread-safe.

cmd/guacone/cmd/osv.go
```diff
@@ -84,11 +84,62 @@ var osvCmd = &cobra.Command{
...
+				select {
+				case <-ticker.C:
+					if len(totalDocs) > 0 {
+						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
```
naveensrinivasan, line 97: The `totalDocs` slice could potentially run into a race condition.

The totalDocs is accessed and modified in the goroutine that handles document ingestion. However, it's also accessed and potentially modified after the wg.Wait() call when the context error is checked and possibly when final ingestion is performed.

If the goroutine is still running and modifying totalDocs when the main goroutine reaches the final ingestion step, a race condition could occur. This is because both goroutines could be trying to access and modify totalDocs at the same time.

mrizzi:
2efd048 :

- refactored `totalDocs` slice management that now happens only inside `ingestion` func
- `ingestionStop` chan to properly stop `ingestion` func
- removed last call to `ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)` because now all the ingestor invocations happen only within the  `ingestion` func
-  `ingestion` func deals with `wg WaitGroup`

cmd/guacone/cmd/osv.go
mrizzi: @naveensrinivasan thanks a lot, it helped me a lot.
I've refactored the whole execution and now `totalDocs` slice is referenced only within the `ingestion` func.

pxp928: Thanks @naveensrinivasan that was a good catch.

pxp928 (dismissed):

cmd/guacone/cmd/osv.go
naveensrinivasan: Thanks. 

- A suggestion is to `break` instead of `flag` for breaking the loop.
- The `gotErr` can also run into a race condition.

```golang
		var gotErr int32
		var wg sync.WaitGroup
		ingestion := func() {
			defer wg.Done()
			var totalDocs []*processor.Document
			const threshold = 1000
		loop:
			for {
				select {
				case <-ticker.C:
					if len(totalDocs) > 0 {
						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
						if err != nil {
							logger.Errorf("unable to ingest documents: %v", err)
							atomic.StoreInt32(&gotErr, 1)
							break loop
						}
						totalDocs = []*processor.Document{}
					}
					ticker.Reset(tickInterval)
				case d := <-docChan:
					totalNum += 1
					totalDocs = append(totalDocs, d)
					if len(totalDocs) >= threshold {
						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
						if err != nil {
							atomic.StoreInt32(&gotErr, 1)
							logger.Errorf("unable to ingest documents: %v", err)
							break loop
						}
						totalDocs = []*processor.Document{}
						ticker.Reset(tickInterval)
					}
				case <-ingestionStop:
					break loop
				case <-ctx.Done():
					return
				}
			}
			for len(docChan) > 0 {
				totalNum += 1
				totalDocs = append(totalDocs, <-docChan)
				if len(totalDocs) >= threshold {
					err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
					if err != nil {
						atomic.StoreInt32(&gotErr, 1)
						logger.Errorf("unable to ingest documents: %v", err)
					}
					totalDocs = []*processor.Document{}
				}
			}
			if len(totalDocs) > 0 {
				err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, csubClient)
				if err != nil {
					atomic.StoreInt32(&gotErr, 1)
					logger.Errorf("unable to ingest documents: %v", err)
				}
			}
		}
```

mrizzi: @naveensrinivasan thanks a lot :pray:

cmd/guacone/cmd/osv.go
```diff
@@ -191,7 +192,7 @@ var osvCmd = &cobra.Command{
 		wg.Wait()
 		cf()
 
-		if gotErr {
+		if gotErr == 1 {
```
naveensrinivasan, line 195: This has to use the `atomic` package to compare. To avoid the race. https://pkg.go.dev/sync/atomic#LoadInt32

cmd/guacone/cmd/osv.go
naveensrinivasan: @mrizzi  Appreciate taking the feedback 👍

pxp928: seems like the flag is a better option for breaking to loop instead of a goto statement

mrizzi: :sweat_smile: 
So, I have no strong opinion on this guys, but I'm concerned the certifier OSV polling is broken.

The only rationale to decide how to implement this that I can see is the consistency with the rest of guac. It looks to me that flag approach is widely used.

@naveensrinivasan @pxp928 WDYT?

naveensrinivasan: I am good :)

pxp928: Yeah, let's stick with the flag. Thanks @mrizzi

pxp928 (approved):

jeffmendoza (approved):

PR #1401, opened 2023-10-15
Included version information which the CLI starts up

naveensrinivasan:
# Description of the PR

- Included the version information when the CLI starts up.

```
{"level":"info","ts":1697396471.853624,"caller":"version/version.go:57","msg":"Commit: dev-7c3b1b9188e868c1ce2c8c21793cd7e6894aa244\n"}
{"level":"info","ts":1697396471.853778,"caller":"version/version.go:64","msg":"Date: 2023-10-15"}
```

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/deps_dev.go
+++ b/cmd/guaccollect/cmd/deps_dev.go
@@ -30,6 +30,7 @@ import (
 	"github.com/guacsec/guac/pkg/handler/collector"
 	"github.com/guacsec/guac/pkg/handler/collector/deps_dev"
 	"github.com/guacsec/guac/pkg/logging"
+	"github.com/guacsec/guac/pkg/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
@@ -52,6 +53,8 @@ var depsDevCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := logging.WithLogger(context.Background())
 		logger := logging.FromContext(ctx)
+
+		version.DumpVersion()
 
 		opts, err := validateDepsDevFlags(
 			viper.GetString("nats-addr"),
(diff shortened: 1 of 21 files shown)


cmd/guaccollect/cmd/deps_dev.go
```diff
@@ -53,6 +54,8 @@ var depsDevCmd = &cobra.Command{
 		ctx := logging.WithLogger(context.Background())
 		logger := logging.FromContext(ctx)
 
+		version.DumpVersion()
```
lumjjb, line 57: i dont think we need to dump the version on every invocation. I think CLIs can have a `--version` flag which should output the version,

pkg/version/version.go
```diff
@@ -40,3 +46,23 @@ func (u uat) RoundTrip(r *http.Request) (*http.Response, error) {
 	r.Header.Set("User-Agent", UserAgent)
 	return u.tr.RoundTrip(r)
 }
+
+// DumpVersion dumps the version information to the log.
+func DumpVersion() {
```
lumjjb, line 51: this implementation needs to be robust , as it stands it is not. Plus it depends on a CLI command.. which I don't want us to have a dependency for this trade-off. I think a better more robust solution is needed here. I believe golang has some in built features to do this.

lumjjb (changes requested):

cmd/guaccollect/cmd/deps_dev.go
naveensrinivasan: The goal was to know which version of the CLI runs when the server starts. 

Right now, that isn't possible.  

That was the reason for this change.

Do you have suggestions for that?

pkg/version/version.go
naveensrinivasan: It is using already using builtin https://github.com/guacsec/guac/blob/7c3b1b9188e868c1ce2c8c21793cd7e6894aa244/.goreleaser.yaml#L67-L69

If the built-in fails, we use the commit SHA and the present date and run the non-released version.

Let me know if you have a better strategy to solve this issue.

Issue #1386, opened 2023-10-11
[bug] docker-compose.yaml starts postgres but does not use it.


jeffmendoza:
**Describe the bug**
docker-compose.yaml starts postgres but does not use it.

**To Reproduce**

1. Download [guac-compose.tar.gz](https://github.com/guacsec/guac/releases/download/v0.2.0/guac-compose.tar.gz)
2. compose up

**Expected behavior**
Only start containers needed

**GUAC version**
v0.2.0

**Additional context**

We should have an easy way to compose-up with postgres/arango when those are ready. For now don't start it with the released compose yaml.

ALSO: The image name used is `postgres`, and should be `docker.io/library/postgres` for podman compatability.

desmax74:
I can take this issue.

pxp928:
we can create a follow-on issue for better/ more configurable way to bring up the proper backends once ready for use.

PR #1430, opened 2023-10-24
[bug] docker-compose.yaml starts postgres but does not use it

desmax74:
# Description of the PR
docker-compose.yaml starts postgres but does not use it.
see : https://github.com/guacsec/guac/issues/1386

# PR Checklist

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,17 +15,18 @@ services:
   #     - "7687:7687"
   #   restart: on-failure
 
-  postgres:
-    image: postgres:15
-    environment:
-      POSTGRES_USER: guac
-      POSTGRES_PASSWORD: guac
-      POSTGRES_HOST_AUTH_METHOD: trust
-    networks: [frontend]
-    ports:
-      - "5432:5432"
-    volumes:
-      - ./container_files/pg:/var/lib/postgresql/data
+  # Ent on postgres is turned down for now since we are currently only using the in memory backend
+  # postgres:
+  #  image: docker.io/library/postgres:16
+  #  environment:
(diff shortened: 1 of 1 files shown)


container_files/guac/guac.yaml
```diff
@@ -34,3 +34,4 @@ interval: 5m
 db-driver: postgres
 db-address: postgres://guac:guac@postgres:5432/guac?sslmode=disable
 db-migrate: true
+db-debug: false
```
pxp928, line 37: what is this change for?

pxp928 (approved):
Thanks! Per the issue, we need a better way to bring up the databases based on the backends selected.

container_files/guac/guac.yaml
mihaimaruseac: +1 to this

mihaimaruseac (approved):

lumjjb (approved):

container_files/guac/guac.yaml
desmax74: @pxp928 is a missing param in the db configuration https://github.com/guacsec/guac/blob/main/pkg/cli/store.go#L64

pxp928: hmmm looks like they are not being used. I think we should remove it for the time being if not needed. Same with `db-migrate`?

desmax74: @pxp928 `db-migrate` param is used https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/ent/migrate/schema.go 
I'll remove only the `db-debug`

PR #1453, opened 2023-10-31
fix check for ingested packages and source IDs

pxp928:
# Description of the PR

check input and results for the IDs as arango and inmem return in different formats which causes an unneeded error. This will change in the future to just return IDs but currently causes errors in arango ingestion 

# PR Checklist

### DIFF
--- a/pkg/assembler/graphql/resolvers/package.resolvers.go
+++ b/pkg/assembler/graphql/resolvers/package.resolvers.go
@@ -32,7 +32,7 @@ func (r *mutationResolver) IngestPackages(ctx context.Context, pkgs []*model.Pkg
 	ingestedPackages, err := r.Backend.IngestPackages(ctx, pkgs)
 	if err == nil {
 		results := helpers.GetPackageAsIds(ingestedPackages)
-		if len(results) != len(ingestedPackages) {
+		if len(results) != len(pkgs) {
 			return nil, fmt.Errorf("could no derive correct package ID information for ingested packages, expected to return %d but have %d", len(ingestedPackages), len(results))
 		}
 		return results, nil
--- a/pkg/assembler/graphql/resolvers/package.resolvers.go
+++ b/pkg/assembler/graphql/resolvers/package.resolvers.go
@@ -33,7 +33,7 @@ func (r *mutationResolver) IngestPackages(ctx context.Context, pkgs []*model.Pkg
 	if err == nil {
 		results := helpers.GetPackageAsIds(ingestedPackages)
 		if len(results) != len(pkgs) {
-			return nil, fmt.Errorf("could no derive correct package ID information for ingested packages, expected to return %d but have %d", len(ingestedPackages), len(results))
+			return nil, fmt.Errorf("could no derive correct package ID information for ingested packages, expected to return %d but have %d", len(pkgs), len(results))
 		}
 		return results, nil
 	}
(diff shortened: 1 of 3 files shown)


pkg/assembler/graphql/resolvers/package.resolvers.go
```diff
@@ -32,7 +32,7 @@ func (r *mutationResolver) IngestPackages(ctx context.Context, pkgs []*model.Pkg
 	ingestedPackages, err := r.Backend.IngestPackages(ctx, pkgs)
 	if err == nil {
 		results := helpers.GetPackageAsIds(ingestedPackages)
-		if len(results) != len(ingestedPackages) {
+		if len(results) != len(pkgs) {
 			return nil, fmt.Errorf("could no derive correct package ID information for ingested packages, expected to return %d but have %d", len(ingestedPackages), len(results))
```
mihaimaruseac, line 36: I think the error message also needs to change.

But in general , we were expecting to get the same number of results as the number of `ingestedPackages` (which should be the same as `pkgs`). What did change?

mihaimaruseac (changes requested):

pkg/assembler/graphql/resolvers/package.resolvers.go
pxp928: so arango uses a nested package that groups everything together by type and namespace...etc. inmem does not and lists out individual packages. How arango is returning it seems like the right way to do it but this is the disconnect that is causing the error in the check. The length of `ingestedPackages` would be 2 in arango's case but contain the the proper 14 pkgVersion IDs. The length of `ingestedPackages` would be 14 in inmem's case. Does that make sense?

pxp928: the difference is that Arango uses batching and collecting everything together. inmem is not and returns individual packages

pxp928: the `results` are checking the `ingestedPackages` to get the IDs, so either way, the proper number of ingested packages are being checked whether we check the length of `ingestedPackages` or the pkgs being passed in

mihaimaruseac: But I think that the arango resolver should return the data formatted consistently with the other resolvers. At this point it looks like changing the backend means chaging the cardinality of the result set for the same user query, the backends are no longer transparent

pxp928: in that case I think the `inmem` should be the one that needs to change for this case. As it should be returning the nested package objects back from a batch ingestion

mihaimaruseac: So the issue is that `helpers.GetPackageAsIds` flattens out all trie traversals into a list, but this is not documented, so on a first glance it looks like it should return a collection of same size as the input. But of course this is not true when you have `A->B` and `A->C` in the trie (anywhere): arango would return `[{A, {B, C}]` as 1 single element and then `helpers.GetPackageAsIds` would return `[{A, B}, {A, C}]` which now would match the original input (where we specify that we want to ingest a package with A, B and one with A, C).

Makes sense now, but let's fix error message to also print the number we compare with and maybe also add a comment to `helpers.GetPackageAsIds` to state that it flattens the trie.

mihaimaruseac (approved):
Modulo the change in error message on the line that follows the comparison

pkg/assembler/graphql/resolvers/package.resolvers.go
pxp928: done

mihaimaruseac (approved):
Thank you!

jeffmendoza (approved):

PR #1440, opened 2023-10-30
Ent: Package,IsDependency concurrent bulk ingestions

mrizzi:
# Description of the PR

Ent optimization done for bulk ingestion of Packages and IsDependency for the sake of "approving" the approach and then apply the same for all the remaining bulk ingestion implementations with Ent.

With current `main` branch, the ingestion of an SBOM (mainly about ingesting `Package` and `IsDependency`), from Postresql perspective, generates ~700 transactions/sec:

main_transactions

With this PR applied, the performance arrives at ~7000  transactions/sec:

PR_transactions

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [x] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #1459, opened 2023-11-03
Ent - Vulnerability endpoints: applied concurrent approach

mrizzi:
# Description of the PR

Ent backend, implemented concurrent approach [1] for:

- `IngestVulnerabilities`
- `IngestVEXStatements`
- `IngestCertifyVulns`

[1] from https://github.com/guacsec/guac/pull/1440

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/backend/certifyVEXStatement.go
+++ b/pkg/assembler/backends/ent/backend/certifyVEXStatement.go
@@ -29,6 +29,7 @@ import (
 	"github.com/guacsec/guac/pkg/assembler/graphql/model"
 	"github.com/pkg/errors"
 	"github.com/vektah/gqlparser/v2/gqlerror"
+	"golang.org/x/sync/errgroup"
 )
 
 func (b *EntBackend) IngestVEXStatement(ctx context.Context, subject model.PackageOrArtifactInput, vulnerability model.VulnerabilityInputSpec, vexStatement model.VexStatementInputSpec) (*model.CertifyVEXStatement, error) {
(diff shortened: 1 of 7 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/ent/backend/certifyVEXStatement.go
```diff
@@ -139,19 +140,30 @@ func (b *EntBackend) IngestVEXStatement(ctx context.Context, subject model.Packa
 }
 
 func (b *EntBackend) IngestVEXStatements(ctx context.Context, subjects model.PackageOrArtifactInputs, vulnerabilities []*model.VulnerabilityInputSpec, vexStatements []*model.VexStatementInputSpec) ([]string, error) {
-	var ids []string
+	var ids = make([]string, len(vexStatements))
+	eg, ctx := errgroup.WithContext(ctx)
```
naveensrinivasan, line 144: Suppose the IngestVEXStatement function receives a `ctx` from errgroup.WithContext(ctx), and this context gets canceled when any goroutine fails, could this potentially interrupt important operations inside IngestVEXStatement, such as database writes?


 Would it lead to data inconsistencies? 
 
 Would it be safer to use the original `ctx` for critical tasks in IngestVEXStatement and the derived context only for tasks that should be stopped if any goroutine fails to ensure critical operations aren't prematurely terminated?

mrizzi: The "upsert" approach adopted in the Ent backend is meant to ensure idempotency to mutation endpoints to avoid any kind of data inconsistency or more complicated tasks management.

pxp928 (approved):

pkg/assembler/backends/ent/backend/certifyVEXStatement.go
naveensrinivasan: Thank you! I wasn't aware of this. 👍

PR #1461, opened 2023-11-06
Ent - IngestSLSAID

desmax74:
# Description of the PR

Ent backend

- `IngestSLSAID` (with bulk) implementations
-  `IngestSLSA`, removed
-   changes consistently all of the tests involved

Refers to https://github.com/guacsec/guac/issues/1198

# PR Checklist


pkg/assembler/backends/ent/backend/slsa.go
```diff
@@ -140,15 +142,18 @@ func upsertSLSA(ctx context.Context, client *ent.Tx, subject model.ArtifactInput
 		UpdateNewValues().
```
mrizzi, line 142: This has to be replaced with `DoNothing()`
```suggestion
		DoNothing().
```

mrizzi (changes requested):
`DoNothing()` has to be applied.

pkg/assembler/backends/ent/backend/slsa.go
desmax74: It skip the update and the test fail with this change

desmax74: ok

desmax74:
@mrizzi

pxp928:
We can now make these updates on main as needed. :)

Issue #1406, opened 2023-10-16
[proposal] Add Go interface to inmem backend between logic and storage maps


jeffmendoza:
The inmem backend serves as our gold-standard canonical backend. All features / nodes are implemented there first and we expect the behavior of the inmem backend to be correct. The implementation does querying, filtering, and ingestion all using Go code, and is backed by Go structs in lists or maps. There is no way to dump, save, backup, or restore the data it is simply all "in memory".

This proposal is to decouple all the logic code in the inmem backend from the storage. This would make it possible to use all the canonical query and filter code we wrote with a different storage system, enabling persistant storage among other options. The storage would be a simple Key Value store and all logic would remain in code as is now, no queries or filters would be passed to the storage system.

The existing inmem backend would be replaced with this decoupled implementation and the default storage would be a Go map. Because a generic Key Value store cannot store Go structs, the storage object would be a string or []byte and the existing Go structs would be serialized to Json.

The proposed interface is:
```go
type Store interface {
	Get(ctx context.Context, collection, key string) (string, error)
	Set(ctx context.Context, collection, key, value string) error
	Keys(ctx context.Context, collection string) ([]string, error)
}

var KeyError = errors.New("Invalid Key")

var CollectionError = errors.New("Invalid Collection")
```

The default implementation would be:
```go
type Store struct {
	m map[string]map[string]string
}

func GetStore() kv.Store {
	return &Store{
		m: make(map[string]map[string]string),
	}
}

func (s *Store) Get(_ context.Context, c, k string) (string, error) {
	col, ok := s.m[c]
	if !ok {
		return "", fmt.Errorf("%w : %s", kv.CollectionError, c)
	}

	val, ok := col[k]
	if !ok {
		return "", fmt.Errorf("%w : %s", kv.KeyError, k)
	}
	return val, nil
}

func (s *Store) Set(_ context.Context, c, k, v string) error {
	if s.m[c] == nil {
		s.m[c] = make(map[string]string)
	}
	s.m[c][k] = v
	return nil
}

func (s *Store) Keys(_ context.Context, c string) ([]string, error) {
	if s.m[c] == nil {
		return nil, fmt.Errorf("%w : %s", kv.CollectionError, c)
	}
	return maps.Keys(s.m[c]), nil
}
```

I've done some limited performance testing on my machine, and the impact of the json serialization is less than 10% over the current implementation.

Some alternatives:

[...]

PR #1475, opened 2023-11-07
Convert default backend from "inmem" to "keyvalue"

jeffmendoza:
Add a new interface under `pkg/assember/kv` that is a very simple key value store. Convert the existing logic in the `inmem` backend to remove the embedded Go lists and maps and use the new interface, now known as `keyvalue`. The default (and currently only) implementation is `pkg/assembler/kv/memmap` which is just a Go map.

Future kv implementations may be added to allow a simple persistent storage that uses the canonical backend logic. 

Full proposal: #1406

Reviewers: note I copied the whole inmem backend and then modified that. To see the diff, take a look at the second commit, the first commit includes the copy so the whole file is new/green there. There were a couple updates to inmem made after I made my copy, so those are added in the third and fourth commit.

# PR Checklist


mihaimaruseac (approved):
Sorry for the delay, took a while to get to all the changes

pxp928 (approved):

PR #1480, opened 2023-11-09
update artifact ingestion to return only ID

shafeeshafee:
# Description of the PR

This PR is a small part of a larger effort to optimize some of the backend ingestions, and merging the work to this branch (`1116-return-ID`). The work in this PR focuses on changing artifact ingestion to return ID on in-mem and ArangoDB.

@mrizzi - tagging for vis. Feel free to let me know if everything looks okay.

After feedback on this, I can carry out larger chunks of the work needed for this issue.

# PR Checklist


mihaimaruseac (approved):
This lgtm to me, but please wait for @mrizzi to look over too before merging

pkg/assembler/backends/arangodb/artifact.go
```diff
@@ -206,11 +206,7 @@ func (c *arangoClient) IngestArtifacts(ctx context.Context, artifacts []*model.A
 UPSERT { algorithm:doc.algorithm, digest:doc.digest }
```
pxp928, line 206: There is an optimization we can do in the query to not use upset here as we no longer need to return the artifact. Utilizing the a composite key and using the ignore duplicates key flag.

pxp928: I can explain more but not near the computer right now.

Issue #1370, opened 2023-10-09
[feature] Support Docker BuildKit's Image Attestation Storage


ridhoq:
**Is your feature request related to a problem? Please describe.**
Today, GUAC supports collection from OCI artifacts via fallback artifacts and, soon, OCI referrers (#1277). Another way that supply chain artifacts are stored in OCI registries is [Docker Buildkit's Image Attestation Storage](https://docs.docker.com/build/attestations/attestation-storage/). The high level approach is that attestations like SBOMs and SLSA Provenance are stored as manifests in a manifest list along side the image manifests. The attestations are marked with annotations to identify itself as an annotation. Here is an example:
```json
{
  "mediaType": "application/vnd.oci.image.index.v1+json",
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:23678f31b3b3586c4fb318aecfe64a96a1f0916ba8faf9b2be2abee63fa9e827",
      "size": 1234,
      "platform": {
        "architecture": "amd64",
        "os": "linux"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.manifest.v1+json",
      "digest": "sha256:02cb9aa7600e73fcf41ee9f0f19cc03122b2d8be43d41ce4b21335118f5dd943",
      "size": 1234,
      "annotations": {
        "vnd.docker.reference.digest": "sha256:23678f31b3b3586c4fb318aecfe64a96a1f0916ba8faf9b2be2abee63fa9e827",
        "vnd.docker.reference.type": "attestation-manifest"
      },
      "platform": {
         "architecture": "unknown",
         "os": "unknown"
      }
    }
  ]
}
```

**Describe the solution you'd like**
Add support for this in the OCI collector

PR #1482, opened 2023-11-10
(feat) add support for attestations stored as image manifest

RealHarshThakur:
# Description of the PR
Currently, guac only works with certain OCI images. Build tools store attestations in a variety of ways as of now and it might take them a while to adopt to using referrers. As of now, Docker stores the attestation in an image manifest and it would be wonderful for GUAC to support that as docker is a popular tool for building containers.  I'm not entirely sure if this is the right way to implement this,  I'd appreciate any feedback to make this change a reality. 

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/parser.go
+++ b/pkg/ingestor/parser/parser.go
@@ -44,6 +44,7 @@ func init() {
 	_ = RegisterDocumentParser(dsse.NewDSSEParser, processor.DocumentDSSE)
 	_ = RegisterDocumentParser(slsa.NewSLSAParser, processor.DocumentITE6SLSA)
 	_ = RegisterDocumentParser(vuln.NewVulnCertificationParser, processor.DocumentITE6Vul)
+	_ = RegisterDocumentParser(spdx.NewSpdxParser, processor.DocumentITE6SPDX)
 	_ = RegisterDocumentParser(spdx.NewSpdxParser, processor.DocumentSPDX)
 	_ = RegisterDocumentParser(cyclonedx.NewCycloneDXParser, processor.DocumentCycloneDX)
 	_ = RegisterDocumentParser(scorecard.NewScorecardParser, processor.DocumentScorecard)
--- a/pkg/ingestor/parser/spdx/parse_spdx.go
+++ b/pkg/ingestor/parser/spdx/parse_spdx.go
@@ -30,10 +30,15 @@ import (
 	"github.com/guacsec/guac/pkg/handler/processor"
 	"github.com/guacsec/guac/pkg/ingestor/parser/common"
 	"github.com/guacsec/guac/pkg/logging"
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	jsoniter "github.com/json-iterator/go"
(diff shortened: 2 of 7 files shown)


ridhoq:
Hi @RealHarshThakur thanks for the PR! This is actually something I created an issue on #1370 so it's great to see somebody picking up this work. There is a bit of overlap here with #1449 which I believe will also cover collection of these types of attestations. But, this change also will move away from `regclient` to `google/go-containerregistry` as the library to access OCI packages. The community is discussing this change in #1449. Don't take it as a reason to stop further work on this - just wanted to bring you up to speed on this topic :)

RealHarshThakur:
My bad 😓 I had it on my todo list before the issue was created and just found this weekend to hack on it. 
From skimming the other PR, I think one way to move forward would be for me to refactor this PR to focus on processor and parser. Essentially, all changes to introduce `DocumentITE6SPDX`.

pkg/ingestor/parser/spdx/parse_spdx.go
```diff
@@ -65,10 +77,23 @@ func NewSpdxParser() common.DocumentParser {
 
 func (s *spdxParser) Parse(ctx context.Context, doc *processor.Document) error {
 	s.doc = doc
-	spdxDoc, err := parseSpdxBlob(doc.Blob)
-	if err != nil {
-		return fmt.Errorf("failed to parse SPDX document: %w", err)
+	var spdxDoc *v2_3.Document
+	if doc.Type == processor.DocumentITE6SPDX {
+		var err error
+		itespdx, err := parseITESPDXBlob(doc.Blob)
```
pxp928, line 83: This should be part of the processor such the parser does not need to change as it will just receive a valid SPDX document at the end.

pkg/ingestor/parser/parser.go
```diff
@@ -44,6 +44,7 @@ func init() {
 	_ = RegisterDocumentParser(dsse.NewDSSEParser, processor.DocumentDSSE)
 	_ = RegisterDocumentParser(slsa.NewSLSAParser, processor.DocumentITE6SLSA)
 	_ = RegisterDocumentParser(vuln.NewVulnCertificationParser, processor.DocumentITE6Vul)
+	_ = RegisterDocumentParser(spdx.NewSpdxParser, processor.DocumentITE6SPDX)
```
pxp928, line 47: it can be expanded [here](https://github.com/guacsec/guac/blob/main/pkg/handler/processor/ite6/ite6.go) similar to how its done for [DSSE](https://github.com/guacsec/guac/blob/main/pkg/handler/processor/dsse/dsse.go#L55)

pxp928 (commented):
Thanks for the PR @RealHarshThakur. Added some comments

pxp928:
@RealHarshThakur marking this as not stale as it is an important topic to keep working on.

pxp928 (commented):
@RealHarshThakur marking this as not stale as it is an important topic to keep working on.

RealHarshThakur:
sorry for delay, I'll try to make some time this month

pxp928:
Adding comment to ensure this does not close.

PR #1497, opened 2023-11-14
CSAF Parser: fix double ingestion

### LINKED ISSUE #1496: [ingestion/data-quality issue] CSAF VEX file creates both CertifyVEX and CertifyVuln
**Describe the bug**

Ingesting a CSAF VEX file creates both `CertifyVEXStatement` and `CertifyVuln` which duplicates the "vulnerabilities" affecting a package which looks wrong considering a CSAF VEX should be reflected exclusively in a `CertifyVEXStatement`.

**To Reproduce**

Run [`Test_csafParser` that expects both `Vex` and `CertifyVuln`](https://github.com/guacsec/guac/blob/a3301cb3e6f8d4d014b9edc4fb6ddcb0b84da05c/pkg/ingestor/parser/csaf/parser_csaf_test.go#L54) to have values

**Expected behavior**
Only a `CertifyVEXStatement` should be created

**GUAC version**
`main` branch

**Ingested document(s)**

Refer to the test data mentioned above

mrizzi:
# Description of the PR

Fixes #1496 

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/csaf/parser_csaf.go
+++ b/pkg/ingestor/parser/csaf/parser_csaf.go
@@ -274,7 +274,6 @@ func (c *csafParser) generateVexIngest(ctx context.Context, vulnInput *generated
 func (c *csafParser) GetPredicates(ctx context.Context) *assembler.IngestPredicates {
 	rv := &assembler.IngestPredicates{}
 	var vis []assembler.VexIngest
-	var cvs []assembler.CertifyVulnIngest
 
 	for _, v := range c.csaf.Vulnerabilities {
 		vuln, err := helpers.CreateVulnInput(v.CVE)
@@ -290,36 +289,10 @@ func (c *csafParser) GetPredicates(ctx context.Context) *assembler.IngestPredica
 				if vi == nil {
 					continue
 				}
-
-				if status == "known_affected" || status == "under_investigation" {
-					vulnData := generated.ScanMetadataInput{
-						TimeScanned: c.csaf.Document.Tracking.CurrentReleaseDate,
-					}
-					cv := assembler.CertifyVulnIngest{
-						Pkg:           vi.Pkg,
-						Vulnerability: vuln,
-						VulnData:      &vulnData,
-					}
-					cvs = append(cvs, cv)
-				} else if status == "known_not_affected" || status == "fixed" {
-					vulnData := generated.ScanMetadataInput{
-						TimeScanned: c.csaf.Document.Tracking.CurrentReleaseDate,
-					}
-					noVuln := generated.VulnerabilityInputSpec{
-						Type: "NoVuln",
-					}
-					cv := assembler.CertifyVulnIngest{
-						Pkg:           vi.Pkg,
-						Vulnerability: &noVuln,
-						VulnData:      &vulnData,
-					}
-					cvs = append(cvs, cv)
-				}
 				vis = append(vis, *vi)
 			}
 		}
 	}
 	rv.Vex = vis
-	rv.CertifyVuln = cvs
 	return rv
 }
--- a/pkg/ingestor/parser/csaf/parser_csaf_test.go
+++ b/pkg/ingestor/parser/csaf/parser_csaf_test.go
@@ -50,8 +50,7 @@ func Test_csafParser(t *testing.T) {
 			},
 		},
 		wantPredicates: &assembler.IngestPredicates{
-			Vex:         testdata.CsafVexIngest,
-			CertifyVuln: testdata.CsafCertifyVulnIngest,
+			Vex: testdata.CsafVexIngest,
 		},
 		wantErr: false,
 	}}


mihaimaruseac (approved):

pkg/ingestor/parser/csaf/parser_csaf.go
```diff
@@ -290,36 +289,10 @@ func (c *csafParser) GetPredicates(ctx context.Context) *assembler.IngestPredica
 				if vi == nil {
 					continue
 				}
-
-				if status == "known_affected" || status == "under_investigation" {
-					vulnData := generated.ScanMetadataInput{
-						TimeScanned: c.csaf.Document.Tracking.CurrentReleaseDate,
```
pxp928, line 296: hmm why is this causing a duplication? Is it because of the timesacnned?

mrizzi: No, it's because of the, further down, `CertifyVulnIngest` creation while only a `VexIngest` should be created.

pxp928: So this is expected. We could be in a situation where we have a VEX for a vulnerability that does not exist. That was the reason for the creation of the  `CertifyVulnIngest`. It is probably being duplicated because of the `TimeScanned` are not the same between the two. We are doing a similar thing for OpenVEX and CDX Vex.

pxp928: This is to ensure that we have a `certifyVuln` node associated with a VEX at all times. `certifyVuln` nodes can have duplicates (based on the data source they came from or from a timescanned). The certifier also takes in an input for `daysSinceLastScan` and in that case, you could have duplicates.

pxp928: Thoughts on the other maintainers? @lumjjb, @mihaimaruseac, @jeffmendoza

mrizzi: > This is to ensure that we have a `certifyVuln` node associated with a VEX at all times. `certifyVuln` nodes can have duplicates (based on the data source they came from or from a timescanned). The certifier also takes in an input for `daysSinceLastScan` and in that case, you could have duplicates.

You lost me.
We have have both `CertifyVEXStatement` and `CertifyVuln` so we haven't a `CertifyVuln` node associated with a VEX all the times.
Am I missing/misunderstanding something?

pxp928: So I am thinking about it from a query standpoint. We are looking for `CertifyVuln` nodes to determine if a package has a vulnerability. If, for whichever reason, the package has a `CertifyVEXStatement` node attached to a vulnerability and package but the same package does not have a `CertifyVuln` node, would that cause issues in the query.

pxp928: @lumjjb @jeffmendoza wanted to get your thoughts on this

jeffmendoza: A VEX statement, on its own, is a round-about assertion that the package is affected by the vuln. Creating a `CertifyVuln` in this case is appropriate. The `Justification` should clarify that it was found in a VEX statement

mrizzi: > A VEX statement, on its own, is a round-about assertion that the package is affected by the vuln. Creating a `CertifyVuln` in this case is appropriate. The `Justification` should clarify that it was found in a VEX statement

I know what you mean but in the usage we did of the VEX statement it makes the `CertifyVuln` "useless". In fact `CertifyVEXStatement` has more information than `CertifyVuln` and it doesn't just say if a package is affected but also if it's not affected, fixed or still under investigation.
So from a user perspective, the `CertifyVEXStatement` is the first thing you want to know and having also a `CertifyVuln` attached for the same vulnerability to the same package is not adding any value to the final user's knowledge.

So guac can keep the "double" ingestion but the `CertifyVuln` looks like a subset of the  `CertifyVEXStatement` information adding no real value but that doubles the "cost" of ingestion and that's the reason behind this PR.

lumjjb: My thoughts are that VEX intention is overloaded today. As a concept VEX acts as a counter assertion of CertifyVuln. However, today, folks are using it as a way to be a vulnerability disclosure report. This is a manifestation of convenience unfortunately. So VEX as a concept should not certify if something is vulnerable. VEX statements as a data format unfortunately does...

My thoughts are that it is fine to have the duplicate of verbs for now, but eventually we want to have a more proper split of the concept vs data format, alike to "Build provenance vs SLSA" and "Software composition vs SBOM"

mrizzi: Thanks a lot to everyone for the contributions.
Let me say I'm happy to close this PR because it looks like having both is fine.

PR #1507, opened 2023-11-16
Inlcuded a faster fmt

nathannaveen:
# Description of the PR

* `make fmt` takes an extremely long time so I created `fmt-local`.
* `fmt-local` doesn't break the CI.
* `fmt-local` works by only running on files that are changed.

# PR Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,9 +58,9 @@ cover: test
 .PHONY: fmt
 fmt:
 	@echo "Testing formatting and imports"
-	test -z "$(shell find . -name '*.go' -not -path './.git/*' -not -wholename './vendor/*' -not -name '*.pb.go' -exec goimports -l -e {} \;)"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -263,3 +263,10 @@ check-goreleaser-tool-check:
 # Check that all the tools are installed.
 .PHONY: check-tools
 check-tools: check-docker-tool-check check-docker-buildx-tool-check check-docker-compose-tool-check check-protoc-tool-check check-golangci-lint-tool-check check-mockgen-tool-check check-goreleaser-tool-check
+
+.PHONY: fmt-local
+fmt-local:
```
lumjjb, line 268: nit: how about fmt-diff which may be more descriptive? (since in my mind local makes me think there's a fmt for remote)

lumjjb (approved):
LGTM! Nice addition - minor nit on naming

Makefile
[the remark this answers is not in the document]
nathannaveen: I was just adding this `fmt` so that people could have the option of running a faster version. But the new `fmt` doesn’t doesn’t break anything so it should be fine if I replaced the old `fmt` with the new one.

I have replaced the old `fmt` with the new `fmt`.

nathannaveen: I have replaced the old `fmt` with the new `fmt` because of https://github.com/guacsec/guac/pull/1507#discussion_r1418968856. So the name won't be `fmt-local` or `fmt-diff`, and will just be `fmt`.

pxp928 (approved):

PR #1511, opened 2023-11-18
Feature/1116 arango noun mutation update to return only IDs

pxp928:
# Description of the PR

Update arango (and ENT vulnerability and Package) to return IDs for mutation. Updated package, source and vulnerability to return flat struct of IDs

# PR Checklist


pxp928:
Working on fixing up the tests here.

pkg/assembler/backends/ent/backend/scorecard_test.go
```diff
@@ -490,35 +490,36 @@ func (s *Suite) TestIngestScorecards() {
...
-				{
-					Source: s1out,
-					Scorecard: &model.Scorecard{
-						Checks: []*model.ScorecardCheck{},
-						Origin: "test origin",
-					},
-				},
-			},
-		},
+		// TODO: Determine why this test fails?
```
pxp928, line 493: @mrizzi I am not sure why this test fails. Can you take a look and I can fix it up.

mihaimaruseac (approved):

jeffmendoza (approved):
For pkg/src/vuln don't we still want to update the backends to return the id struct and remove the helpers from the resolvers that pull the ids out?

pxp928:
> For pkg/src/vuln don't we still want to update the backends to return the id struct and remove the helpers from the resolvers that pull the ids out?

yes but this broke the e2e so that is the reason why I left it. (until we update key/value inmem)

PR #1514, opened 2023-11-20
Update all resolvers for Return ID

jeffmendoza:
Update all the resolvers to call the new methods, remove all old methods from backend interface.

This will still compile, but not run right until I update keyvalue in a followup.

# PR Checklist


mihaimaruseac (commented):
We need to update the tests too, it seems they're using the old symbols

jeffmendoza:
I'll just do it all together with the next one I guess

pxp928 (approved):

PR #1536, opened 2023-11-28
Complete Ent Return -ID

jeffmendoza:
Should be able to merge this branch after this.

A couple ingests were still building the nodes, these are not fixed/updated. I left TODOs, as we just want get get the interfaces updated across the backends so that we can merge this branch into main.

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/backend/certify.go
+++ b/pkg/assembler/backends/ent/backend/certify.go
@@ -53,20 +53,21 @@ func (b *EntBackend) CertifyGood(ctx context.Context, filter *model.CertifyGoodS
 	return collect(records, toModelCertifyGood), nil
 }
 
-func (b *EntBackend) IngestCertifyBad(ctx context.Context, subject model.PackageSourceOrArtifactInput, pkgMatchType *model.MatchFlags, spec model.CertifyBadInputSpec) (*model.CertifyBad, error) {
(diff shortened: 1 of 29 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/ent/backend/certify.go
```diff
@@ -53,20 +53,21 @@ func (b *EntBackend) CertifyGood(ctx context.Context, filter *model.CertifyGoodS
 	return collect(records, toModelCertifyGood), nil
 }
 
-func (b *EntBackend) IngestCertifyBad(ctx context.Context, subject model.PackageSourceOrArtifactInput, pkgMatchType *model.MatchFlags, spec model.CertifyBadInputSpec) (*model.CertifyBad, error) {
+func (b *EntBackend) IngestCertifyBadID(ctx context.Context, subject model.PackageSourceOrArtifactInput, pkgMatchType *model.MatchFlags, spec model.CertifyBadInputSpec) (string, error) {
```
mihaimaruseac, line 56: Should the return type be ID instead of string? Slightly concerned we might be using stringly-typed APIs.

pxp928: So it is specified as `ID`: https://github.com/guacsec/guac/blob/1116-return-ID/pkg/assembler/graphql/schema/certifyBad.graphql#L140 on the graphQL schema but translates to string in the resolvers.

pxp928 (approved):

pkg/assembler/backends/ent/backend/certify.go
mihaimaruseac: That makes sense, thanks for confirming!

PR #1581, opened 2023-12-12
Fix some error returns without unlocks.

jeffmendoza:
When looking at the last lock issues, I noticed some returns missing unlocks.

# PR Checklist


mihaimaruseac (approved):
This LGTMs but I'm wondering if we can fix this better by having `defer` for the unlock?

jeffmendoza:
> This LGTMs but I'm wondering if we can fix this better by having defer for the unlock?

All the verbs and single-level noun ingests work with defer, but the nested nouns (pkg/src/vuln) have more complex logic. If each stage (type/ns/name/version) were put in their own function that could work.

PR #1582, opened 2023-12-12
fix vulns query not properly evaluating isDependency

pxp928:
# Description of the PR

Update vulnerability query to properly evaluate `isDependency` as the dependency package can now be associated with the `pkgVersion`. Only if the pkg version is not specified, it would create an edge to the dependency package name node. This removes the need to run unnecessary `neighbors` query.

This also fixes the bug where arango DB would fail to run the query.

# PR Checklist

### DIFF
--- a/pkg/assembler/helpers/purl_test.go
+++ b/pkg/assembler/helpers/purl_test.go
@@ -34,6 +34,9 @@ func TestPurlConvert(t *testing.T) {
 		expected *model.PkgInputSpec
 	}{
 		{
+			purlUri:  "pkg:maven/US_export_policy/US_export_policy",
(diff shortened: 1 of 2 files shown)


pkg/assembler/helpers/purl_test.go
```diff
@@ -34,6 +34,9 @@ func TestPurlConvert(t *testing.T) {
 		expected *model.PkgInputSpec
 	}{
 		{
+			purlUri:  "pkg:maven/US_export_policy/US_export_policy",
```
pxp928, line 37: new check added to ensure that the purl was properly being evaluated.

mihaimaruseac (approved):

jeffmendoza (approved):

PR #1611, opened 2024-01-01
feature[add query-for-package-url] inital commit

### LINKED ISSUE #1538: [feature] Query for package URL (purl)
A subset of the `guacone` tooling is centered around using purl. For example:
- `guacone patch` command uses `start-purl` and `stop-purl`
- `guacone query vuln` uses `purl` as an argument

However, when querying for packages using the GraphQL API, there is no way to actually get the purl of a package without constructing it manually yourself like this:
```bash
cat results.json | jq '.data.findSoftware[] | "pkg:" + .type + "/" + .namespaces[0].namespace + "/" + .namespaces[0].names[0].name'
```
 It's fairly trivial to do so, but it adds friction.

We could use a custom [GraphQL schema directive](https://graphql.org/learn/queries/#directives). In our case, we could use a schema directive on the `id` field on `PackageVersion` to return a purl.

```gql
directive @packageURL(packageVersionID: String) on FIELD_DEFINITION
```

`gqlgen` does support [schema directives](https://gqlgen.com/reference/directives/) but we would have to do further investigation if this particular use case is supported.

Alternatively, we could implement a `findSoftwarePurl` query similar to [`findSoftware`](https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/search.graphql). However, it would require maintaining another GraphQL query for all backends. It would also potentially require a user to make two queries to get the purl for a package.

**Additional context**
Inspired by some schema directive examples from [here](https://the-guild.dev/graphql/tools/docs/schema-directives), specifically, the [date formatting one](https://the-guild.dev/graphql/tools/docs/schema-directives#formatting-date-strings)

kanchan-dhamane:
# Description of the PR
Adds the `Purl` field in `PackagVersion` module
Adds the custom resolver for the `Namespaces` field in `Package` module. It updates the `Purl` field in `PackagVersion` object.

"Fixes #1538"


mihaimaruseac (approved):
Thank you!

pkg/assembler/helpers/namespaces.resolver.go
```diff
@@ -0,0 +1,31 @@
...
+			ID:        namespace.ID,
+			Namespace: namespace.Namespace,
+		}
+		names := make([]*model.PackageName, len(namespace.Names))
+		for j, name := range namespace.Names {
+			names[j] = name
+			versions := make([]*model.PackageVersion, len(name.Versions))
+			for k, version := range name.Versions {
+				versions[k] = version
+				version.Purl = fmt.Sprintf("pkg:%s/%s/%s", packageObj.Type, namespace.Namespace, name.Name)
```
pxp928, line 23: This is great! We could also add in the version and any qualifiers to the end of the purl. This would give the complete purl without the users having to add in the version and qualifier after the fact.

pxp928 (commented):
This is great! One comment to make it even better with version, subpath and qualifiers included in the purl.

pkg/assembler/helpers/namespaces.resolver.go
pxp928, line 23: pkg/assembler/helpers/purl.go will help with reconstructing the purl with the version, subpath and qualifiers.

kanchan-dhamane: @pxp928 updated the code to use `PkgToPurl` function to generate purl.
Why the `purl` format is different than `purl` present in `sbom` file? (just curious)

kanchan-dhamane: It matches. My bad, I got confused between `cpe` and `purl` field. In the `cpe` field, the format is `<name>:<version>`, while in the purl field, it is `<name>@<version>`.
Before using the `PkgToPurl` function, I was constructing the purl incorrectly by using `:` instead of `@`. Thanks for recommending the `PkgToPurl` function.

pxp928:
hmm not sure what happened but will have to run `make generate` to fix the lint issues that are coming up. Once that is fixed, it will auto-merge.

kanchan-dhamane:
can we try executing workflow one more time?
there is no diff after executing `make generate` command.
`make generated_up_to_date` command is failing, but it's getting executed successfully locally.

pxp928:
> can we try executing workflow one more time? there is no diff after executing `make generate` command. `make generated_up_to_date` command is failing, but it's getting executed successfully locally.

re-ran but no good. Might be a bug, can you rebase on main it will re-trigger the workflow and might fix it.

Issue #731, opened 2023-04-17
Working around NATs 64M entry size limit


lumjjb:
Due to the limits of NATs having a data limit of 64MB, we can't pass in documents through the regular collector flow. In the longer term, we want to have a solution that will be able to pass a reference to a document with a temporary data store that can be pointed to in the NATs entry. However, in the meantime, the alternative is that for file ingestion, we will do the ingestion as part of the binary and talk directly to the GraphQL mutation API. I.e., we will use `guacone files` for file ingestion which will do ingestion (parsing) locally.

lumjjb:
Design doc: https://docs.google.com/document/d/1NL6KOvoD8uj_caACH3S8LNIAOKbGOCaWYo5lMwq7yt0/edit

lumjjb:
For this issue, in the time being to not add in any additional complexity, we have a short term gap close by having all the collectors mimic guacone and talk directly to the graphQL endpoint #1104

lumjjb:
Another proposal that came up last community call from @dejanb is to have an API to handle the documents directly by the ingestor, this would not introduce any new running services.

pxp928:
Based on further discussion, the decision was made to use a blob store (via cloud provider) to store SBOMs and emit an entry to the URI in the pub/sub queue. For quick demos and testing, the pub/sub will not be used to allow for quick usability without the overhead of the blob store. 

**Tasks based on these:**

- Remove nats, guaccollect, and guacIngest from docker-compse (for quick way to run without having to setup S3 or NATS)
- Add all collectors to both guacone and guacCollect (need to check which ones are missing from each)
    - Add deps.dev collector into guacone
- Provide instructions for setting up S3 (or cloud-provided storage) and connect via go-cloud for development or production environment using NATS
    - Go-cloud does support file system (so can be used for testing)
    - All auth is environment-based
    - Go-cloud wraps around messaging service
    - https://gocloud.dev/howto/pubsub/publish/
    - https://gocloud.dev/howto/pubsub/subscribe/
- Replace NATS with go-cloud to allow for other messaging services to be used (for now can keep NATS)
- NATS message structure
    - CD Events: https://github.com/cdevents/spec/pull/171
    - Reference an item in a bucket
- Open Questions: Do we delete the SBOMs once they are ingested or keep them?

afrittoli:
FYI - https://github.com/cdevents/spec/pull/171 is now merged, I hope we'll be able to release v0.4 in the next month with the SBOM URI in.

[...]

PR #1630, opened 2024-01-09
Utilize gocloud and blob store to work around pubsub message size

pxp928:
# Description of the PR

Address the sizing issue for NATS in issue https://github.com/guacsec/guac/issues/731

The key is the hash of the SBOM such that it can be retrieved later if needed based on hasSBOM data.

Tested via S3 and works by reading and writing to the bucket.
Note: currently it does not delete items from the bucket.

For local testing:

Filesystem can be used as documented here: https://gocloud.dev/howto/blob/#local

# PR Checklist


pkg/assembler/backends/arangodb/hasSBOM_test.go
```diff
@@ -159,14 +159,14 @@ var includedTestOccurrences = []testOccurrence{{
...
 
-var includedHasSBOM = &model.HasSBOMInputSpec{
-	URI:              "sbom_URI",
-	Algorithm:        "sbom_algorithm",
-	Digest:           "sbom_digest",
-	DownloadLocation: "sbom_download_location",
-	Origin:           "sbom_origin",
-	Collector:        "sbom_collector",
-}
+// var includedHasSBOM = &model.HasSBOMInputSpec{
```
pxp928, line 162: to fix lint issues. This will be uncommented once the tests are fixed.

mihaimaruseac (approved):

container_files/guac/guac.yaml
```diff
@@ -1,6 +1,9 @@
 # nats
 nats-addr: nats://nats:4222
 
+# blob store setup. Setup with blob store of choice via https://gocloud.dev/howto/blob/
+blob-addr: file:///path/to/dir
+
```
jeffmendoza, line 6: We should probably have a shared volume/directory between the host and ingestor container and have this point to that. Can do this as a followup.

pkg/blob/blob.go
```diff
@@ -0,0 +1,101 @@
...
+// such as S3, google cloud bucket, azure blob store can be used.
+// Authentication is setup via environment variables. Please refer to for
+// full documentation https://gocloud.dev/howto/blob/
+func NewBlobStore(ctx context.Context, url string) (*blobStore, error) {
+	bucket, err := blob.OpenBucket(ctx, url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open bucket with error: %w", err)
+	}
+	return &blobStore{
+		url:    url,
```
jeffmendoza, line 47: looks like url is never used, don't save if not needed.

pkg/handler/processor/process/process.go
```diff
@@ -88,8 +91,20 @@ func Subscribe(ctx context.Context, em collector.Emitter) error {
...
+		documentBytes, err := blobStore.Read(ctx, blobStoreKey)
+		if err != nil {
+			logger.Errorf("[processor: %s] failed read document to blob store: %v", uuidString, err)
+			return nil
+		}
+
 		doc := processor.Document{}
-		err := json.Unmarshal(d, &doc)
+		err = json.Unmarshal(documentBytes, &doc)
 		if err != nil {
```
jeffmendoza, line 108: Change to `if err := json.Unmarshal(documentBytes, &doc); err != nil {`

pkg/events/events.go
```diff
@@ -0,0 +1,58 @@
...
+	return ce, nil
+}
+
+// DecodeEventSubject takes in the collectedEvent bytes and converts it back into a CloudEvent to retrieve the necessary fields.
+// The returned value is the subject of the event.
+func DecodeEventSubject(ctx context.Context, collectedEvent []byte) (string, error) {
+	decodedEvent := cloudevents.NewEvent()
+
+	err := json.Unmarshal(collectedEvent, &decodedEvent)
+	if err != nil {
```
jeffmendoza, line 54: same here: `if err := json.Unmarshal(collectedEvent, &decodedEvent); err != nil {`

pkg/handler/collector/collector.go
```diff
@@ -109,17 +113,43 @@ func Collect(ctx context.Context, emitter Emitter, handleErr ErrHandler) error {
...
+	if err != nil {
+		return fmt.Errorf("failed create an event: %w", err)
+	}
+
+	keyByte, err := json.Marshal(cdEvent)
+	if err != nil {
+		return fmt.Errorf("failed marshal of document key: %w", err)
+	}
+
+	err = emitter.Publish(ctx, emitter.SubjectNameDocCollected, keyByte)
```
jeffmendoza, line 144: same here, don't ever do `err = `, always `err :=` unless explicitly setting an err that is outside an if or for statment.

jeffmendoza (approved):
Awesome, some nits

cmd/README.md
```diff
@@ -47,6 +47,7 @@ services:
 - what it does: runs the ingestor connected to nats and GraphQL
 - options:
   - nats addr
+  - blob addr
```
pxp928, line 50: Before we merge this, should we do a new release as this will cause the docker-compose to break.

pkg/handler/collector/collector.go
```diff
@@ -109,17 +113,43 @@ func Collect(ctx context.Context, emitter Emitter, handleErr ErrHandler) error {
...
 		return fmt.Errorf("failed marshal of document: %w", err)
 	}
-	err = emitter.Publish(ctx, emitter.SubjectNameDocCollected, docByte)
+
+	err = blobStore.Write(ctx, getHash(d.Blob), docByte)
+	if err != nil {
+		return fmt.Errorf("failed write document to blob store: %w", err)
+	}
+
+	cdEvent, err := events.CreateArtifactPubEvent(ctx, getHash(d.Blob))
```
lumjjb, line 134: can we maybe have a getKey helper function that takes in a blob and spits out the key in case we need to change it in the future? I do think that we probably think that we need hash + size as well. thoughts? Also, maybe it would be better if `CreateArtifactPubEvent` tool in the document metadata instead?

pkg/events/events.go
```diff
@@ -0,0 +1,58 @@
...
+
+	cdevents "github.com/cdevents/sdk-go/pkg/api"
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+)
+
+// CreateArtifactPubEvent creates a NewArtifactPublishedEvent CDEvent that is published as a CloudEvent.
+// The key is the sha256 of the *processor.Document bytes that will be used to retrieve the processor.Document
+// stored in the blob store.
+// TODO (pxp928): change to use URI Reference field once the PR merges: https://github.com/cdevents/spec/pull/171
+func CreateArtifactPubEvent(ctx context.Context, key string) (*cloudevents.Event, error) {
```
lumjjb, line 31: should this take in the `Blob` info instead and return an event?

pkg/handler/collector/collector_test.go
```diff
@@ -147,10 +159,22 @@ func testSubscribe(ctx context.Context, transportFunc func(processor.DocumentTre
...
+		blobStoreKey, err := events.DecodeEventSubject(ctx, d)
+		if err != nil {
+			logger.Errorf("[processor: %s] failed decode event: %v", uuidString, err)
+			return nil
+		}
+
+		documentBytes, err := blobStore.Read(ctx, blobStoreKey)
+		if err != nil {
+			return fmt.Errorf("failed read document to blob store: %w", err)
+		}
```
lumjjb, line 172: i think it would be helpful to abstract this where this should just deal with the document/blob struct instead of being exposed to the blob store internals?

pxp928: this is already abstracted from the blob store internals. It calls the `blobStore` struct `Read` method

pkg/events/events.go
pxp928: all the event cares about is the `key` to publish via the event. The `blob` would require reconstruction of the key (repeated steps) when the key is already known.

container_files/guac/guac.yaml
pxp928: Created an issue to capture this and will do in another PR: https://github.com/guacsec/guac/issues/1650

cmd/README.md
pxp928: Released v0.4.0

PR #1647, opened 2024-01-19
update dsse processor to not guess unpacked payload

### LINKED ISSUE #1633: [bug] `DocumentITE6Generic` does not have a corresponding parser
**Describe the bug**

`DocumentITE6Generic` document type is classified when the `payloadType` for DSSE is evaluated. This evaluation is wrong, and rather it should be using the `predicateType` as described here: https://github.com/in-toto/attestation/tree/main/spec/predicates

https://github.com/guacsec/guac/blob/main/pkg/handler/processor/dsse/dsse.go#L70C24-L85

We should also remove `DocumentITE6Generic` as it is not being used nor is there a parser associated with it.

**To Reproduce**
Create a DSSE with a SBOM payload

**Expected behavior**
Unpack of the DSSE with the proper document type that leads to the proper parser

**Screenshots**

**GUAC version**
main

**Additional context**

lumjjb:
# Description of the PR

Fixes #1633 . The format of the DSSE payload should not be handled by the processor and it should leave the handler pipeline to handle it via the guesser. DSSE implementation should be agnostic and not introspect its payload.

# PR Checklist

### DIFF
--- a/pkg/handler/processor/dsse/dsse_test.go
+++ b/pkg/handler/processor/dsse/dsse_test.go
@@ -17,9 +17,9 @@ package dsse
 
 import (
 	"encoding/base64"
-	"reflect"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"github.com/guacsec/guac/pkg/handler/processor"
 	"github.com/secure-systems-lab/go-securesystemslib/dsse"
 )
@@ -97,8 +97,8 @@ var (
 	}
 	ite6SLSADoc = processor.Document{
 		Blob:   []byte(ite6SLSA),
-		Type:   processor.DocumentITE6Generic,
-		Format: processor.FormatJSON,
+		Type:   processor.DocumentUnknown,
+		Format: processor.FormatUnknown,
 		SourceInformation: processor.SourceInformation{
 			Collector: "TestCollector",
 			Source:    "TestSource",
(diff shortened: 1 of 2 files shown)


pkg/handler/processor/dsse/dsse_test.go
```diff
@@ -97,8 +97,8 @@ var (
 	}
 	ite6SLSADoc = processor.Document{
 		Blob:   []byte(ite6SLSA),
-		Type:   processor.DocumentITE6Generic,
```
pxp928, line 100: Can we remove `DocumentITE6Generic` type also as we do not want anyone else to use it.

pxp928 (approved):
LGTM!

pkg/handler/processor/dsse/dsse_test.go
lumjjb: i think we can but i think its still useful as a catch all where then we start to see errors come up to some ITE6 things we dont handle as part of https://github.com/guacsec/guac/blob/main/pkg/handler/processor/guesser/type_ite6.go#L31

pxp928: Should we error at that level? Because it will go to the parser and error there instead. We should catch it here.

lumjjb: i think the question is whether we want the error message to say "Document with unknown type" or "ITE6 document failed to parse". I feel like the later provides more contextual information.

pxp928: yeah, makes sense.

jeffmendoza (approved):

PR #1658, opened 2024-01-23
fix: s3 collector

### LINKED ISSUE #1644: [bug] S3 collector fails to ingest
**Describe the bug**
guacone collect s3 fail to ingest data from an S3 bucket. Running the following command:
```
guacone collect s3 --s3-bucket <bucket name> --s3-region us-west-2 --s3-mp sqs --s3-queues <queue name>
```
The command returns the following output:
```
{"level":"info","ts":1704712935.496344,"caller":"logging/logger.go:70","msg":"Logging at info level"}
{"level":"info","ts":1704712935.496462,"caller":"cli/init.go:67","msg":"Using config file: /Users/amisingh/Projects/guac/guac.yaml"}
{"level":"info","ts":1704712935.497071,"caller":"cmd/s3.go:141","msg":"completed ingestion"}
```
but nothing is ingested. 

Further action needs to be done to check what is causing this error. Also, need to determine if kafka and sqs are needed for the S3 collector to function as it adds a lot of complexity. 

**Expected behavior**
Successful ingestion from the S3 bucket

**GUAC version**
main

**Additional context**
Currently S3 collector is commented out from `guacone` until this is resolved for the new release: https://github.com/guacsec/guac/pull/1645

dejanb:
# Description of the PR

This PR contains fixes fro S3 collector. There were a few issues in the collector:

* Wrong usage of cancel context
* Order of parameters
* Incomplete update of AWS GO dependencies

Fixes #1644

Additionally, I added some more documentation and examples in the command itself. I can also add some proper docs and examples, but we don't have a place in docs yet for this. There are some more improvements that we can make, but I'd tackle that in follow-up PRs

# PR Checklist


jeffmendoza (approved):
Thanks for fixing this!

cmd/guacone/cmd/s3.go
```diff
@@ -49,8 +50,29 @@ type s3Options struct {
 
 var s3Cmd = &cobra.Command{
 	Use:   "s3 [flags]",
-	Short: "listens to kafka/sqs s3 events to download documents and add them to the GUAC graph, or directly downloads from s3",
-	Args:  cobra.MinimumNArgs(0),
+	Short: "takes SBOMs and attestations from S3 compatible bucket and injects them to GUAC graph",
+	Long: `S3 collector can download one item from the storage, the whole bucket or listen to storage events using sqs/kafka (poll) and download the files as they are uploaded.
```
pxp928, line 54: "the whole bucket or listen to storage events using sqs/kafka (poll)"

Is sqs or kafka needed in this case? Could we not do polling similar to the GCP bucket? It seems like it adds complexity.

pxp928: Could we keep rechecking the bucket but only grab items based on the creation or modification datetime?

dejanb: We can do that, but it's not how people use S3 service idiomatically. And it seems to me it will in the end add more complexity (handling failures and retries always end up the biggest part of the solution like that). 
I'm not sure why is the current approach considered too complex. It hides everything behind proper interfaces, so people can extend it for their needs. I understand it adds two more dependencies, but we already have a lot of dependencies in the project so it's not so overwhelming.
Without proper notification support, we definitely couldn't use S3 collector for our needs.

pxp928: I understand your point the queue method is much more robust. So, it's not about the dependencies, it's more about the de-duplication of code. I am working with https://gocloud.dev/ which provides abstraction for both the blob store and the pubsub. I am already working to integrate this with the rest of GUAC (allowing users to use any pubsub they want and not be tied to NATS). Similarly, providing the mechanism to add a blob store that can be used (in combination with CDEvents) to pass large documents more efficiently between the `guaccollect` and `guacingest`. Lets chat in an issue or offline and I can walk you through what I am thinking and if that makes sense to do here. I will not hold up this PR any longer :)

PR #1659, opened 2024-01-23
Use graphql.HasOperationContext in arangodb assembler

ridhoq:
# Description of the PR
It seems we're not entering the logic to determine what graphql fields are present in some of the arangodb methods. This is because we are checking `ctx.Value("graphql").(graphql.OperationContext)` which isn't actually present. This is because the operation context is stored using the value `operation_context` as seen [here](https://github.com/99designs/gqlgen/blob/master/graphql/context_operation.go#L54). Instead of doing this logic here, we can just call the[ `HasOperationContext` function](https://github.com/99designs/gqlgen/blob/master/graphql/context_operation.go#L75) provided by the `gqlgen` library.

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/arangodb/pkg.go
+++ b/pkg/assembler/backends/arangodb/pkg.go
@@ -433,7 +433,7 @@ func (c *arangoClient) Packages(ctx context.Context, pkgSpec *model.PkgSpec) ([]
 		return []*model.Package{p}, nil
 	}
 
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):
Nice catch!

pxp928 (approved):
LGTM. Good catch on that

Issue #1560, opened 2023-12-07
Timer for Github Polling is Missing


nathannaveen:
**Is your feature request related to a problem? Please describe.**

The Polling option for the Github API is missing a ticker https://github.com/guacsec/guac/blob/2b1e1ae18addee55e7cd2ce87d1c4993d9f64d3c/cmd/guaccollect/cmd/github.go#L84

So the polling option is never re-running, causing the polling flag to do nothing.

**Describe the solution**

The solution to this is to use a timer to keep on re-running, similar to https://github.com/guacsec/guac/blob/2b1e1ae18addee55e7cd2ce87d1c4993d9f64d3c/cmd/guacone/cmd/osv.go#L92-L93

PR #1678, opened 2024-01-31
Included Polling for Github Collect

nathannaveen:
# Description of the PR

* Fixes https://github.com/guacsec/guac/issues/1560

# PR Checklist

### DIFF
--- a/pkg/handler/collector/github/github.go
+++ b/pkg/handler/collector/github/github.go
@@ -185,12 +185,23 @@ func (g *githubCollector) RetrieveArtifacts(ctx context.Context, docChannel chan
 		}
 	} else {
 		if g.poll {
-			g.fetchWorkflowRunArtifacts(ctx, docChannel)
-			select {
-			case <-ctx.Done():
-				return ctx.Err()
-			case <-time.After(g.interval):
+			ticker := time.NewTicker(g.interval)
+
+			poll := func(ctx context.Context) {
+				logger := logging.FromContext(ctx)
+				for {
+					select {
+					case <-ticker.C:
+						logger.Infof("Polling...")
+						g.fetchWorkflowRunArtifacts(ctx, docChannel)
+						ticker.Reset(g.interval)
+					case <-ctx.Done():
+						return
+					}
+				}
 			}
+
+			go poll(ctx)
 		} else {
 			g.fetchWorkflowRunArtifacts(ctx, docChannel)
 		}


pxp928 (approved):

cmd/guaccollect/cmd/files.go
```diff
@@ -113,7 +113,7 @@ func getCollectorPublish(ctx context.Context, blobStore *blob.BlobStore, pubsub
 	}, nil
 }
 
-func initializeNATsandCollector(ctx context.Context, pubsubAddr string, blobAddr string) {
+func initializeNATsandCollector(ctx context.Context, pubsubAddr string, blobAddr string, isPoll bool) {
```
SantiagoTorres, line 116: I wonder if, for the sake of future-proofing, is poll can be moved to an integer to indicate the polling interval. Currently as things stand, `g` holds two flags: 1) poll (yes/no) and 2) interval (numeric) that are tightly coupled (i.e., interval should never be another value unless poll is true). 

Not sure if this make sense!

nathannaveen: I am not sure if I understand you correctly, please correct me if I am wrong.

My understanding of your comment is that we should change `isPoll` to the actual poll interval. If the poll interval is `0` we don’t poll, and if it is any other positive value the we know that we have to poll.

SantiagoTorres: Essentially, this way you don't need to bookkeep two different variables and reconcile when they are conflicting (i.e., choose a winner).

nathannaveen: Thanks @SantiagoTorres!

After spending some time thinking about this, I realized that replacing the two flags (`poll` and `interval`) with just the `interval` flag would make it confusing for other developers trying to understand the code.

My reasoning is that it will become confusing for people who want to edit the code because they won’t actually know that polling is `false` when `interval` is equal to zero. 

Even if we add comments to state this, most people ignore comments and they won’t know why their polling is messing up because there would be no actual `poll` flag.

cmd/guaccollect/cmd/files.go
```diff
@@ -171,7 +171,9 @@ func initializeNATsandCollector(ctx context.Context, pubsubAddr string, blobAddr
 		if err := collector.Collect(ctx, emit, errHandler); err != nil {
 			logger.Errorf("Unhandled error in the collector: %s", err)
 		}
-		done <- true
+		if !isPoll { // poll collectors should not exit
+			done <- true
+		}
```
jeffmendoza, line 176: The oci and deps dev collector already support polling, this isn't necessary. A polling collector should exit when the context is cancelled.

nathannaveen: Removed this, thanks for the review!

cmd/guaccollect/cmd/files.go
```diff
@@ -171,7 +171,6 @@ func initializeNATsandCollector(ctx context.Context, pubsubAddr string, blobAddr
 		if err := collector.Collect(ctx, emit, errHandler); err != nil {
 			logger.Errorf("Unhandled error in the collector: %s", err)
 		}
-		done <- true
```
jeffmendoza, line 174: Don't remove this line, or the command will not exit when the collector errors.

nathannaveen: Whoops, I deleted that line by mistake!

jeffmendoza (approved):

PR #1681, opened 2024-02-01
add purl helper to convert from allPkgTree fragment

lumjjb:
# Description of the PR

add purl helper to convert from allPkgTree fragment, in preparation for implementing recursive query known.

# PR Checklist

### DIFF
--- a/pkg/assembler/helpers/purl.go
+++ b/pkg/assembler/helpers/purl.go
@@ -42,6 +42,32 @@ func PurlToPkg(purlUri string) (*model.PkgInputSpec, error) {
 	return purlConvert(p)
 }
 
+// AllPkgTreeToPurl takes one package trie evaluation and converts it into a PURL
+// it will only do this for one PURL, and will ignore other pkg tries in the fragment
+func AllPkgTreeToPurl(v *model.AllPkgTree) string {
+	ns := v.Namespaces[0]
+	nsStr := ns.Namespace
+
+	name := ns.Names[0]
+	nameStr := name.Name
+
+	versionStr := ""
+	subpathStr := ""
+	qualifierStrs := []string{}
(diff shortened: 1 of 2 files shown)


pkg/assembler/helpers/purl.go
```diff
@@ -42,6 +42,30 @@ func PurlToPkg(purlUri string) (*model.PkgInputSpec, error) {
 	return purlConvert(p)
 }
 
+func AllPkgTreeToPurl(v *model.AllPkgTree) string {
+	ns := v.Namespaces[0]
```
mdeicas, line 46: There can be more than one actual software package in this trie, but this function only extracts one. To cover the general case I think that iteration over the namespaces, names, and versions is needed.

mdeicas (changes requested):

pkg/assembler/helpers/purl.go
pxp928: Based on the query used, it could only return one or many. What is the context around this @lumjjb?

lumjjb: yea this is for singular use, let me add a docstring to explain the utility.

pxp928: So it will always be a singular result?

lumjjb: i was doing this for https://github.com/guacsec/guac/blob/recursive-known2/cmd/guacone/cmd/known.go#L497-L515, so am happy to just roll this into a private function but thought it'd be useful

pxp928 (approved):

pkg/assembler/helpers/purl.go
pxp928: with the comment there it should be good.

mdeicas (approved):

Issue #1505, opened 2023-11-15
[feature] What's the next actionable critical dependency?


lumjjb:
**Is your feature request related to a problem? Please describe.**

With the new [experimental REST interfaces](https://github.com/guacsec/guac/issues/1326) being proposed, and based on some discussion around being able to get value out of GUAC while waiting for ingestion to complete (due to the lack of data from users, or that the process of ingestion of large amounts of data taking a while), we want to enable some use cases which will provide value to users that will provide instant value upon setup of GUAC and iteratively get better when more and more data gets ingested.

We discussed several options including:

- Do i have an SBOM, If so where does that  SBOM live?
- What is my most widely used dependency
- Top level package “Scorecard” (for ossf scorecards, licenses, etc.)

This issue describes the first option.

**Describe the solution you'd like**

Rewriting the framing, expanding beyond just widely used dependencies to 
"What's the next actionable critical dependency?" 

As SBOMs and other metadata are ingested into GUAC, we want to be able to get a good idea of what are the critical risks within and organization that we can make actionable steps on. 

One example, would be figuring out which is the most widely used dependency. This could be:
1. Checking what the package with the highest number of IsDependency edges / or HasSBOM edges are
1. Having a HasMetadata node for criticality and weighing that score based on that
1. For those packages, rank them based on their scorecards score or vulnerability scores

The output should be a table/list with the packages to pay attention to as well as a list of problems that they have that are actionable (for example, low scorecards score, or X critical vulnerabilities). This list can then be used by a security operator or developer to triage.

**Describe alternatives you've considered**

Other alternatives to have issues opened up for:
- Do i have an SBOM, If so where does that  SBOM live?
- Top level package “Scorecard” (for ossf scorecards, licenses, etc.)

nathannaveen:
I would like to work on this

mdeicas:
@nathannaveen, @pxp928, and I chatted about this. The plan is to first identify critical dependencies using hasSbom nodes and expose that as an endpoint in the Rest API, and then add other endpoints that incorporate additional metrics such scorecard score, vuln information, and metadata. Pagination in the GQL API #1525  is needed for this analysis to scale, but we can revisit when that is added.

nathannaveen:
[...]

PR #1705, opened 2024-02-12
Get Next Actionable Critical Dependencies Part 1

nathannaveen:
# Description of the PR

* The code gets all dependencies using `hasSBOM`
* This code finds the number of dependents for each package by finding the number of packages that depends on each individual package
* Also added a REST endpoint.
* Design doc: https://docs.google.com/document/d/1892EPekBkHbOSV2iL_61HGV0uzJHsO205mEKkHIonZ0/edit?usp=sharing
* Addresses the first part of https://github.com/guacsec/guac/issues/1505. “Checking what the package with the highest number of IsDependency edges / or HasSBOM edges are”.

# PR Checklist


nathannaveen:
@mdeicas 👀

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,193 @@
...
+
+	sboms, err := model.HasSBOMs(ctx, gqlClient, model.HasSBOMSpec{})
+
+	if err != nil {
+		return nil, fmt.Errorf("error getting dependencies: %v", err)
+	}
+
+	for _, resp := range sboms.HasSBOM {
+		// Skip entries from "deps.dev" because they are inconsistent.
+		if resp.Origin == "deps.dev" {
```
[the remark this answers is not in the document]
nathannaveen: Done, thanks!

lumjjb:
@mdeicas , would you be able to help review this?

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dependencies
+
+import (
+	"context"
+	"fmt"
+	"github.com/Khan/genqlient/graphql"
+	"github.com/guacsec/guac/cmd/guacone/cmd"
```
jeffmendoza, line 22: We shouldn't be calling anything in `cmd/`, move it out to it's own package.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"github.com/guacsec/guac/pkg/handler/collector/deps_dev"
+	"sort"
```
jeffmendoza, line 25: Separate out stdlib imports from others.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func New(ctx context.Context, gqlClient graphql.Client) Dependencies {
+	return &dependencies{
+		gqlClient: gqlClient,
+		ctx:       ctx,
+	}
+}
+
+func (deps *dependencies) GetSortedDependents() ([]PackageName, error) {
```
jeffmendoza, line 53: The interface and object are all not needed, there is no need to save the state in an object. Just make this a function:
`func GetSortedDependents(ctx context.Context, gqlClient graphql.Client) ([]PackageName, error)`

pkg/dependencies/dependents_test.go
```diff
@@ -0,0 +1,411 @@
...
+package dependencies
+
+import (
+	"context"
+	"fmt"
+	"github.com/Khan/genqlient/graphql"
+	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"net/http"
+	"testing"
+	"time"
```
jeffmendoza, line 27: Separate out stdlib imports from others.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+
+	sort.Slice(packagesArr, func(i, j int) bool {
+		return packagesArr[i].DependentCount > packagesArr[j].DependentCount
+	})
+	return packagesArr, nil
+}
+
+// findAllDependents queries a GraphQL endpoint to find all dependencies and constructs a map of dependencyNode.
+func findAllDependents(gqlClient graphql.Client) (map[string]dependencyNode, error) {
+	ctx := context.Background()
```
jeffmendoza, line 74: ctx should be passed in, not background.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+
+	// Initialize a map to hold package names mapped to their dependency nodes. This will be the returned map
+	packages := map[string]dependencyNode{}
+
+	// Initialize maps to hold dependency and dependent edges.
+	// dependentEdges maps a package ID to the IDs of packages that depend on it (DepPkg to Pkg).
+	// dependencyEdges maps a package ID to the IDs of its dependencies (Pkg to DepPkg).
+	dependentEdges := make(map[string][]string)  // id -> dependents ids
+	dependencyEdges := make(map[string][]string) // id -> dependency ids
```
jeffmendoza, line 83: Use either `make` or `{}` consistently, don't mix.
L77 uses `{}` and here uses `make`

mdeicas (changes requested):
Thanks for working on this @nathannaveen!! I added some comments on the code but also wanted to ask about the granularity of the dependency relationship at a higher level (see the comment on this).

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,194 @@
...
+			pkgId := dependency.Package.Namespaces[0].Names[0].Versions[0].Id
+
+			if len(dependency.DependencyPackage.Namespaces[0].Names[0].Versions) == 0 {
+				findMatchingDepPkgVersionIDs, err := cmd.FindDepPkgVersionIDs(ctx, gqlClient, dependency.DependencyPackage.Type,
+					dependency.DependencyPackage.Namespaces[0].Namespace,
+					dependency.DependencyPackage.Namespaces[0].Names[0].Name, dependency.VersionRange)
+				if err != nil {
+					return nil, fmt.Errorf("error from FindMatchingDepPkgVersionIDs:%w", err)
+				}
+				depPkgIds = append(depPkgIds, findMatchingDepPkgVersionIDs...)
```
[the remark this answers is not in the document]
pxp928: no there if the version is an empty string, it would use the version range:
https://github.com/guacsec/guac/blob/431a2863ffd7ceef08e48661611704803dba0ab0/pkg/ingestor/parser/common/helpers.go#L119-L124

pkg/dependencies/dependents.go
nathannaveen: Moved it out of `cmd`

pkg/dependencies/dependents.go
nathannaveen: Fixed the imports.

pkg/dependencies/dependents_test.go
nathannaveen: Fixed the imports.

pkg/dependencies/dependents.go
nathannaveen: Removed the interface and refactored the function to take in `ctx` and `gqlClient` as parameters.

pkg/dependencies/dependents.go
nathannaveen: Taking `ctx` as a parameter.

pkg/dependencies/dependents.go
nathannaveen: Replaced all of them with `make` not `{}`.

Thanks for the reviews!

pkg/guacrest/server/server.go
```diff
@@ -36,10 +37,34 @@ func (s *DefaultServer) HealthCheck(ctx context.Context, request gen.HealthCheck
...
-	return nil, fmt.Errorf("Unimplemented")
+	if request.Params.Sort == "frequency" {
+		deps := dependencies.New(ctx, s.gqlClient)
+		packages, err := deps.GetSortedDependents()
+		if err != nil {
+			return gen.AnalyzeDependencies500JSONResponse{
+				InternalServerErrorJSONResponse: gen.InternalServerErrorJSONResponse{
+					Message: fmt.Sprintf("failed to get sorted dependents %v", err.Error()),
+				},
+			}, fmt.Errorf("failed to get sorted dependencies %v", err)
```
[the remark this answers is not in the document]
nathannaveen: Returning `nil` as the error.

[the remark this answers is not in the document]
nathannaveen: If the returned error is always `nil` and we only output an error with the json response, should the function be returning an error?

pkg/guacrest/server/server.go
```diff
@@ -36,10 +37,34 @@ func (s *DefaultServer) HealthCheck(ctx context.Context, request gen.HealthCheck
 }
 
 func (s *DefaultServer) AnalyzeDependencies(ctx context.Context, request gen.AnalyzeDependenciesRequestObject) (gen.AnalyzeDependenciesResponseObject, error) {
-	//return gen.AnalyzeDependencies200JSONResponse{
-	//  PurlListJSONResponse: []gen.Purl{},
-	//}, nil
-	return nil, fmt.Errorf("Unimplemented")
+	if request.Params.Sort == "frequency" {
```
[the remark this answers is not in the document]
nathannaveen: Thanks, the code is now using the constant.

pkg/dependencies/dependents.go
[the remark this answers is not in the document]
nathannaveen: Documented the functions.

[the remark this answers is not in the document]
nathannaveen: You are right, I have updated the name of `GetSortedDependencies` and `findAllDependents` to `GetDependenciesBySortedDependentCnt` and `findDependentsOfDependencies` respectively.

cmd/guacone/cmd/vulnerability.go
```diff
@@ -606,7 +606,7 @@ func searchPkgViaHasSBOM(ctx context.Context, gqlclient graphql.Client, searchSt
 	return path, tableRows, nil
 }
 
-func findDepPkgVersionIDs(ctx context.Context, gqlclient graphql.Client, depPkgType string, depPkgNameSpace string, depPkgName string, versionRange string) ([]string, error) {
+func FindDepPkgVersionIDs(ctx context.Context, gqlclient graphql.Client, depPkgType string, depPkgNameSpace string, depPkgName string, versionRange string) ([]string, error) {
```
jeffmendoza, line 609: If this is copied out now, then it can be removed here and the one in pkg/dependencies can be called.

pkg/guacrest/server/server.go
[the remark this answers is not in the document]
mdeicas: That's right, we don't really have a use for it.`oapi-codegen` generates the interface this way to make it easier to handle default errors (https://github.com/guacsec/guac/blob/main/pkg/guacrest/generated/server.go#L415), but then that error wouldn't be documented in the spec.

pkg/guacrest/server/server.go
```diff
@@ -36,10 +37,34 @@ func (s *DefaultServer) HealthCheck(ctx context.Context, request gen.HealthCheck
...
+			})
+		}
+
+		val := gen.AnalyzeDependencies200JSONResponse{
+			PackageNameListJSONResponse: packageNames,
+		}
+
+		return val, nil
+	}
+	return gen.AnalyzeDependencies200JSONResponse{PackageNameListJSONResponse: []gen.PackageName{}}, nil
```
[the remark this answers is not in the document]
mdeicas: I think it would be good to also check if the argument is one of the supported enums.  For example, if the argument is `sort=foo`, then the output would be "scorecard sort is unimplemented".

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,237 @@
...
+				}
+
+				// First we need to find all the packages that have pkgName as a dependency
+
+				// Initialize a visited map and a queue for BFS to find all packages that have pkgName as a dependency.
+				visited := make(map[string]bool)
+				queue := []string{depPkgId}
+
+				// Perform BFS to mark visited nodes.
+				for len(queue) > 0 {
```
mdeicas, line 140: There are two very similar graph traversals, here and on line 164. I think it would be better to factor them out to a helper function. E.g. `func reachableNodesOf(start string, edges map[string][]string) map[string]bool {...}`. Updating `packages` can be done after calling the helper function.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,237 @@
...
+				// Map the IDs to their names.
+				idToName[depPkgId] = depPkgName
+				idToName[pkgId] = pkgName
+
+				// Skip "guac" files.
+				if dependency.DependencyPackage.Type == "guac" && dependency.DependencyPackage.Namespaces[0].Namespace == "files" {
+					continue
+				}
+
+				// First we need to find all the packages that have pkgName as a dependency
```
mdeicas, line 133: If I'm reading this correctly, this comment doesn't seem accurate... this bfs finds the dependencies of `depPkgId`, not the dependents of `pkgName`.

pkg/dependencies/dependents_test.go
mdeicas, line 1: This package doesn't build anymore (not picked up by the CI because of the build tag).

pkg/dependencies/dependents_test.go
```diff
@@ -0,0 +1,413 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build integrationMerge
+
+package dependencies
```
mdeicas, line 18: Use `package dependencies_test`.

mdeicas (changes requested):

pkg/guacrest/server/server.go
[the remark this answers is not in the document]
nathannaveen: The code is now using a switch case to check for the scorecard sort, or any other type of sort.

cmd/guacone/cmd/vulnerability.go
nathannaveen: Removed the duplicate one in vulnerability.go

pkg/dependencies/dependents_test.go
nathannaveen: Changed the build tag

pkg/dependencies/dependents_test.go
nathannaveen: We can’t use `package dependencies_test` because `findDependentsOfDependencies` is an internal function, and won’t be able to be called in the package `dependencies_test`.

pkg/dependencies/dependents.go
nathannaveen: I don’t think we should do this, because the readability won’t be affected that much, but we would have to preform a second loop for the following code snippet, which would further slow down the code.

``` go
for len(queue) > 0 {
	n := len(queue)

	for i := 0; i < n; i++ { // go through the entire row
		nodeId := queue[0]
		queue = queue[1:]

		node := idToName[nodeId]

		for depPkgNodeId := range visited {
			depPkgNode := idToName[depPkgNodeId]
			if _, ok := packages[depPkgNode]; !ok {
				packages[depPkgNode] = dependencyNode{dependents: make(map[string]bool)}
			}

			// Mark the node as a dependent.
			packages[depPkgNode].dependents[node] = true
		}

		// Add dependent nodes to the queue.
		queue = append(queue, dependentEdges[nodeId]...)
	}
}
```

pkg/dependencies/dependents.go
nathannaveen: You are correct, I have changed the comment to `First we need to find all the packages that have pkgName as a dependent`.

pkg/dependencies/dependents.go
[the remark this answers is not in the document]
nathannaveen: @mdeicas and I chatted about this, and were thinking about keeping an endpoint for this current feature being created with this PR and including another endpoint for the feature described in this comment.

@pxp928 @jeffmendoza @lumjjb @mihaimaruseac Thoughts?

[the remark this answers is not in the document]
mihaimaruseac: Separating the endpoints and documenting the differences seems ok to me, this seems to be a complex usecase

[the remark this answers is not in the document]
pxp928: +1 to making two different endpoints

pkg/dependencies/dependents.go
mdeicas: I think that factoring this out would make it easier to see that that some parts of this code are executed more than once on the same inputs. 

At a high level, this function finds the transitive closure of a graph of where the nodes are `Packages` and edges are dependent relationships. The input graph has the nodes and edges defined by the `IsDependency`s from a single `HasSbom`. 

One way to do this is, for every node `v` in the graph, find the reachable nodes of `v` and add to the result. On the other hand, this code does: for every edge `(v, u)`, find the reachable nodes of `u`, the nodes that reach `v`, and then add to the result. 

There are few issues with this approach. First, the graph traversals (i.e. getting reachable / reached nodes) may be done on the same node more than once, because they are done per edge, and nodes can appear in more than one edge. Secondly, for all `v`, it isn't required to find both the nodes that reach `v` and the nodes that `v` reaches, just one of these is enough to find the transitive closure. These changes would also improve readability.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,237 @@
...
+				}
+
+				// Next we want to find all the packages that are dependencies of pkgName.
+				// We need to add them all to the dependencies of all nodes that have pkgName as a dependency.
+
+				// Reset the queue to find all packages that are dependencies of pkgName.
+				queue = []string{pkgId}
+
+				// Perform BFS to add pkgName and dependencies of it to all nodes that have pkgName as a dependency.
+				for len(queue) > 0 {
```
mdeicas, line 164: Is it possible to remove one of these loops, in order to reduce the nesting depth of this BFS?

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,237 @@
...
+	for _, resp := range sboms.HasSBOM {
+		// Skip entries from "deps.dev" because they are inconsistent.
+		if resp.Origin == deps_dev.DepsCollector {
+			continue
+		}
+		// Iterate through the included dependencies of each SBOM.
+		for _, dependency := range resp.IncludedDependencies {
+			// Construct unique names for the dependency package and the package itself.
+			// TODO: Make the names actually unique, not just add "_".
+			depPkgName := dependency.DependencyPackage.Type + "_" + dependency.DependencyPackage.Namespaces[0].Namespace + "_" + dependency.DependencyPackage.Namespaces[0].Names[0].Name
```
mdeicas, line 105: Is it possible to have the API endpoint output the package as a purl?

pkg/dependencies/dependents.go
mdeicas: Sorry to nitpick! But I think this isn't 100% accurate, there may be _other_  `IsDependency` relationships, not yet iterated through and so not accumulated in the two edge sets, that give new direct dependencies of `pkgName`.  This finds the dependencies of `depPkgId` (along with `depPkgId` itself), not `pkgId`/`pkgName`. 

Also, I think the comments on lines 157-163 need similar changes.

pkg/dependencies/dependents.go
nathannaveen: I have refactored the two graph traversals out into a function, thanks!

pkg/dependencies/dependents.go
nathannaveen: Thanks, I see what you mean. I have updated the comments to include that we are finding all the packages from the packages that are already scanned.

pkg/dependencies/dependents.go
nathannaveen: We can’t remove these loops because we have to first get all the dependent packages, and then assign all the newly acquired dependency packages as the dependencies of the dependent packages.

pkg/dependencies/dependents.go
nathannaveen: I don’t think that we can output as a purl because we have two types of data for each package, the package name, and the number of dependents.

pxp928:
@nathannaveen once PR https://github.com/guacsec/guac/pull/1807 merged, rebase and your tests should pass.

nathannaveen:
> @nathannaveen once PR #1807 merged, rebase and your tests should pass.

Thanks @pxp928!

pkg/dependencies/dependents.go
lumjjb: +1 on outputting the package name as a PURL instead of `_` delimetered strings (see https://github.com/guacsec/guac/blob/main/pkg/assembler/helpers/purl.go#L47). I'm not sure what you mean we can't do this because we have two dimensions of data? Would it be a PURL to data mapping?

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,230 @@
...
+// a map of dependencyNode, where each node represents a package and its dependents. This function is designed to
+// identify and map out the relationships between packages based on the SBOM data. It filters out inconsistent data
+// from "deps.dev" and handles packages with and without specific version IDs. The function employs a breadth-first search
+// (BFS) algorithm to traverse the dependency graph and populate the map with packages and their respective dependents.
+//
+// Returns:
+//   - A map where the key is a string representing the package name, and the value is a dependencyNode containing
+//     the dependents of the package.
+//   - An error
+func findDependentsOfDependencies(ctx context.Context, gqlClient graphql.Client) (map[string]dependencyNode, error) {
```
lumjjb, line 77: can you help break out this function and refactor some of the blocks out to aid readability?

pkg/dependencies/dependents.go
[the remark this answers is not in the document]
lumjjb: would the code be shared for most of this? can we have this as an option or as a flag? Since we are doing the counts already.

cmd/guacone/cmd/vulnerability.go
```diff
@@ -606,41 +606,41 @@ func searchPkgViaHasSBOM(ctx context.Context, gqlclient graphql.Client, searchSt
...
+//		// vulns even if some paths cannot be followed.
+//		matchingDepPkgVersions = nil
+//		//return nil, nil, fmt.Errorf("error determining dependent version matches: %w", err)
+//	}
+//
+//	for matchingDepPkgVersion := range matchingDepPkgVersions {
+//		matchingDepPkgVersionIDs = append(matchingDepPkgVersionIDs, depPkgVersionsMap[matchingDepPkgVersion])
+//	}
+//	return matchingDepPkgVersionIDs, nil
+//}
```
lumjjb, line 643: any reason we are keeping the old code around?

lumjjb (changes requested):

pkg/dependencies/dependents.go
nathannaveen: Oh, I don't think I understood what @mdeicas meant before; I have updated the spec so that it outputs the packages as purls. Thanks!

pkg/dependencies/dependents.go
nathannaveen: I refactored out part of the function into `updatePackagesAndNames`.

cmd/guacone/cmd/vulnerability.go
nathannaveen: Oh, sorry, that was my mistake; I commented it out but forgot to delete it. Thanks!

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+					return nil, fmt.Errorf("error from FindMatchingDepPkgVersionIDs:%w", err)
+				}
+				depPkgIds = append(depPkgIds, findMatchingDepPkgVersionIDs...)
+			} else {
+				depPkgIds = append(depPkgIds, dependency.DependencyPackage.Namespaces[0].Names[0].Versions[0].Id)
+			}
+
+			for _, depPkgId := range depPkgIds {
+				// Inside the loop where you iterate through dependencies
+				updatePackagesAndNames(idToName, packages, depPkgId, pkgId, depPkgName, pkgName, dependency.DependencyPackage.Type, dependency.DependencyPackage.Namespaces[0].Namespace, dependencyEdges, dependentEdges)
```
mdeicas, line 125: Nit: I think it would be helpful to check for type "guac" and namespace "files" before the function call, to reduce the number of parameters in this function.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+	// Skip "guac" files.
+	if depPkgType == "guac" && depPkgNamespace == "files" {
+		return
+	}
+
+	// Map the IDs to their names.
+	idToName[depPkgId] = depPkgName
+	idToName[pkgId] = pkgName
+
+	// First, we need to find all the packages that have pkgName as a dependency.
```
mdeicas, line 153: I think that the code first finds packages that have `pkgName` as a *dependent*, on line 157. What this comment describes is done second, on line 162.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+	// Note that we are only searching of packages with pkgName as a dependency from the packages that have scanned so far.
```
mdeicas, line 154: Nit: `... from the *edges* that we have scanned ...`

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// This dependencyPackages map finds all packages that have pkgName as a dependent out of our pre-scanned packages.
+	dependencyPackages := traverseGraph(depPkgId, dependencyEdges)
+
+	// Next we want to find all the packages that are dependencies of pkgName.
```
mdeicas, line 159: What follows this comment finds the *dependents* of `pkgName`, or alternatively the packages that *have* `pkgName` as a dependency.  Also see my other review comment -- I think that at some point the order of the `traverseGraph` calls was swapped but these comments did not go with it? 

Also, since this wording can be tricky, I would recommend sticking to either the direct form of `the packages that are dependencies / dependents of foo` or the indirect form of `the packages that have foo as a dependent / dependency`,  instead of swapping between the two.

pkg/guacrest/server/server.go
```diff
@@ -37,7 +38,37 @@ func (s *DefaultServer) HealthCheck(ctx context.Context, request gen.HealthCheck
...
+				InternalServerErrorJSONResponse: gen.InternalServerErrorJSONResponse{
+					Message: err.Error(),
+				},
+			}, nil
+		}
+
+		var packageNames []gen.PackageName
+
+		for _, p := range packages {
+			pac := p // have to do this because we don't want packageNames to keep on appending a pointer of the same variable p.
```
mdeicas, line 55: Nit: I don't think you need this anymore because the address of `pac` is not stored.

mdeicas (commented):
Thanks for making the changes @nathannaveen! A few more minor comments.

pkg/dependencies/dependents_test.go
```diff
@@ -0,0 +1,419 @@
...
+	name     string
+	version  string
+	typeName string
+}
+
+func Test_findAllDependencies(t *testing.T) {
+	graphqlEndpoint := "http://localhost:8080/query"
+
+	httpClient := http.Client{}
+	gqlClient := graphql.NewClient(graphqlEndpoint, &httpClient)
```
mdeicas, line 54: Could you also switch the tests to use https://github.com/guacsec/guac/blob/main/internal/testing/graphqlClients/testsetup.go#L38? This will start up a new graphql server for each test and avoid using the global one that the CI starts.

nathannaveen: Switched, thanks!

pkg/guacrest/server/server.go
nathannaveen: Done

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+	// dependentEdges maps a package ID to the IDs of packages that depend on it (DepPkg to Pkg).
+	// dependencyEdges maps a package ID to the IDs of its dependencies (Pkg to DepPkg).
+	dependentEdges := make(map[string][]string)  // id -> dependents ids
+	dependencyEdges := make(map[string][]string) // id -> dependency ids
+
+	// Initialize a map to convert package IDs to their names.
+	idToName := make(map[string]string)
+
+	sboms, err := model.HasSBOMs(ctx, gqlClient, model.HasSBOMSpec{})
+
```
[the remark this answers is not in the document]
nathannaveen: Done

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, fmt.Errorf("error getting dependencies: %v", err)
+	}
+
+	for _, resp := range sboms.HasSBOM {
+		// Skip entries from "deps.dev" because they are inconsistent.
+		if resp.Origin == deps_dev.DepsCollector {
+			continue
+		}
+		// Iterate through the included dependencies of each SBOM.
+		for _, dependency := range resp.IncludedDependencies {
```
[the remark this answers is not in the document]
nathannaveen: Changed, thanks

pkg/dependencies/dependents.go
nathannaveen: Done, the code is now checking them outside of the helper func.

pkg/dependencies/dependents.go
nathannaveen: Done

pkg/dependencies/dependents.go
nathannaveen: Done

pkg/dependencies/dependents.go
nathannaveen: Done. The comments were getting kind of confusing before, thanks for the reviews!

nathannaveen:
Everyone, thanks for your reviews! 

This PR has been open for a pretty long time, so if there are any nits in the future, can we create an issue for them so that we can get this PR merged in?

pkg/dependencies/dependents_test.go
```diff
@@ -0,0 +1,413 @@
...
+}
+
+type specInfo struct {
+	name     string
+	version  string
+	typeName string
+}
+
+func Test_findAllDependencies(t *testing.T) {
+	gqlClient := clients.SetupTest(t)
```
mdeicas, line 51: This should be called inside each subtest, otherwise each test may affect each other.

pkg/dependencies/dependents.go
```diff
@@ -0,0 +1,242 @@
...
+	idToName[pkgId] = pkgName
+
+	// First, we need to find all the packages that are dependencies of pkgName.
+	// We need to add them all to the dependencies of all nodes that have pkgName as a dependent.
+	// Note that we are only searching for dependencies of pkgName from the edges that have scanned so far
+	dependencyPackages := traverseGraph(depPkgId, dependencyEdges)
+
+	// Next we want to find all the packages that have pkgName as a dependency.
+	// Note that we are only searching of packages with pkgName as a dependency from the edges that have scanned so far.
+	// This dependentPackages map finds all packages that have pkgName as a dependent out of our pre-scanned packages.
```
mdeicas, line 159: On line 159, it should be `that have pkgName as a dependency` instead of `that have pkgName as a dependent`.

mdeicas (approved):
Aside from these two comments, LGTM @nathannaveen!!

pkg/dependencies/dependents_test.go
nathannaveen: Oh, thanks didn't think of that!

pkg/dependencies/dependents.go
nathannaveen: Changed the wording, thanks!

lumjjb (approved):

PR #1708, opened 2024-02-16
Id or inputspec

pxp928:
# Description of the PR

Fixes issue https://github.com/guacsec/guac/issues/1261 and updates the `keyvalue` backend to use noun IDs else use the inputspec for Verb ingestion. 

Updates the assembler (bulk and single ingestion) to update the `IDorInputSpec` for each noun with its IDs before passing them to the verbs to be used. 

NOTE: The backends must ensure that the IDs are returned in the same order as the array of InputSpecs, added comments on noun ingestion to call this out.

This PR also deletes all the old unit tests for each backend, in favor of the centralized test suite.

Updates still need to be made on the `arango` and `ent` backend. Backend test suite needs to be updated to include ingested noun IDs in verb ingestion.

# PR Checklist


pxp928:
Working to fix up the failing tests

pxp928:
Ready for review

pxp928:
@mrizzi FYI. Will need to make some updates on the ENT backend for this to be fully utilized. I will start some of the work today and we can sync up on it.

mihaimaruseac (commented):
I'm wondering, wouldn't it be better to make the input specs have an ID field and remove the `!`s and move the requirement for the fields to be specified and not null in the root resolvers instead? It would at least remove remove all code changes for the verbs, I think.

Makefile
```diff
@@ -23,9 +23,7 @@ test: generate
 # To run it locally you can run the following command: make start-integration-service
 .PHONY: integration-test
 integration-test: generate check-env
-	# "-p 1" because We don't want the tests in internal/testing/backend
-	# and pkg/assembler/backend running at same time
-	go test -p 1 -tags=integration ./...
+	go test -tags=integration ./...
```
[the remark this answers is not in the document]
pxp928: we no longer have `pkg/assembler/backend` tests so this should not be an issue? Correct me if I am wrong @jeffmendoza

pkg/assembler/clients/helpers/bulk.go
```diff
@@ -180,110 +187,224 @@ func GetBulkAssembler(ctx context.Context, gqlclient graphql.Client) func([]asse
...
 	}
 }
 
-func ingestPackages(ctx context.Context, client graphql.Client, v []model.PkgInputSpec) ([]string, error) {
-	response, err := model.IngestPackages(ctx, client, v)
+// ingestPackages takes in the map of IDorPkgInput which contains the pkgInputSpec and outputs a map that contains the pkgIDs to be used for verb ingestion
+func ingestPackages(ctx context.Context, client graphql.Client, packageInputMap map[string]*model.IDorPkgInput) (map[string]*model.IDorPkgInput, error) {
+	var keys []string
+	var pkgInputs []model.IDorPkgInput
+	pkgInputs = make([]model.IDorPkgInput, 0)
```
jeffmendoza, line 222: you can remove all these `slice = make...`. The append will create the slice.

cmd/guacone/cmd/gcs_test.go
```diff
@@ -83,17 +81,18 @@ func TestValidateGCSFlags(t *testing.T) {
...
-	rootCmd.AddCommand(collectCmd)
-	rootCmd.AddCommand(filesCmd)
-	bz2Path := "./../../../internal/testing/testdata/exampledata/busybox-cyclonedx.json.bz2"
-	rootCmd.SetArgs([]string{"collect", "files", bz2Path})
-	err := rootCmd.Execute()
-	if err != nil {
-		t.Fatal(err)
-	}
-}
+// TODO: what is the purpose of this test?
```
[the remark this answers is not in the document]
pxp928: hmmm strange part is...how did this ever work?

pkg/assembler/graphql/schema/artifact.graphql
```diff
@@ -53,14 +53,26 @@ input ArtifactInputSpec {
...
   artifacts(artifactSpec: ArtifactSpec!): [Artifact!]!
 }
 
 extend type Mutation {
-  "Ingests a new artifact and returns it. The returned ID can be empty string."
-  ingestArtifact(artifact: ArtifactInputSpec): ID!
-  "Bulk ingests new artifacts and returns a list of them. The returned array of IDs can be a an array of empty string."
-  ingestArtifacts(artifacts: [ArtifactInputSpec!]!): [ID!]!
+  "Ingests a new artifact and returns it."
+  ingestArtifact(artifact: IDorArtifactInput): ID!
```
jeffmendoza, line 75: Why do the noun ingestions use "IDorNounIput" instead of the existing "NounInputSpec"?

pxp928:
> I'm wondering, wouldn't it be better to make the input specs have an ID field and remove the !s and move the requirement for the fields to be specified and not null in the root resolvers instead? It would at least remove remove all code changes for the verbs, I think.

We chatted about this and this would be the cleanest approach without removing which fields are required.

mihaimaruseac (approved):

jeffmendoza (approved):
Not sure about the schema change for noun ingests. Maybe ask the other maintainers.

Otherwise looks good with a small nit.

Makefile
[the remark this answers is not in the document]
jeffmendoza: This is correct.

mihaimaruseac:
Approving in this case, though I still think that it adds one extra indirection in the query. The generated code now does double verification that the fields are not null, if we remove the `!` the generated code would do none and we'd add it manually in the top-level resolvers. But I think this also changes the types of some arguments, so let's go with this at least for now.

jeffmendoza:
I like the "union" approach as it is more clear to a reader of the schema what is and isn't required for an ingest, otherwise it would just be documentation or reading code to tell you that you can either supply ID or the rest of the fields in the object.

pkg/assembler/clients/helpers/bulk.go
pxp928: so this causes an error when the `packageInputMap` is nil. That is the reason to add all these in. It resulted in these errors:
```
IngestSources failed with error: returned error 422 Unprocessable Entity: {\"errors\":[{\"message\":\"cannot be null\",\"path\":[\"variable\",\"sources\"],\"extensions\":{\"code\":\"GRAPHQL_VALIDATION_FAILED\"}}],\"data\":null}",
```

pkg/assembler/graphql/schema/artifact.graphql
pxp928: Yeah, this I did to keep it consistent. It would always be the `NounInputSpec`

pxp928: I can change them all back if that is what the others think is best...

pxp928: @mihaimaruseac ?

mihaimaruseac: I think it's better to keep all consistent

pkg/assembler/clients/helpers/bulk.go
pxp928: The resulting slice is nil if the append does not occur and this is not allowed by the schema.

jeffmendoza: ah, got it.

cmd/guacone/cmd/gcs_test.go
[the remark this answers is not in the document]
pxp928: Moved this test to under processor unit test where it belongs.

Issue #1525, opened 2023-11-21
[feature] Enable pagination for GUAC GraphQL APIs


lumjjb:
**Is your feature request related to a problem? Please describe.**
Pagination as a feature has been brought up multiple times. For example, for data display and also to enable faster queries (i.e. prevent UI stalling).

**Describe the solution you'd like**
Enable certain queries that may return large amounts of data to allow pagination.

**Describe alternatives you've considered**
NA

**Additional context**

pxp928:
Based on recommended approaches, the [Relay-style](https://relay.dev/graphql/connections.htm) seems to be the most recommended and compatible with Apollo and other graphql clients. 

Note: ApolloClient can be used to enable caching of the results.

The graphQL schema needs to be updated with the following (per noun and verb):

```
type ArtifactConnection {
    totalCount: Int!
    pageInfo: PageInfo!
    edges: [ArtifactsEdge!]!
}

type ArtifactsEdge {
  cursor: ID!
  node: Artifact
}

extend type Query {
  "Returns all artifacts matching a filter."
  artifacts(artifactSpec: ArtifactSpec!): [Artifact!]!
  artifactsList(artifactSpec: ArtifactSpec!, after: ID, first: Int, before: ID, last: Int): ArtifactConnection
}
```
with `PageInfo` defied as below:
```
type PageInfo {
    hasNextPage: Boolean!
    hasPreviousPage: Boolean!
    startCursor: ID
    endCursor: ID
}
``` 

With the `Edge`, `Connection` defined as:

- A **connection** is a paginated field on an object — for example, the friends field on a user or the comments field on a blog post.
- An **edge** has metadata about one object in the paginated list, and includes a cursor to allow pagination starting from that object.
- A **node** represents the actual object you were looking for.
- **pageInfo** lets the client know if there are more pages of data to fetch. In the Relay specification, it doesn’t tell you the total number of items, because the client cache doesn’t need that info. It would be up to the developer to expose that information through another field.

Example test implementation for the key/value backend can be found here: https://github.com/guacsec/guac/compare/main...pxp928:artifact-ff:test-pagination-artifact?expand=1

Example implementation with ENT: https://betterprogramming.pub/clean-architecture-with-ent-and-gqlgen-a789933a3665

Other useful docs: https://medium.com/@chris.czurylo/implementing-pagination-in-graphql-and-go-using-gqlgen-2ea3786a71dc

pxp928:
[...]

PR #1720, opened 2024-02-21
Add Pagination to the Rest API

mdeicas:
# Description of the PR

### Pagination Scheme

The proposed scheme is based on opaque cursors that point to entries in an input dataset. Each request will (optionally) contain a `PageSize` integer field and a `Cursor` string field. Each response will contain a `NextCursor` field and optionally a `TotalCount` field. If there are no more results, the `NextCursor` field will be the empty string. 

Using opaque cursors enables different cursor formats for different use cases. An endpoint that makes calls to the GraphQL server can use the GraphQL server generated cursors, and an endpoint that directly uses the backend database can use the database primary keys or similar. 

I added an implementation of this pagination scheme that uses base64 encoded indices as the cursors. Endpoints aren't required to use it, but I think it'll work for most cases. 

### HTTP request
The pagination spec is passed as query parameters, like in:
`http://0.0.0.0:8081/query/dependencies?purl=guac&PageSize=3&Cursor=AwAAAAAAAAA`

## Alternatives
One alternative approach is a simpler limit + offset scheme, but this doesn't support interfacing with the GraphQL pagination scheme (see #1525). Another approach is to implement a more expressive scheme similar to the GQL Relay-style pagination. This would allow users to paginate backwards through results, and possibly to paginate starting from any entry in a page. However, this requires more support and documentation, and I don't think the functionality gained is that important. 

The pagination spec could also be passed in the http header. 

# PR Checklist

### DIFF
--- a/pkg/guacrest/openapi.yaml
+++ b/pkg/guacrest/openapi.yaml
@@ -100,8 +100,9 @@ components:
       type: string
     PaginationInfo:
       type: object
-      required:
-        - NextCursor
+      description: >
+        Contains the cursor to retrieve more pages. If there are no more, 
+        NextCursor will be nil.
       properties: 
         NextCursor:
           type: string
--- a/pkg/guacrest/openapi.yaml
+++ b/pkg/guacrest/openapi.yaml
@@ -33,6 +33,7 @@ paths:
       summary: Retrieve the dependencies of a package
       operationId: retrieveDependencies
       parameters:
+        - $ref: "#/components/parameters/PaginationSpec"
         - name: purl
           description: the purl of the dependent package
           in: query
@@ -51,8 +52,9 @@ paths:
   "/analysis/dependencies":
     get:
       summary: Identify the most important dependencies
-      operationId: analyzeDependencies 
+      operationId: analyzeDependencies
       parameters:
+        - $ref: "#/components/parameters/PaginationSpec"
         - name: sort
           description: >
             The sort order of the packages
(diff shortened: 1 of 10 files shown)


pkg/guacrest/openapi.yaml
```diff
@@ -77,41 +79,74 @@ paths:
...
+            type: integer 
+          Cursor: 
+            type: string
   schemas:
     Purl:
       type: string
+    PaginationInfo:
+      type: object
+      required:
+        - NextCursor
```
nathannaveen, line 104: Should `NextCursor` be required?

I am asking because `NextCursor` is being set to an empty string on the last page of the REST output which clients would have to interpret correctly.

If `NextCursor` wasn’t required and could be nullable, then we could set `NextCursor` to nil. 

i.e.

``` go
end := start + uint64(pagesize) // end is exclusive
var nextCursor *string // Use a pointer to string to allow nil value
if end < inputLength {
    cursorStr := cursorFromIndex(end)
    nextCursor = &cursorFromIndex(end) // Only set cursorFromIndex if there are more pages
}

info := models.PaginationInfo{
    NextCursor: nextCursor, // This can now be nil if there are no more items
    TotalCount: PointerOf(len(lst)),
}
```

And, for the client side handling
* When `NextCursor` isn’t provided in the response or is `nil`, then they know that there are no more pages to fetch.
* If `NetCursor` isn’t provided from the used, we could interpret this as a request for the first page of data.

If we did this we would also have to provide comments as to the `NextCursor` being `nil`.

jeffmendoza (approved):

pxp928 (approved):
Looks great! Thanks @mdeicas. Nathan does have a good comment. So the PR will auto-merge once that is addressed.

pkg/guacrest/openapi.yaml
pxp928: @mdeicas just need to address this comment and PR is ready to merge.

PR #1806, opened 2024-04-03
Add guaccollect files option to set origin to blob path

nchelluri:
# Description of the PR

When the new `--use-blob-url` command line arg is set, the created nodes will have an origin whose value is the blob URL, instead of the source URL.

To test this, I ran `go run ./cmd/guacone collect files ~/Downloads/demo.json`, `go run ./cmd/guaccollect files ~/Downloads/demo.json --service-poll=false`, and `go run ./cmd/guaccollect files ~/Downloads/demo.json --service-poll=false --use-blob-url`. (`guacingest` and `guacgql` were running first, and `demo.json` is a small sample SPDX SBOM file.) In each case the command completed successfully and the HasSBOM node looked ok to me in the GraphQL Playground. I also expanded the File Collector automated test.

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/files.go
+++ b/cmd/guaccollect/cmd/files.go
@@ -45,8 +45,8 @@ type filesOptions struct {
 	blobAddr string
 	// poll location
 	poll bool
-	// use blob path for origin instead of source path
-	useBlobPath bool
+	// use blob URL for origin instead of source URL
+	useBlobURL bool
 }
 
 var filesCmd = &cobra.Command{
@@ -73,7 +73,7 @@ you have access to read and write to the respective blob store.`,
 			viper.GetString("pubsub-addr"),
 			viper.GetString("blob-addr"),
 			viper.GetBool("service-poll"),
-			viper.GetBool("use-blob-path"),
+			viper.GetBool("use-blob-url"),
 			args)
 		if err != nil {
 			fmt.Printf("unable to validate flags: %v\n", err)
@@ -85,7 +85,7 @@ you have access to read and write to the respective blob store.`,
 		logger := logging.FromContext(ctx)
 
 		// Register collector
-		fileCollector := file.NewFileCollector(ctx, opts.path, opts.poll, 30*time.Second, opts.useBlobPath)
+		fileCollector := file.NewFileCollector(ctx, opts.path, opts.poll, 30*time.Second, opts.useBlobURL)
 		err = collector.RegisterDocumentCollector(fileCollector, file.FileCollector)
 		if err != nil {
 			logger.Fatalf("unable to register file collector: %v", err)
@@ -95,13 +95,13 @@ you have access to read and write to the respective blob store.`,
 	},
 }
 
-func validateFilesFlags(pubsubAddr string, blobAddr string, poll bool, useBlobPath bool, args []string) (filesOptions, error) {
+func validateFilesFlags(pubsubAddr, blobAddr string, poll, useBlobURL bool, args []string) (filesOptions, error) {
 	var opts filesOptions
 
 	opts.pubsubAddr = pubsubAddr
 	opts.blobAddr = blobAddr
 	opts.poll = poll
-	opts.useBlobPath = useBlobPath
+	opts.useBlobURL = useBlobURL
 
 	if len(args) != 1 {
 		return opts, fmt.Errorf("expected positional argument for file_path")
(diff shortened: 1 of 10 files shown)


nchelluri:
I rebased this on top of https://github.com/guacsec/guac/pull/1807 in order to get that fix in so my tests will pass.

pkg/handler/collector/file/file_test.go
```diff
@@ -92,6 +112,7 @@ func Test_fileCollector_RetrieveArtifacts(t *testing.T) {
 				lastChecked: tt.fields.lastChecked,
 				poll:        tt.fields.poll,
 				interval:    tt.fields.interval,
+				useBlobPath: tt.fields.useBlobPath,
```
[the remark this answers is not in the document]
nchelluri: I looked more closely at this and decided that there was no change for me to make here.

1. Calling `NewFileCollector()` instead of instantiating the struct directly does not allow one to set `lastChecked`.
2. I think that if the error handler `eh()` does not get executed (and I think `eh()` usually or maybe always _does not_ get executed when the "with canceled poll" testcase is run) then the value of `wantErr` is not important. I am not sure how big a deal this is; I spent time trying to understand all of the execution and interaction here but after a while I didn't come to a conclusion and also decided it was unimportant. So, I will resolve this conversation.

cmd/guaccollect/cmd/files.go
```diff
@@ -81,7 +85,7 @@ you have access to read and write to the respective blob store.`,
 		logger := logging.FromContext(ctx)
 
 		// Register collector
-		fileCollector := file.NewFileCollector(ctx, opts.path, opts.poll, 30*time.Second)
+		fileCollector := file.NewFileCollector(ctx, opts.path, opts.poll, 30*time.Second, opts.useBlobURL)
```
pxp928, line 88: this will need to happen with the other collectors also. Either being called from `guaccollect` or `guacone` (just to be safe).

nchelluri: I am not 100% sure that I understand the request so I will ask for some clarification here.

1. I believe that when Jeff and I talked about the request he said to add it to the `guaccollect files` collector only, which is complete.
2. With `guacone collect files` I started to add it but I don't think it actually pushes the file to any blob store (I may be incorrect). So nothing to do here? (Perhaps the non-file collectors do add the file to the blob store bucket?)
3. I think that leaves deps_dev, gcs, git, github, oci, and s3 collectors. I believe you are asking me to provide the same command-line option to each of these collectors, and if the option is set, replace the node origin values with the blob store address instead of the original address?

pxp928: > With guacone collect files I started to add it but I don't think it actually pushes the file to any blob store (I may be incorrect). So nothing to do here? (Perhaps the non-file collectors do add the file to the blob store bucket?)


yes, correct `guacone` does not push to blob store. Sorry about that. 

> I think that leaves deps_dev, gcs, git, github, oci, and s3 collectors. I believe you are asking me to provide the same command-line option to each of these collectors, and if the option is set, replace the node origin values with the blob store address instead of the original address?

yes.

pxp928: We could do the other in another PR to close this one out.

nchelluri: Let's do it in another PR. I will work on that shortly.

jeffmendoza:
> I rebased this on top of #1807 in order to get that fix in so my tests will pass.

This is merged now, so rebase this pr.

cmd/guaccollect/cmd/files.go
```diff
@@ -186,5 +191,15 @@ func initializeNATsandCollector(ctx context.Context, pubsubAddr string, blobAddr
 }
 
 func init() {
+	set, err := cli.BuildFlags([]string{"use-blob-url"})
```
jeffmendoza, line 194: After the subsequent PRs that add this to other collectors, I imagine this will move to the top level command

jeffmendoza (approved):

cmd/guaccollect/cmd/files.go
pxp928: +1

pkg/handler/collector/git/git.go
```diff
@@ -47,7 +47,7 @@ type gitDocumentCollector struct {
 }
 
 func NewGitDocumentCollector(ctx context.Context, url string, dir string, poll bool, interval time.Duration) *gitDocumentCollector {
-	fileCollector := file.NewFileCollector(ctx, dir, false, time.Second)
+	fileCollector := file.NewFileCollector(ctx, dir, false, time.Second, false)
```
pxp928, line 50: In the subsequent PR, I would expect this also to be true.

cmd/guaccollect/cmd/files.go
```diff
@@ -44,6 +45,8 @@ type filesOptions struct {
 	blobAddr string
 	// poll location
 	poll bool
+	// use blob URL for origin instead of source URL
+	useBlobURL bool
```
pxp928, line 49: we should add to the comment here saying that if the blob store is persistent and we want to store the blob store source location.

Issue #1833, opened 2024-04-11
[feature] Add new field to store document reference for persistent storage


pxp928:
**Is your feature request related to a problem? Please describe.**
With the new addition of the blob store, GUAC now can store documents that were collected and ingested into GUAC. This will allow users to find the original document (or re-ingest in the case of failure) if they need. In order to accurately determine the location of a document in the blob store, a document reference (blob store key) needs to be stored in the DB.

Currently, replacing the `origin` with this information, as we did in https://github.com/guacsec/guac/pull/1811 would result in loss of the original location from where the document originated. We should add a new field to preserve the existing functionality.

This would require an update to the graphQL schema to add in a new field called `documentRef` that would be empty if the persistent blob store is not being used. We would also add a new field to the `SourceInformation` struct in `Document` to contain the new field:
```
type Document struct {
	Blob              []byte
	Type              DocumentType
	Format            FormatType
	Encoding          EncodingType
	SourceInformation SourceInformation
}
```
```
// SourceInformation provides additional information about where the document comes from
type SourceInformation struct {
	// Collector describes the name of the collector providing this information
	Collector string
	// Source describes the source which the collector got this information
	Source string
	// DocumentRef describes the location of the document in the blob store
	DocumentRef string
}
```

 We could either do this for all "verbs" or we can do it for only non-ephemeral documents like `hasSBOM`, `hasSLSA`, and `CertifyVEX`.

Though, in the future, we may want to move to a more in-toto attestation-specific ingestion for verbs like `certifyBad`, `certifyGood`, `hasSourceAt`, `pkgEquals` to name a few. So that we have evidence that something was attested by someone at this time. We again store this in the blob store for record keeping. 

**Describe the solution you'd like**
Update to the graphQL schema to add in a new field called `documentRef` that would be empty if the persistent blob store is not being used. 

Decision needs to be made if this should be done for all "verbs" or only a select few.

**Describe alternatives you've considered**

We could concatenate the strings together via a separator value (maybe something like https://github.com/guacsec/guac/blob/main/pkg/assembler/helpers/package.go#L27 so that it does not interfere in the future)?

For example:

`mcr.microsoft.com/oss/kubernetes/kubectl@sha256:8035089a59a6f8577255f494c1ced250e1206667d8462869fc0deeca98d79427guac-empty-@@sha256:8534561615616161894984126517`

That way we can split out the source in the future and get both the original source and the new blob store key back. 

[...]

PR #1834, opened 2024-04-11
Update graphQL schema to add documentRef field to all verbs

pxp928:
# Description of the PR

Update graphQL schema to add `documentRef` field to all verbs.

Subsequent PRs will include update to the `processor.Document`, parsers, and backend databases.

Updates the graphQL schema based on issue: https://github.com/guacsec/guac/issues/1833

# PR Checklist

### DIFF
--- a/pkg/assembler/graphql/schema/isDependency.graphql
+++ b/pkg/assembler/graphql/schema/isDependency.graphql
@@ -44,6 +44,8 @@ type IsDependency {
   origin: String!
   "GUAC collector for the document"
   collector: String!
+  "Reference location of the document in the persistent blob store (if that is configured)"
+  documentRef: String!
 }
 
 """
@@ -63,6 +65,7 @@ input IsDependencySpec {
   justification: String
   origin: String
   collector: String
+  documentRef: String
 }
 
 "IsDependencyInputSpec is the input to record a new dependency."
@@ -73,6 +76,7 @@ input IsDependencyInputSpec {
   justification: String!
   origin: String!
   collector: String!
+  documentRef: String!
 }
 
 extend type Query {
--- a/pkg/assembler/graphql/schema/isDependency.graphql
+++ b/pkg/assembler/graphql/schema/isDependency.graphql
@@ -45,7 +45,7 @@ type IsDependency {
   "GUAC collector for the document"
   collector: String!
   "Reference location of the document in the persistent blob store (if that is configured)"
-  documentRef: String!
+  documentRef: String
 }
 
 """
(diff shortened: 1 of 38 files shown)


pkg/assembler/graphql/schema/isDependency.graphql
```diff
@@ -44,6 +44,8 @@ type IsDependency {
   origin: String!
   "GUAC collector for the document"
   collector: String!
+  "Reference location of the document in the persistent blob store (if that is configured)"
+  documentRef: String!
```
mihaimaruseac, line 48: I think these should be `String` (and use `null` when location is not configured). Or do we want to special case `""` in code instead and require this all the time?

mihaimaruseac (approved):
LGTM, modulo the discussion on `String` vs `String!`

pkg/assembler/graphql/schema/isDependency.graphql
pxp928: For all the "verbs" or just the `isDependency` or `isOccur` ones?

mihaimaruseac: All verbs, I think we should be consistent

pxp928: in that case the `inputSpec` would also have to be just `String` for `documentRef`

mihaimaruseac: Yes. All will be optional

pxp928: hmm all the other `inputSpec` we require. I rather keep it consistent and make it an empty string `""`. What do you think?

mihaimaruseac: But then you would have to always create and pass an empty string in code to build the request and then on resolver side have a special check for `""` (we don't want to store empty strings in the db if we can avoid it, in general). For `String`, you only need to perform the check against `nil` in resolver, but client doesn't need to change if there is no document.

mihaimaruseac: Overall, I'm ok either way, but we should still have a discussion if we want to take the cost of specifying an additional parameter each time or not

pxp928: Makes sense, I can make it optional.

pxp928: Updated to make it optional `String` for both the `type` and `InputSpec`

mihaimaruseac (approved):

lumjjb (approved):

Issue #1846, opened 2024-04-16
[feature] The documentRef GraphQL field is populated by the collectors


nchelluri:
**Is your feature request related to a problem? Please describe.**
After closing https://github.com/guacsec/guac/issues/1833 we now have a `DocumentRef` field in the `SourceInformation` type that corresponds to the `documentRef` field in GraphQL verb nodes. The next (and last) step is to populate this field. As part of this work we also want to undo the overloading of the GraphQL `origin` field that was performed in the recently created `--use-blob-url` CLI flag for the File collector (done in https://github.com/guacsec/guac/pull/1811) and leave `origin` as it was prior that work and store the blob key only (and always) in the new `documentRef` field. I am also suggesting that we switch to using the term "blob key" as that seems to match the terminology used in the method signature here: https://pkg.go.dev/gocloud.dev/blob?utm_source=godoc#Bucket.NewWriter

This work is a follow-up to https://github.com/guacsec/guac/pull/1844. It is aimed to ease reingestion of documents.

**Describe the solution you'd like**
The collectors should populate the `SourceInformation.DocumentRef` field with the blob key (an example value for the field is `sha256:4c47f63847cd2b73d77f9008bb8bfa359e6201205004a2c26048a8711fd1c746`).

PR #1847, opened 2024-04-16
Populate SourceInformation.DocumentRef in collectors

nchelluri:
# Description of the PR

Fixes https://github.com/guacsec/guac/issues/1846

# PR Checklist


cmd/guaccollect/cmd/deps_dev.go
```diff
@@ -114,8 +114,18 @@ you have access to read and write to the respective blob store.`,
 	},
 }
 
-func validateDepsDevFlags(pubsubAddr string, blobAddr string, csubAddr string, csubTls bool, csubTlsSkipVerify bool, useCsub bool, poll bool, retrieveDependencies bool, args []string,
-	enablePrometheus bool, prometheusPort int,
+func validateDepsDevFlags(
```
nchelluri, line 117: There are some changes like this where in the full-diff view it looks like I made purely aesthetic changes but that is mostly not how it came about: originally I had added a flag to the collector commands and that required touching each one and I just tidied up a bit as I went. But when I was done I thought: maybe it make more sense to not over-complicate each of the collector commands by adding a flag to them and instead just always populate it. I ran this by @pxp928 and we discussed that this route was preferable because it left the code a bit simpler. In the in-memory backend we still populate the `documentRef` field to no real purpose, but it shouldn't hurt anything and allowed me to get rid of all of the new flags.

pkg/handler/collector/deps_dev/deps_dev_test.go
```diff
@@ -262,6 +263,15 @@ func Test_depsCollector_RetrieveArtifacts(t *testing.T) {
 				t.Errorf("Wanted %v elements, but got %v", len(tt.want), len(collectedDocs))
 			}
 			for i := range collectedDocs {
+				// Why do we do this here instead of in the test case definition?
```
pxp928, line 266: `normalizeTimeStampAndScorecard` is done such that the compare passes for the test (as the timestamps maybe different)

pxp928: Not sure if I answered your question. Let me know and we can discuss.

pkg/handler/collector/github/github_test.go
```diff
@@ -282,6 +285,9 @@ func Test_githubCollector_RetrieveArtifacts(t *testing.T) {
 	mockClient := &MockGithubClient{}
 	mockData := mockDataSource()
 
+	// TODO: Currently, len(collectedDocs) == 0. Fix this, and implement a
```
pxp928, line 288: FYI @nathannaveen

pkg/handler/collector/s3/s3.go
```diff
@@ -89,8 +90,9 @@ func retrieve(s S3Collector, ctx context.Context, docChannel chan<- *processor.D
 			Format:   processor.FormatUnknown,
 			Encoding: bucket.ExtractEncoding(enc, item),
 			SourceInformation: processor.SourceInformation{
-				Collector: S3CollectorType,
-				Source:    "S3",
+				Collector:   S3CollectorType,
+				Source:      "S3",
```
pxp928, line 94: This source is wrong. Not something you did in this PR but has been wrong for some time. This should be `item`

pkg/handler/collector/s3/s3.go
```diff
@@ -124,8 +126,9 @@ func retrieve(s S3Collector, ctx context.Context, docChannel chan<- *processor.D
 					Format:   processor.FormatUnknown,
 					Encoding: bucket.ExtractEncoding(enc, item),
 					SourceInformation: processor.SourceInformation{
-						Collector: S3CollectorType,
-						Source:    "S3",
+						Collector:   S3CollectorType,
+						Source:      "S3",
```
pxp928, line 130: same here

pkg/handler/collector/s3/s3.go
```diff
@@ -214,8 +217,9 @@ func retrieveWithPoll(s S3Collector, ctx context.Context, docChannel chan<- *pro
 						Format:   processor.FormatUnknown,
 						Encoding: bucket.ExtractEncoding(enc, item),
 						SourceInformation: processor.SourceInformation{
-							Collector: S3CollectorType,
-							Source:    "S3",
+							Collector:   S3CollectorType,
+							Source:      "S3",
```
pxp928, line 221: and here

pxp928 (approved):
Overall looks good! Thanks

pkg/handler/collector/s3/s3.go
nchelluri: Fixed in b46408d5e5c9db78f79cb7da7c5b9962015c6df5.

pkg/handler/collector/deps_dev/deps_dev_test.go
nchelluri: Fixed in e114cf3e688203b1eecd3c300cce9983b940c605. I actually think we should use a better comparison mechanism like `github.com/stretcher/testify/assert` but that is a discussion for another day.

lumjjb (approved):

PR #1864, opened 2024-04-24
Add a client testing library

mdeicas:
In writing a REST API transitive dependencies endpoint, I found I needed an easier way to write tests. This PR is my solution for that, but I also think that more generally testing clients of the Graphql server isn't currently well-supported and this PR can address that.

## How clients are currently tested
One approach that has been taken is to start a new http server for each test and to specify the graph with the [`IngestPredicates`](https://github.com/guacsec/guac/blob/main/pkg/assembler/assembler.go#L32) type. Using `IngestPredicates` to specify the graph is appealing because all of the ingestion code is already implemented in the [assembler function](https://github.com/guacsec/guac/blob/main/pkg/assembler/clients/helpers/bulk.go#L30). However, this leads to very verbose tests because each noun must be specified with its corresponding InputSpec every time it is referenced. Additionally, the verb InputSpecs (e.g. HashEqualInputSpec) are required, but their content is often irrelevant to tests. See the [patch planning tests](https://github.com/guacsec/guac/blob/main/pkg/guacanalytics/patchPlanning_test.go) for where this approach of test setup is used, and that about 900 LOC are used to define 12 graphs.  Verbose tests or tests with duplicate code aren't necessarily bad, but I think this approach does limit readability quite a bit and makes writing tests more difficult. It also requires specifying data that may not be relevant to a test, such as the verb InputSpecs. 

Another approach that has been taken (by https://github.com/guacsec/guac/pull/1705) is to rely on the graphql server that the CI starts during the integration test phase and to create helper functions and types (e.g. [shortcut type](https://github.com/guacsec/guac/blob/cf32140f273dac8b70a4a9fbc506e9d2ccc9191d/pkg/dependencies/dependents_test.go#L39) for IsDependency in that PR) to make it easier and less verbose to specify the graph. However, using the Graphql server that is global to the CI integration tests isn't ideal.

## Description of PR

I think that there should be centralized support to test clients in order to avoid reinventing the wheel and to make testing easier. This PR adds a client testing library that provides a concise way to specify the Guac graph for ingestion (generalizing the approach taken by https://github.com/guacsec/guac/pull/1705), where all nouns can be referenced by a single string and the verb InputSpecs are optional. Additionally, it adds some functions to start up the graphql server for tests. 

Changes to this library should scale linearly with changes / additions to the GQL API. I didn't implement all nouns and verbs, but since this is a testing library I think they can be added as needed. 

See [example use](https://github.com/mdeicas/guac/blob/31267311c89969d49345d803d6d1357e05fb8c4f/pkg/guacrest/server/retrieveDependencies_test.go#L54)

## Alternatives

Another approach to testing clients would be to stub the Graphql client API. To my understanding, this would support smallish unit tests, but does not support end-to-end tests of functionality that makes many, varied calls to the graphql server. 

Yet another approach would be to support defining the Guac graph in a concise way, as this library provides, but then map that representation to the `IngestPredicates` type and use the assembler function, instead writing the ingestion code here. However, that the `HasSbomIncludesSpec` only takes IDs complicates this a bit, and also slightly limits what graphs can be defined.

# PR Checklist

### DIFF
--- a/internal/testing/graphqlClients/guacdata.go
+++ b/internal/testing/graphqlClients/guacdata.go
@@ -1,0 +1,441 @@
+//
+// Copyright 2024 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package clients
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/Khan/genqlient/graphql"
+	_ "github.com/guacsec/guac/pkg/assembler/backends/keyvalue"
+	gql "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"github.com/guacsec/guac/pkg/assembler/helpers"
+)
+
+const (
+	// nouns
+	defaultHashAlgorithm   = "sha256"
+	defaultSourceType      = "test-type"
+	defaultSourceNamespace = "test-namespace"
+
+	// IsOccurrence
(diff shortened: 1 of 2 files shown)


pxp928 (approved):
Awesome. This will help with the testing!

mdeicas:
Closing because this was included in #1867 so it has already been merged.

PR #1867, opened 2024-04-26
Add a transitive dependencies endpoint to the REST API

mdeicas:
# Description of the PR

Adds an endpoint to the REST API to get the transitive dependencies of an artifact or package. 

This PR is dependent on and duplicates the changes from #1864 (everything in `internal/testing`). Once that is merged, I'll rebase this one.

### Some design decisions taken were:
- The graph is traversed using the GraphQL dependencies endpoint
- We assume that all verbs used in the traversal attach to package version nodes, not package name nodes.
- Errors that are not actionable to the client are not surfaced. Instead, they are logged and a generic error is returned. 

### Other things done on this PR are:
- adding HashEqual and IsOccurrence query to gql client
- updated the pagination library to take a pointer to the pagination spec, in order to move the checking for nil and applying the default into the library 
- The logger is attached to incoming requests by HTTP middleware, log statements now contain the corresponding HTTP request method and path, and each successful request is logged.

### Explanation of the `linkCondition` parameter
This parameter intends to configure the accuracy of the results. If `digest` is specified, predicates attached to packages are not used for the graph traversal, because this can lead inconsistencies. For example, suppose there is an artifact A1 that is an occurrence of package P, and there is an SBOM attached to P. It may be misleading to use that SBOM in finding the dependencies of A1, because that SBOM may have been generated for some other artifact A2 -- we don't know. 

Asking for accuracy by setting `linkCondition = digest` requires good data (e.g. SBOMs need to provide the digest of the subject, which I believe Syft does not). In the absence of this, predicates attached to packages can be used for the grpah traversal by setting `linkCondition = name`. 

# PR Checklist


jeffmendoza (approved):
This is really well laid out and high quality. Thanks Marco!

Probably the design decisions and/or algorithm may be tweaked as folks use this and have feedback, but that can be iterated on.

mihaimaruseac (approved):
Thank you, Marco!

pkg/guacrest/server/retrieveDependencies.go
```diff
@@ -0,0 +1,394 @@
...
+	// Nodes equivalent to the start node are not dependencies
+	for node := range nodesEquivalentToStart {
+		delete(visited, node.GetId())
+	}
+	return maps.Values(visited), nil
+}
+
+/********* Implementations of the interface *********/
+
+func (eg byDigest) getDirectDependencies(ctx context.Context, v node) ([]node, error) {
```
lumjjb, line 131: very nice abstractions!

internal/testing/graphqlClients/testsetup.go
```diff
@@ -0,0 +1,101 @@
...
+
+	closeFunc := func() {
+		err := srv.Close()
+		if err != nil {
+			t.Logf("Error closing graphql server listener")
+		} else {
+			t.Logf("Graphql server shut down")
+		}
+	}
+	t.Cleanup(closeFunc)
```
lumjjb, line 84: nice!

PR #1832, opened 2024-04-09
Update error handling on ingestion

jeffmendoza:
Bulk emitter will return errors now. It continues to try to ingest all verbs even if one fails. It stops ingestion on a single noun failure. Guacone files command will log all errors and continue to ingest all files found even if one errors. Summary should list errors. Guacingest service will log errors and continue, it will ack messages even if errored.

# Description of the PR

# PR Checklist

- [ ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [ ] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If OpenAPI spec is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #1874, opened 2024-04-29
[FIX] Ingestor should not ack message on failure

pxp928:
# Description of the PR

return error from emitter such that if ingestion fails and message is not ack. 

Before it was being logged and that resulted in the error being acked. 

# PR Checklist

### DIFF
--- a/cmd/guacingest/cmd/ingest.go
+++ b/cmd/guacingest/cmd/ingest.go
@@ -17,8 +17,10 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):

cmd/guacingest/cmd/ingest.go
```diff
@@ -95,7 +95,7 @@ func ingest(cmd *cobra.Command, args []string) {
 
 	emit := func(d *processor.Document) error {
 		if err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, transport, csubClient); err != nil {
-			logger.Errorf("unable to ingest document %q : %v", d.SourceInformation.Source, err)
+			return fmt.Errorf("unable to ingest document %q : %v", d.SourceInformation.Source, err)
```
jeffmendoza, line 98: This undoes the change in https://github.com/guacsec/guac/pull/1832 where we want to ack and log an error if there is an error in parsing the document so that we don't leave a bad document in the pubsub to be re-ingested forever.

jeffmendoza:
We need to differentiate between recoverable errors where we want to retry and unrecoverable errors where we want to dequeue the message.

cmd/guacingest/cmd/ingest.go
pxp928: hmm, should we add a health check for the graphQL server to see if it is down and use the error to determine that it is a "recoverable error".

pxp928: fixed!

jeffmendoza (approved):

PR #1886, opened 2024-05-03
Document that the Equals predicates equate only two nouns

mdeicas:
# Description of the PR
The current documentation seems to indicate that a single HashEqual, PkgEqual, or VulnEqual predicate may equate more than two nouns.

# PR Checklist


mihaimaruseac (approved):

mihaimaruseac:
We could also split it to have one field for each noun instead of a list, but that's a more involved change

pxp928:
> We could also split it to have one field for each noun instead of a list, but that's a more involved change

@mihaimaruseac that is what we technically do in the backend implementation: https://github.com/guacsec/guac/blob/main/pkg/assembler/backends/ent/schema/hashequal.go#L38-L39

pxp928 (approved):

mihaimaruseac:
Yes, but then we can enforce this from the GQL schema

pxp928:
> Yes, but then we can enforce this from the GQL schema

yup I agree.

PR #1932, opened 2024-05-28
Support direct connections to ent from the rest api

mdeicas:
# Description of the PR

Adds configuration to allow endpoints from the REST API to connect directly to Ent. This will allow bypassing the GraphQl API to use database native functionality.

This option can be enabled with `guacrest --db-direct-connection`. Current endpoint implementations can be overridden to use this optimized implementation by adding methods to `EntConnectedServer`. If `--db-direct-connection` is specified but an endpoint does not have a corresponding `EntConnectedServer` implementation, then the default one will be used.

# PR Checklist

### DIFF
--- a/cmd/guacrest/cmd/server.go
+++ b/cmd/guacrest/cmd/server.go
@@ -27,6 +27,8 @@ import (
 
 	"github.com/Khan/genqlient/graphql"
 	"github.com/go-chi/chi"
+	"github.com/guacsec/guac/pkg/assembler/backends/ent"
+	"github.com/guacsec/guac/pkg/assembler/backends/ent/backend"
 	"github.com/guacsec/guac/pkg/cli"
 	gen "github.com/guacsec/guac/pkg/guacrest/generated"
 	"github.com/guacsec/guac/pkg/guacrest/server"
@@ -40,7 +42,7 @@ func startServer() {
 	httpClient := &http.Client{Transport: cli.HTTPHeaderTransport(ctx, flags.headerFile, http.DefaultTransport)}
 	gqlClient := getGraphqlServerClientOrExit(ctx, httpClient)
 
-	restApiHandler  := gen.Handler(gen.NewStrictHandler(server.NewDefaultServer(gqlClient), nil))
+	restApiHandler := gen.Handler(gen.NewStrictHandler(getRestApiHandlerOrExit(ctx, gqlClient), nil))
 
 	router := chi.NewRouter()
 	router.Use(server.AddLoggerToCtxMiddleware, server.LogRequestsMiddleware)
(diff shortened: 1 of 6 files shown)


pxp928:
Very nice! @mdeicas

cmd/guacrest/cmd/server.go
```diff
@@ -90,6 +96,37 @@ func startServer() {
...
+		return handler
+	}
+	return server.NewDefaultServer(gqlClient)
+}
+
+func getEntClientOrExit(ctx context.Context) *ent.Client {
+	logger := logging.FromContext(ctx)
+
+	if flags.dbDriver != dialect.Postgres {
+		logger.Warnf("Only Postgres is currently supported")
```
lumjjb, line 116: nit: start with lowercase on messages, could also use `Fatalf` here, probably would be and error level log than a warn level log since it is a failure which prevents it from running.

pkg/cli/store.go
```diff
@@ -85,6 +85,7 @@ func init() {
 	set.String("rest-api-server-port", "8081", "port to serve the REST API from")
 	set.String("rest-api-tls-cert-file", "", "path to the TLS certificate in PEM format for rest api server")
 	set.String("rest-api-tls-key-file", "", "path to the TLS key in PEM format for rest api server")
+	set.Bool("db-direct-connection", false, "connect directly to the database that backs the gql API for optimized endpoint implementations")
```
lumjjb, line 88: hmm (ignorant) question on maintaining this separate direct db connection - how will this change with graphQL changes and ent usage? will we have to maintain 2 separate implementations?

mihaimaruseac:
Hmm, won't this create opportunities for the two backends to get out of sync? You'll now be able to write data that would not pass graphql ontology validations.

mdeicas:
> Hmm, won't this create opportunities for the two backends to get out of sync? You'll now be able to write data that would not pass graphql ontology validations.

Good point, that's true. This *should* only be for read access to the database, though. I think we can enforce this by adding a hook to the ent client to reject all writes, and also generally through code reviews.

mihaimaruseac:
If we make it only read queries that works for me

mdeicas:
@mrizzi @dejanb for visibility and any thoughts, I think you were interested in this!

mihaimaruseac (approved):
Thank you!

pkg/cli/store.go
lumjjb: +1 on @mihaimaruseac 's suggestion on read-only.. One thing i'd want to add is perhaps an experimental comment on this feature, unless we all vote on this being a supported path going forward. I can see the cost of maintaining this feature being significant in sync and not drifting. I'd put a warning on the use of the feature.

If we do decide that we want to support this then we have to add it to the PR template that all changes need to be checked here as well. (which i don't think we are deciding yet).... 

tl;dr; i'd lgtm if we add experimental and have a warning banner when users use this. Otherwise we should discuss this at the maintainers meeting.

mdeicas: I agree, added a log and modified the flag description to reflect this!

lumjjb (approved):

PR #1933, opened 2024-05-29
remove GetMatchFlagsFromPkgInput helper as it was not needed for isDependency

pxp928:
# Description of the PR

GetMatchFlagsFromPkgInput was left as a remnant when isDependency could be created at the pkgName level. Changes were madeto the SPDX and CDX parser to create isDependency only at the pkgVersion level. SBOMs in general will always specify a version and not a range. Deps.dev also resolves to a version.

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/spdx/parse_spdx_test.go
+++ b/pkg/ingestor/parser/spdx/parse_spdx_test.go
@@ -517,7 +517,7 @@ func Test_spdxParser(t *testing.T) {
 					{
 						Pkg:             pUrlToPkgDiscardError("pkg:guac/spdx/testsbom"),
 						DepPkg:          pUrlToPkgDiscardError("pkg:guac/files/sha1:ba1c68d88439599dcca7594d610030a19eda4f63?filename=./include-file"),
-						DepPkgMatchFlag: generated.MatchFlags{Pkg: generated.PkgMatchTypeAllVersions},
+						DepPkgMatchFlag: generated.MatchFlags{Pkg: generated.PkgMatchTypeSpecificVersion},
 						IsDependency: &generated.IsDependencyInputSpec{
 							DependencyType: generated.DependencyTypeUnknown,
 							Justification:  "top-level package GUAC heuristic connecting to each file/package",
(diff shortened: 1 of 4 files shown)


pxp928:
@lumjjb let me know your thoughts on this.

pkg/ingestor/parser/spdx/parse_spdx_test.go
```diff
@@ -517,7 +517,7 @@ func Test_spdxParser(t *testing.T) {
 					{
 						Pkg:             pUrlToPkgDiscardError("pkg:guac/spdx/testsbom"),
 						DepPkg:          pUrlToPkgDiscardError("pkg:guac/files/sha1:ba1c68d88439599dcca7594d610030a19eda4f63?filename=./include-file"),
-						DepPkgMatchFlag: generated.MatchFlags{Pkg: generated.PkgMatchTypeAllVersions},
+						DepPkgMatchFlag: generated.MatchFlags{Pkg: generated.PkgMatchTypeSpecificVersion},
```
lumjjb, line 520: ah... i think in this case, it is set to `AllVersions` since we do not actually version files... 

see my comment in https://docs.google.com/document/d/11WqkncYYob8MtNkcvTZiYcjbvclT15UKFh6coDjJToU/edit, on "At the same time, propose changing files to have the similar format of: `guac/generic/file/sha256@…`, so that they can be referenced via their specific version."

pxp928: I will work to change the ways the files purls are created and handled.

pxp928: check how the file purl is being parsed.

pxp928: digest should not be part of the version to ensure any star graphs.

pxp928: @lumjjb just to clarify here the digest should not be part of the version? We would need to for isDep to work only at the pkgVersion level?

lumjjb: i think based on our discussion, there will never be a version here, BUT it should be ok since it defaults to no version being the empty version, so to work towards not using "AllVersions" where possible, this is fine.

lumjjb (approved):

pkg/ingestor/parser/spdx/parse_spdx_test.go
pxp928: Ok. The current behavior is that `files` do not have a version (which will result in an empty version string). So this is all good.

pxp928: For example: `"pkg:guac/files/sha1:ba1c68d88439599dcca7594d610030a19eda4f63?filename=./include-file"` 

will create:

Type - `guac`
Namespace - `files`
name - `sha1:ba1c68d88439599dcca7594d610030a19eda4f63?`
version - `""`
qualifier - `"filename"` = `"./include-file"`
subpath - `""`

pxp928:
Once this PR merges, I will create a follow on PR to remove this functionality from graphQL and the backends. Meaning that `isDep` will only have a `pkgVersion` -> `pkgVersion` relationship.

mihaimaruseac (approved):

PR #1938, opened 2024-05-30
Including the Scorecard API

### LINKED ISSUE #1892: [feature] Provide Option to Use OpenSSF Scorecard REST API for Scorecard Ingestion
## Background

Currently, GUAC is calculating the OpenSSF Scorecard scores directly using the Scorecard library. This requires having a GitHub token to access the repository data. However, there are cases where we may not have access to a GitHub token, such as when ingesting an SBOM without the associated repository information.

## Proposal

To address this, we propose adding an option to use the OpenSSF Scorecard REST API (https://github.com/ossf/scorecard?tab=readme-ov-file#scorecard-rest-api) to retrieve the Scorecard scores, in addition to the existing library-based approach.

## Key Features

1. User Option: Provide a configuration option or flag that allows users to choose between the following two Scorecard ingestion modes:
    1. Library-based Scorecard: Use the Scorecard library to calculate the scores directly, requiring a GitHub token.
    2. REST API-based Scorecard: Use the OpenSSF Scorecard REST API to retrieve the scores, without requiring a GitHub token.
2. Preferred Scorecard Source: When the REST API-based mode is enabled, GUAC should first attempt to calculate the Scorecard score using the library-based approach. If successful, this score should be used as the most recent and default score. If the library-based approach fails (e.g., due to missing GitHub token), GUAC should then fall back to using the REST API to retrieve the Scorecard score.
3. Scorecard History: When using the REST API-based mode, GUAC should also store the timeline of Scorecard scores for each project, allowing users to track the security posture over time.

## Key Benefits

1. Expanded Scorecard Coverage: The REST API-based mode allows GUAC to retrieve Scorecard data for a wider range of repositories, even when GitHub API tokens are not available. This significantly expands the set of SBOMs that can be ingested with Scorecard data.
2. No GitHub Token Required: The REST API-based mode does not require a GitHub token, which removes a potential barrier to accessing Scorecard data.
3. Scorecard History: The REST API provides a timeline of Scorecard scores, which can be useful for tracking the security posture of a project over time.

## Proposed Implementation

1. Add a new configuration option or flag to allow users to choose between the library-based and REST API-based Scorecard ingestion modes.
2. Implement a new Scorecard provider that uses the OpenSSF Scorecard REST API to fetch the Scorecard data.
3. Update the existing Scorecard ingestion logic to support both the library-based and REST API-based approaches, with the preferred source being the library-based score if available.
4. Store the Scorecard score history when using the REST API-based mode.
5. Document the new Scorecard ingestion options and their tradeoffs in the GUAC documentation.

neilnaveen:
# Description of the PR

- Fixes #1892
- Updated tests
- Added a README for certifier/scorecard

# PR Checklist


pxp928 (commented):
@neilnaveen looks like you have a small lint issue.

cmd/guacone/cmd/scorecard.go
```diff
@@ -204,4 +203,6 @@ func validateScorecardFlags(
 
 func init() {
 	certifierCmd.AddCommand(scorecardCmd)
+	scorecardCmd.Flags().Bool("use-scorecard-api", false, "use the scorecard API")
+	viper.BindPFlag("use-scorecard-api", scorecardCmd.Flags().Lookup("use-scorecard-api"))
```
nathannaveen, line 207: Could you add the flags in a similar manner to: https://github.com/guacsec/guac/blob/8b702d02d6fc63b263f19de5ef1cc835e722ac78/cmd/guacone/cmd/vulnerability.go#L628-L637

nathannaveen (commented):
Thanks for doing this, LGTM, except for a small nit.

cmd/guacone/cmd/scorecard.go
nathannaveen: Doing this should also fix the lint.

neilnaveen: Thanks for the review. I have fixed the problem!

pkg/certifier/scorecard/README.md
```diff
@@ -0,0 +1,29 @@
+# Scorecard Certifier
```
lumjjb, line 1: nice addition!

lumjjb (changes requested):

pxp928:
ping to keep it from closing.

PR #1997, opened 2024-07-02
[ENT] #1996 Improve package's qualifiers query

### LINKED ISSUE #1996: [bug] [ENT] HasMetadata ingestion fails due to package's qualifiers
**Describe the bug**
HasMetadata ingestion fails due to package's qualifiers mismatch when there are two packages with all the same fields and whose qualifiers are one a superset of the other.

**To Reproduce**

1. package A `pkg:maven/io.github.crac/org-crac@0.1.1.redhat-00002?repository_url=https://maven.repository.redhat.com/ga/&type=jar` is ingested
2. package B `pkg:maven/io.github.crac/org-crac@0.1.1.redhat-00002?type=jar` is ingested
3. ingestion of HasMetadata for package B  fails with the error:
    ```
    failed to execute IngestHasMetadata :: input: generateDependencyCreate :: getPkgVersion :: ent: package_version not singular
    ```

**Expected behavior**
Ingestion of HasMetadata for package B  should run successfully.

**Screenshots**

**GUAC version**
latest `main` (73061939a8baba217dd7ab9b2486a9e4ffc460c7)

**Additional context**
It works fine with `memmap` backend

mrizzi:
# Description of the PR

The issue's due to the fact that, when adding an `HasMetadata` node, there's a uniqueness check for the package specs to identify a single package.

In the scenario described in  #1996, both packages
`pkg:maven/io.github.crac/org-crac@0.1.1.redhat-00002?repository_url=https://maven.repository.redhat.com/ga/&type=jar` 
and
`pkg:maven/io.github.crac/org-crac@0.1.1.redhat-00002?type=jar`
have been already ingested.
The `QualifiersMatch` method checks for the provided qualifiers into the DB, i.e. `type=jar`, hence finding both the packages satisfy this condition and hence generating the error because the `HasMetadata` ingestion requires the package to be uniquely identified.

The fix adds a further check that, after having checked all of input package qualifiers are available, it also checks the size of the array of the qualifiers to ensure package with a superset of the input qualifiers, e.g. `repository_url=https://maven.repository.redhat.com/ga/`, aren't considered as a valid matching package.

Fixes #1996 

# PR Checklist

### DIFF
--- a/internal/testing/testdata/models.go
+++ b/internal/testing/testdata/models.go
@@ -546,3 +546,17 @@ var P6out = []*model.Package{{
 		}},
 	}},
 }}
+
+var P7 = &model.PkgInputSpec{
+	Type:      P5.Type,
+	Namespace: P5.Namespace,
+	Name:      P5.Name,
+	Version:   P5.Version,
+	Qualifiers: []*model.PackageQualifierInputSpec{{
+		Key:   P5.Qualifiers[0].Key,
+		Value: P5.Qualifiers[0].Value,
+	}, {
+		Key:   "repository_url",
+		Value: "https://alternative.report.url/",
+	}},
+}
--- a/internal/testing/backend/hasMetadata_test.go
+++ b/internal/testing/backend/hasMetadata_test.go
@@ -761,6 +761,40 @@ func TestHasMetadata(t *testing.T) {
 					DocumentRef: "test",
 				},
 			},
+		}, {
+			Name:  "Two packages with superset qualifiers",
+			InPkg: []*model.PkgInputSpec{testdata.P5, testdata.P7},
+			Calls: []call{
+				{
+					Sub: model.PackageSourceOrArtifactInput{
+						Package: &model.IDorPkgInput{PackageInput: testdata.P5},
+					},
+					Match: &model.MatchFlags{
+						Pkg: model.PkgMatchTypeSpecificVersion,
+					},
+					HM: &model.HasMetadataInputSpec{
+						Key:           "key1",
+						Value:         "value1",
+						Timestamp:     time.Unix(1e9, 0),
+						Justification: "another test justification",
+					},
+				},
+			},
+			Query: &model.HasMetadataSpec{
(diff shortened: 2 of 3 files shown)


pxp928 (dismissed):
LGTM! Thanks @mrizzi

internal/testing/testdata/models.go
```diff
@@ -546,3 +546,17 @@ var P6out = []*model.Package{{
...
+var P7 = &model.PkgInputSpec{
+	Type:      P5.Type,
+	Namespace: P5.Namespace,
+	Name:      P5.Name,
+	Version:   P5.Version,
+	Qualifiers: []*model.PackageQualifierInputSpec{{
+		Key:   P5.Qualifiers[0].Key,
+		Value: P5.Qualifiers[0].Value,
+	}, {
+		Key:   "repository_url",
```
pxp928, line 559: Based on the helper function defined here: https://github.com/guacsec/guac/blob/main/pkg/assembler/helpers/purl.go#L176-L223

There would not be a `repository_url` qualifier to begin with. So does this usecase still come up?

pxp928: Ah this is not an `image` so the helper function does not apply. In that case this would be required.

pxp928 (approved):
LGTM based on the usecase!

internal/testing/testdata/models.go
mrizzi: Well, the issue doesn't refer to a specific `key` for qualifier but whenever a package's qualifiers are a superset of another package (with all the other fields being the same).

In any case, also for the specific test data key `repository_url`, it still applies to make package types, e.g. `TypeMaven`, because it's removed only for `TypeOCI` and `TypeDocker` types AFAIU.

pxp928: Yes, that makes sense. I saw the `repository_url` and it re-triggered me to re-review. False alarm.

mihaimaruseac (approved):

mihaimaruseac:
Thank you!

Issue #2058, opened 2024-08-01
[feature] Questions regarding adding REST Endpoints for Vulnerability and Legal info in an SBOM


nathannaveen:
When talking about GUAC, a common issue that pops up is that it's hard to find your vulnerabilities or legal information in a specific SBOM. Currently, to find this, the user would have to make multiple different graphql calls which is cumbersome, to solve this we could use a couple REST endpoints.

Since this comes up often, I would like to start working on it. But, before I do, I'd like to get some feedback from the GUAC community on a few of my questions.

This feature is probably going to be made up of three REST endpoints.

1. The first endpoint would be to find the latest SBOM for a specific package or artifact. 
2. The second endpoint would be to find the vulnerabilities in the latest SBOM for a package or artifact.
3. The third endpoint would be to find the licenses in the latest SBOM for a package or artifact.

***

Here are some of my questions regarding this feature:

1. When searching for the latest SBOM for a given package or artifact, do we want to search for the package or artifact via `id` or the spec?
2. To find the latest SBOM, I was thinking of comparing the SBOMs via their version. If both of them don't have a version, we compare them via their timestamp. However, since we can search via artifact as well as package, we need to modify this thought process because artifacts don't have versions. Therefore, we would have to compare all artifacts using their timestamps. Does this approach make sense, or is there a better way to handle this?
3. When querying for vulnerabilities in the latest SBOM, should we include transitive dependencies or only direct dependencies? Or should we make it a flag to search all dependencies? Also note that with a big enough use case searching through transitive dependencies for vulnerabilities could possibly take a very long time.

funnelfiasco:
What's the use case for searching a single SBOM versus the entire graph? One of the nice things about GUAC is the fact that it operates across multiple SBOMs at once, so I'm not sure why we'd want to look at a single SBOM.

> When querying for vulnerabilities in the latest SBOM, should we include transitive dependencies or only direct dependencies? Or should we make it a flag to search all dependencies? Also note that with a big enough use case searching through transitive dependencies for vulnerabilities could possibly take a very long time.

I think being able to choose between the two is good. The more indirect the dependency, the harder it will be to remediate, but people will still want to know about it. The performance issue is a concern, though, so having a way to specify a max depth (or a simple direct-dependencies-only) would be good for those larger instances.

jeffmendoza:
>What's the use case for searching a single SBOM versus the entire graph? 

[...]

PR #2064, opened 2024-08-05
Find Latest SBOM for a Package or Artifact

nathannaveen:
# Description of the PR

When talking about GUAC, a common issue that pops up is that it's hard to find your vulnerabilities or legal information in a specific SBOM. Currently, to find this, the user would have to make multiple different graphql calls which is cumbersome, to solve this we could use a couple REST endpoints.

This PR finds the latest SBOM for a specific package or artifact and outputs it via REST. We can then find vulnerability and legal information based on the latest SBOM. In future PRs there will be REST endpoints for vulnerabilities in the latest SBOM and licenses in the latest SBOM.

This is based on the design ideas in https://github.com/guacsec/guac/issues/2058.

# PR Checklist


pkg/guacrest/openapi.yaml
```diff
@@ -99,7 +99,30 @@ paths:
           $ref: "#/components/responses/InternalServerError"
         "502":
           $ref: "#/components/responses/BadGateway"
-
+  "/query/latest-sbom":
+    get:
+      summary: Retrieve the latest SBOM for a given package
+      operationId: findLatestSBOM
+      parameters:
```
mdeicas, line 106: I'm not sure if I agree that searching by id is the best choice here:
- it requires a previous call to the GQL API (and an understanding of the package trie) which is partly what the REST API intends to avoid.
- I think it makes sense to be consistent across the REST API, and the other endpoint doesn't expose the internal identifiers

I also don't think that searching for SBOMs by package is the right approach:
- SBOMs are generated for and should be associated with artifacts, not packages (https://blog.deps.dev/zillions-of-sboms/). In some cases the data quality isn't there and they will be associated with packages, but I think it makes sense to have the interface work with the "right" way. 
- this interface hides many details -- does it mean the latest SBOM associated with the package, or the latest SBOM associated with the latest artifact occurrence of the package, or the latest among those two? 


IMO the best interface is with a digest. It is precise and subsumes searching for an SBOM by package.

pkg/guacrest/helpers/sbom.go
```diff
@@ -0,0 +1,154 @@
...
+	model "github.com/guacsec/guac/pkg/assembler/clients/generated"
+	"go.uber.org/zap"
+	"strings"
+
+	"github.com/Khan/genqlient/graphql"
+	"github.com/Masterminds/semver"
+	"github.com/guacsec/guac/pkg/logging"
+)
+
+func LatestSBOMForAGivenId(ctx context.Context, client graphql.Client, id string) (*model.AllHasSBOMTree, error) {
```
mdeicas, line 30: Can you write tests for this? See the tests in `pkg/guacrest/server/retrieveDependencies_test.go` as a reference.

mdeicas (changes requested):

pkg/guacrest/openapi.yaml
pxp928: @mdeicas did you see the design issue: https://github.com/guacsec/guac/issues/2058?

mdeicas: I did - I agree that it shouldn't be the role of these endpoints to "help you look for software", that the interface should identify the inputs precisely. But I guess I don't see what benefits there are to Guac ids in the interface. On the other hand, digests and purls (a purl as an identifier for a node in Guac), which are also precise, avoid exposing the internal identifiers.

nathannaveen:
@mdeicas I will be taking the feedback from your review here into the new PR #2089.

nathannaveen:
Closing this PR, as #2089 has a better design.

PR #2080, opened 2024-08-16
remove daysSinceLastScan as it is redundant with certifier interval

### LINKED ISSUE #2077: [ingestion/clearlydefined] clearlydefined fails to run when the osv certifier has already run
**Describe the bug**
 clearlydefined fails to run when the osv certifier has already run. This occurs because root package query checks if certifyVulns are already collected and skips the package. This should not happen when the clearly defined certifier is run. 
 
 Question is, should this check be removed? As we want the certifiers to run on OSV and clearlydefined on various intervals, re-adding new vulns or licenses as they are found. The check was only added to prevent the certifier from running again manually and adding duplicate vuln nodes. This, in the long term, is actually what we want to happen in GUAC as we know that the information is up-to-date.

 
**To Reproduce**

1. `go run ./cmd/guacgql --gql-debug --gql-backend=ent`
2. `go run ./cmd/guacone collect files spdx_vuln.json`
3. `go run ./cmd/guacone certifier osv`
4. `go run ./cmd/guacone certifier cd`

You will notice that it completes without adding any licenses.

**Expected behavior**
Add all the license information based on clearly defined.

**GUAC version**
v0.8.0

pxp928:
# Description of the PR

fixes #2077 

# PR Checklist

### DIFF
--- a/internal/testing/e2e/e2e
+++ b/internal/testing/e2e/e2e
@@ -148,8 +148,6 @@ for command in "${ingestion_commands[@]}"; do
   echo @@@@ Waiting
   sleep 60
 
-  go run ${GUAC_DIR}"/cmd/guacone" certifier osv -p=false
-
   if [ $? -ne 0 ]; then
     echo "Error: Command '$command' failed" | tee -a "$LOG_FILE"
     exit 1
--- a/cmd/guaccollect/cmd/license.go
+++ b/cmd/guaccollect/cmd/license.go
@@ -45,9 +45,6 @@ type cdOptions struct {
 	interval time.Duration
 	// enable/disable message publish to queue
 	publishToQueue bool
-	// days since the last vulnerability scan was run.
-	// 0 means only run once
-	daysSinceLastScan int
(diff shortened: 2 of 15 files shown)


jeffmendoza (approved):

mihaimaruseac (commented):
I think this makes sense, but leaving just a comment so that others get a chance to look at it too

pxp928:
Looks like I need to fix the e2e test here

pxp928:
should be fixed now

internal/testing/e2e/e2e
```diff
@@ -148,8 +148,6 @@ for command in "${ingestion_commands[@]}"; do
   echo @@@@ Waiting
   sleep 60
 
-  go run ${GUAC_DIR}"/cmd/guacone" certifier osv -p=false
```
pxp928, line 151: removed `guacone certifier osv` as `--add-vuln-on-ingest=true` is already being called. This caused duplicate vulns to show up and break the e2e test.

lumjjb (approved):

PR #2086, opened 2024-08-19
Atlas migration image

pxp928:
# Description of the PR

Create and publish an atlas migration docker image

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/ent/migrate/Dockerfile
+++ b/pkg/assembler/backends/ent/migrate/Dockerfile
@@ -1,21 +1,9 @@
-FROM alpine:3.18 AS migrations
+FROM arigaio/atlas:latest-alpine
 
 WORKDIR /app
 
 # Copy the migration files
 COPY migrations ./migrations
-
-# Install Atlas CLI
-RUN apk add --no-cache curl && \
-    curl -sSf https://atlasgo.sh | sh
-
-FROM alpine:3.18
-
-WORKDIR /app
-
-COPY --from=migrations /app/migrations /app/migrations
-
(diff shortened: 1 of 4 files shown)


mihaimaruseac (approved):

pkg/assembler/backends/ent/migrate/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM alpine:3.18 AS migrations
+
+WORKDIR /app
+
+# Copy the migration files
+COPY migrations ./migrations
+
+# Install Atlas CLI
+RUN apk add --no-cache curl && \
+    curl -sSf https://atlasgo.sh | sh
```
jeffmendoza, line 10: Can we use the upstream image: https://atlasgo.io/guides/deploying/image#defining-the-dockerfile?
If not, can we use `go install` instead of `curl | sh`?

pxp928: So I originally had this but Atlas takes in a URL and I wanted to this work with postgres via the standard pg env variables (PGHOST PGPORT PGDATABASE PGUSER PGPASSWORD). That was the reason for creating the script. The base image `arigaio/atlas:latest`  does not contain `bin/sh` which causes the build to fail. I also was a bit iffy on the curl install but the `go install` is the better approach.

On further check there is an alpine version of ` arigaio/atlas:latest-alpine` that I can use. Updated!

jeffmendoza (approved):

pkg/assembler/backends/ent/migrate/Dockerfile
jeffmendoza: Cool, yes the entrypoint looks nice.

PR #2153, opened 2024-09-27
Search for Vulns via Artifact

nathannaveen:
# Description of the PR

With the hasSBOM nodes being attached to artifacts as well as packages, we have to also search hasSBOM nodes via artifacts as well as packages to find vuln nodes.

This PR includes the functionality to search for vuln nodes via artifacts in addition to our already implemented functionality to search via packages.

## Feature change:

* With this PR, the user has to include the input type (purl, uri, artifact) when searching for vulns. This is because the user can now pass in an artifact, a purl, or a uri into the query. But, trying to understand the input is very complicated process, so leaving it as the user input is probably the best solution for now instead of us trying to guess the inputs type.
* I will create a PR for guac-docs regarding this so that the demos use commands that have the types passed in.

# PR Checklist

### DIFF
--- a/cmd/guacone/cmd/vulnerability.go
+++ b/cmd/guacone/cmd/vulnerability.go
@@ -127,34 +127,30 @@ func printVulnInfo(ctx context.Context, gqlclient graphql.Client, t table.Writer
 	var path []string
 	var tableRows []table.Row
 
-	// The primaryCall parameter in searchForSBOMViaPkg is there for us to know whether
-	// the searchString is expected to be a PURL, and we are searching via a purl.
-	depVulnPath, depVulnTableRows, err := guacanalytics.SearchForSBOMViaPkg(ctx, gqlclient, opts.searchString, opts.depth, opts.isPurl)
-	if err != nil {
-		logger.Fatalf("error searching via hasSBOM: %v", err)
+	var depVulnPath []string
+	var depVulnTableRows []table.Row
+	//check if artifact
+	split := strings.Split(opts.searchString, ":")
+	if len(split) == 2 {
+		var err error
+		// The primaryCall parameter in searchForSBOMViaPkg is there for us to know whether
+		// the searchString is expected to be a PURL, and we are searching via a purl.
+		depVulnPath, depVulnTableRows, err = guacanalytics.SearchForSBOMViaArtifact(ctx, gqlclient, opts.searchString, opts.depth)
+		if err != nil {
+			logger.Fatalf("error searching via hasSBOM: %v", err)
+		}
+	} else {
+		var err error
+		// The primaryCall parameter in searchForSBOMViaPkg is there for us to know whether
+		// the searchString is expected to be a PURL, and we are searching via a purl.
+		depVulnPath, depVulnTableRows, err = guacanalytics.SearchForSBOMViaPkg(ctx, gqlclient, opts.searchString, opts.depth, opts.isPurl)
+		if err != nil {
+			logger.Fatalf("error searching via hasSBOM: %v", err)
+		}
 	}
 
 	path = append(path, depVulnPath...)
 	tableRows = append(tableRows, depVulnTableRows...)
-
-	if len(depVulnPath) == 0 {
-		occur := searchArtToPkg(ctx, gqlclient, opts.searchString, logger)
-
-		if occur != nil && len(occur.IsOccurrence) > 0 {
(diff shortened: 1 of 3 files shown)


cmd/guacone/cmd/vulnerability.go
```diff
@@ -50,14 +51,28 @@ type queryOptions struct {
 	vulnerabilityID string
 	depth           int
 	pathsToReturn   int
+	inputType       string
 }
 
 var queryVulnCmd = &cobra.Command{
-	Use:   "vuln [flags] <purl/sbomURI>",
+	Use:   "vuln <type> <input>",
```
funnelfiasco, line 58: If I'm reading it right, this is a compatability-breaking change in the CLI. We should make sure any docs and demos that use the `vuln` command are updated accordingly.

Alternatively, is there a way we can figure out from the `<input>` what the `<type>` is so that the user doesn't have to specify it? That would, I think, preserve compatibility.

cmd/guacone/cmd/vulnerability.go
```diff
@@ -50,14 +51,28 @@ type queryOptions struct {
...
 	Short: "query if a package is affected by the specified vulnerability",
+	Long: `The vuln command allows you to query whether a specific package, SBOM URI, or artifact is affected by a given vulnerability.
+
+Positional Arguments:
+  <type>    Specify the input type: 'artifact', 'uri', or 'purl'
+  <input>   The corresponding input based on the specified type`,
 	Run: func(cmd *cobra.Command, args []string) {
 		ctx := logging.WithLogger(context.Background())
 
+		// Ensure exactly two arguments are provided: <type> and <input>
```
funnelfiasco, line 68: Are we dropping support for the optional flags in the current version or do they get handled separately from this check?

cmd/guacone/cmd/vulnerability.go
nathannaveen: @funnelfiasco, thank you for the reminder, I forgot to change the docs. So, I have created a PR to the docs https://github.com/guacsec/guac-docs/pull/150.

Additionally, I have no idea how to figure out the type from the input because of these couple reasons. If you know out a way to get around these issues, then please let me know, and I will update the code!

* I was thinking of first checking whether the input string has a colon in it (":") because artifacts are created in the format of `<"algorithm">:<"digest">`. The issue with this is that purls also have colons in them.
* Then I thought about checking if the input string has a `pkg:` at the beginning. But, SBOM URIs can have that in them as well.

cmd/guacone/cmd/vulnerability.go
nathannaveen: We aren't dropping the support for optional flags, they are still being handled. The only part of the code this PR has changed, is the part for the command line arguments.

https://github.com/guacsec/guac/blob/f292dbcc182b00f6e891c4b6513d081a8bf86a02/cmd/guacone/cmd/vulnerability.go#L580-L587

cmd/guacone/cmd/vulnerability.go
funnelfiasco: Thanks for the docs PR! I'll review shortly. As for detecting the type, I figured that would probably be the answer, but it was worth asking 😄

pkg/guacanalytics/searchForSBOM.go
```diff
@@ -30,6 +32,198 @@ func getVulnAndVexNeighborsForPackage(ctx context.Context, gqlclient graphql.Cli
...
+	checkedCertifyVulnIDs := make(map[string]bool)
+
+	for _, result := range collectedPkgVersionResults {
+		for _, neighbor := range result.pkgVersionNeighborResponse.Neighbors {
+			switch n := neighbor.(type) {
+			case *model.NeighborsNeighborsCertifyVuln:
+				if !checkedCertifyVulnIDs[n.Id] && n.Vulnerability.Type != noVulnType {
+					checkedCertifyVulnIDs[n.Id] = true
+					for _, vuln := range n.Vulnerability.VulnerabilityIDs {
+						if !AlreadyIncludedTableRows[vuln.VulnerabilityID] {
```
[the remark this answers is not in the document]
nathannaveen: I have fixed the duplicates, it turns out that I fixed this issue for when we search via artifact, but not when we search via package.

pxp928 (approved):
LGTM, thanks!

mihaimaruseac (approved):

PR #2204, opened 2024-10-18
update hasSBOM to filter deps.dev by default, also return only values specified

pxp928:
# Description of the PR

This by default changes the `hasSBOM` queries to filter our deps.dev collector SBOMs. It also adds the ability to filter out the returned values to make the query more efficient.

# PR Checklist

- [x] If GraphQL schema is changed, GraphQL client updates/additions have been made
- [x] If ent schema is changed, `make generate` has been run

### DIFF
--- a/pkg/assembler/clients/operations/hasSBOM.graphql
+++ b/pkg/assembler/clients/operations/hasSBOM.graphql
@@ -50,7 +50,7 @@ query HasSBOMs($filter: HasSBOMSpec!) {
     ...AllHasSBOMTree
   }
 }
-query HasSBOMList($filter: HasSBOMSpec!, $after: ID, $first: Int, $getIncludedSoftware: Boolean!, $getIncludedDependencies: Boolean!, $getIncludedOccurrences: Boolean!) {
+query HasSBOMList($filter: HasSBOMSpec!, $after: ID, $first: Int, $getIncludedSoftware: Boolean, $getIncludedDependencies: Boolean, $getIncludedOccurrences: Boolean) {
   HasSBOMList(hasSBOMSpec: $filter, after: $after, first: $first, getIncludedSoftware: $getIncludedSoftware, getIncludedDependencies: $getIncludedDependencies, getIncludedOccurrences: $getIncludedOccurrences) {
     totalCount
     edges {
--- a/pkg/assembler/clients/operations/hasSBOM.graphql
+++ b/pkg/assembler/clients/operations/hasSBOM.graphql
@@ -50,9 +50,8 @@ query HasSBOMs($filter: HasSBOMSpec!) {
     ...AllHasSBOMTree
   }
 }
-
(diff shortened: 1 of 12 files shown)


pkg/assembler/clients/operations/hasSBOM.graphql
```diff
@@ -50,9 +50,8 @@ query HasSBOMs($filter: HasSBOMSpec!) {
     ...AllHasSBOMTree
   }
 }
-
-query HasSBOMList($filter: HasSBOMSpec!, $after: ID, $first: Int) {
-  HasSBOMList(hasSBOMSpec: $filter, after: $after, first: $first) {
+query HasSBOMList($filter: HasSBOMSpec!, $after: ID, $first: Int, $getIncludedSoftware: Boolean, $getIncludedDependencies: Boolean, $getIncludedOccurrences: Boolean) {
```
pxp928, line 53: we can do the filtering based on what the user asks based on the graphQL query but this does not work when we are using it on the client side and we do not have this ability. This is the reason to add the boolean values to modify the output.

pxp928: What are your thoughts on this @mihaimaruseac? Is there a better way to do this?

mihaimaruseac: It seems this is returning different nodes, right? For included software it's only `Package` or `Artifact`, whereas for dependencies is `IsDependency` and for occurences is `IsOccurence`. If so, then users can request what they need from the query

pxp928: how?

mihaimaruseac: From the schema: https://github.com/guacsec/guac/blob/1a04f1399485a1ada505496ec9445d3b1ef4bc47/pkg/assembler/graphql/schema/hasSBOM.graphql#L67-L69

So a user that only wants the occurences would have something like

```gql
HasSBOMList(...) {
  edges {
    node {
      includedOccurrences   # THIS ONE
         ...
    }
  }
}
```

instead of including everything.

Then we would probably add these booleans on the client side to send the query with only what is relevant

mihaimaruseac: (my gql example might not be fully correct now, cannot test myself right now)

pxp928: But how does that work from the client side query? We don't have the ability to specify what the user wants, unless we create more client side queries that filter out the values...

pxp928: > HasSBOMList(...) {
  edges {
    node {
      includedOccurrences   # THIS ONE
         ...
    }
  }
}
Yeah this is what I was thinking, I will change the PR to that.

PR #2205, opened 2024-10-18
change hasSBOMList to add filter based on client usage

pxp928:
# Description of the PR

change hasSBOM filter based on client usage

# PR Checklist

### DIFF
--- a/pkg/assembler/backends/helper/preloads.go
+++ b/pkg/assembler/backends/helper/preloads.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2023 The GUAC Authors.
+// Copyright 2024 The GUAC Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/assembler/backends/helper/preloads.go
+++ b/pkg/assembler/backends/helper/preloads.go
@@ -1,0 +1,68 @@
+//
+// Copyright 2023 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package helper
+
+import (
+	"context"
+
+	"github.com/99designs/gqlgen/graphql"
+)
+
+// getPreloads get the specific graphQL query fields that are requested.
+// graphql.CollectAllFields only provides the top level fields and none of the nested fields below it.
+// getPreloads recursively goes through the fields and retrieves each nested field below it.
+// for example:
+/*
+  type
+  namespaces {
+    namespace
+    names {
+      name
+      tag
+      commit
+    }
+  }
+  will return:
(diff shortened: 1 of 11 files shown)


pkg/assembler/backends/helper/preloads.go
```diff
@@ -0,0 +1,68 @@
...
+    names {
+      name
+      tag
+      commit
+    }
+  }
+  will return:
+  fields: [type namespaces namespaces.namespace namespaces.names namespaces.names.name namespaces.names.tag namespaces.names.commit]
+*/
+func GetPreloads(ctx context.Context) []string {
```
pxp928, line 41: how is this @mihaimaruseac ?

pxp928: ugh this causes the tests to fail:

https://github.com/99designs/gqlgen/blob/f9eac5c040f084424d13ea2edc068c0cda46da9c/graphql/context_operation.go#L61-L66


any way to get around this?

mihaimaruseac: This is it!

Hmm, for the test failure, I think the link is pointing to something else?

mihaimaruseac (approved):

pxp928:
yeah they have a built in panic that is causing the tests to fail.....

mihaimaruseac:
It's this one: https://github.com/guacsec/guac/blob/95ebb0672404bbad33a14318c2f0d07a098c46ef/internal/testing/backend/hasSBOM_test.go#L325-L340

Let me look into it, it comes from a different backend by the look of it

pxp928:
its related to the preloads not being there for the tests, which cause failure...

mihaimaruseac:
Maybe https://github.com/99designs/gqlgen/discussions/1793 ?

pkg/assembler/backends/helper/preloads.go
pxp928: cool this is fixed! Thanks @mihaimaruseac!

mihaimaruseac (approved):

pkg/assembler/graphql/schema/hasSBOM.graphql
```diff
@@ -116,9 +116,9 @@ type HasSBOMEdge {
 }
 
 extend type Query {
-  "Returns all SBOM certifications."
+  "Returns all SBOM certifications (by default it filters out all deps.dev hasSBOMs)."
```
lumjjb, line 119: How do i run a query if i want to include deps.dev results? i wonder if this is something we want to bake in on the server side or just have it be a default client query? Like HasSbomOpiniated query?

I guess we would have to add the schema change where HasSBOMSpec has an exclude collectors?

lumjjb (commented):
The part on filtering out the non-requested graphql fields LGTM. I have some questions on the semantics for baking deps.dev filtering into the hasSbom spec.. This seems fairly specific and not part of the data model... I'm less opinionated if we decide to end up deprecating the gql interface, otherwise it seems like a backend implementation specific detail in the server logic.

pkg/assembler/graphql/schema/hasSBOM.graphql
pxp928: so if you specify the hasSBOM spec collect = deps.dev, it will include deps.dev SBOM otherwise they will be ignored

pxp928: As this point, where we removed deps.dev collector long term, they would never be used in the query.

pxp928: > I guess we would have to add the schema change where HasSBOMSpec has an exclude collectors?

I don't think we want to bake in `not equal` into graphql..

pxp928: @lumjjb I removed the HasSbom opinionated filtering on deps.dev! PTAL

lumjjb (approved):

PR #2219, opened 2024-10-27
Collect additional metadata for vulnerabilities from OSV

hown3d:
# Description of the PR

This PR introduces a new flag, `add-vuln-metadata`, that enables the collection of severity information for CVEs directly from the OSV API. When this flag is enabled, the collected severity data is ingested into GUAC as a `VulnerabilityMetadata` nodes, enriching vulnerability data with additional context on severity levels.

## Key Changes:

- New Flag: Adds the add-vuln-metadata CLI flag.
- Ingestion to GUAC: Captures the retrieved severity data and stores it as a `VulnerabilityMetadata` node, allowing for improved vulnerability assessment and analysis within GUAC.

This enhancement provides better visibility into CVE severity, enabling users to prioritize vulnerabilities more effectively based on standardized severity metrics.

# PR Checklist

Signed-off-by: Lukas Hoehl <lukas.hoehl@stackit.de>


pxp928:
Thanks @hown3d this is a great addition! Will review this soon!

pxp928:
> For now I've only included the [severity field](https://ossf.github.io/osv-schema/#severity-field) from the OSV API.
> 
> I see potential in ingesting additional metadata like [summary](https://ossf.github.io/osv-schema/#summary-details-fields), [details](https://ossf.github.io/osv-schema/#summary-details-fields) and [references](https://ossf.github.io/osv-schema/#summary-details-fields) into GUAC. Those are not part of the vulnerability attestation (yet?) though, so it's not clear how to parse these from the attestation into `VulnerabilityMetadata` nodes.


For the vulnerability summary, details..etc, our thought process was that information did not need to be stored in the graph. Rather, it can be retrieved client side when needed. The CVE - severity field are needed to make decisions.

pxp928:
Sorry for the delay! I will review this soon!

cmd/guacingest/cmd/ingest.go
```diff
@@ -99,7 +101,7 @@ func ingest(cmd *cobra.Command, args []string) {
 	defer csubClient.Close()
 
 	emit := func(d *processor.Document) error {
-		if _, err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, transport, csubClient, opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion); err != nil {
+		if _, err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, transport, csubClient, opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion, opts.addVulnMetadata); err != nil {
```
pxp928, line 104: we should remove this from the on ingestion. I think we default to false on ingestion but have it as an option for the certifiers.

cmd/guacone/cmd/deps_dev.go
```diff
@@ -87,7 +88,7 @@ var depsDevCmd = &cobra.Command{
 		emit := func(d *processor.Document) error {
 			totalNum += 1
 
-			if _, err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, transport, csc, opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion); err != nil {
+			if _, err := ingestor.Ingest(ctx, d, opts.graphqlEndpoint, transport, csc, opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion, opts.addVulnMetadata); err != nil {
```
pxp928, line 91: same for all these, we should remove it on ingestion.

cmd/guacone/cmd/osv.go
```diff
@@ -125,7 +134,7 @@ var osvCmd = &cobra.Command{
 				case <-ticker.C:
 					if len(totalDocs) > 0 {
 						err = ingestor.MergedIngest(ctx, totalDocs, opts.graphqlEndpoint, transport, csubClient,
-							opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion)
+							opts.queryVulnOnIngestion, opts.queryLicenseOnIngestion, opts.addVulnMetadata)
```
pxp928, line 137: can be removed from here.

cmd/guacone/cmd/osv.go
```diff
@@ -90,7 +93,13 @@ var osvCmd = &cobra.Command{
 		logger := logging.FromContext(ctx)
 		transport := cli.HTTPHeaderTransport(ctx, opts.headerFile, http.DefaultTransport)
 
-		if err := certify.RegisterCertifier(osv.NewOSVCertificationParser, certifier.CertifierOSV); err != nil {
+		if err := certify.RegisterCertifier(func() certifier.Certifier {
+			certifierOpts := []osv.CertifierOpts{}
+			if opts.addVulnMetadata {
+				certifierOpts = append(certifierOpts, osv.WithVulnerabilityMetadata())
```
pxp928, line 99: keep this as we should do it when running the certifier

cmd/guaccollect/cmd/osv.go
```diff
@@ -108,7 +111,13 @@ you have access to read and write to the respective blob store.`,
 		ctx := logging.WithLogger(context.Background())
 		logger := logging.FromContext(ctx)
 
-		if err := certify.RegisterCertifier(osv.NewOSVCertificationParser, certifier.CertifierOSV); err != nil {
+		if err := certify.RegisterCertifier(func() certifier.Certifier {
+			certifierOpts := []osv.CertifierOpts{}
+			if opts.addVulnMetadata {
+				certifierOpts = append(certifierOpts, osv.WithVulnerabilityMetadata())
```
pxp928, line 117: keep this as we should do it when running the certifier

go.mod
```diff
@@ -291,7 +290,7 @@ require (
 	deps.dev/api/v3 v3.0.0-20240923041156-0312db85d6d6
 	entgo.io/contrib v0.6.0
 	entgo.io/ent v0.14.0
-	github.com/99designs/gqlgen v0.17.54
+	github.com/99designs/gqlgen v0.17.55
```
pxp928, line 293: This should be up-to date on the main branch so we can remove this and this will remove all the generated code in this PR.

pkg/cli/store.go
```diff
@@ -67,6 +67,7 @@ func init() {
 
 	// the ingestor will query and ingest OSV for vulnerabilities
 	set.Bool("add-vuln-on-ingest", false, "if enabled, the ingestor will query and ingest OSV for vulnerabilities. Warning: This will increase ingestion times")
+	set.Bool("add-vuln-metadata", false, "if enabled, the ingestor will add metadata to vulnerabilities from OSV")
```
pxp928, line 70: we can remove this.

pkg/ingestor/ingestor.go
```diff
@@ -44,11 +44,12 @@ func Ingest(
 	csubClient csub_client.Client,
 	scanForVulns bool,
 	scanForLicense bool,
+	addVulnMetadata bool,
```
pxp928, line 47: same here, this can be removed.

pkg/ingestor/parser/parser.go
```diff
@@ -75,7 +73,7 @@ func RegisterDocumentParser(p func() common.DocumentParser, d processor.Document
 }
 
 // ParseDocumentTree takes the DocumentTree and create graph inputs (nodes and edges) per document node.
-func ParseDocumentTree(ctx context.Context, docTree processor.DocumentTree, scanForVulns bool, scanForLicense bool) ([]assembler.IngestPredicates, []*common.IdentifierStrings, error) {
+func ParseDocumentTree(ctx context.Context, docTree processor.DocumentTree, scanForVulns bool, scanForLicense bool, addVulnMetadata bool) ([]assembler.IngestPredicates, []*common.IdentifierStrings, error) {
```
pxp928, line 76: `addVulnMetadata` can be removed here

pxp928 (commented):
Overall this is a great PR, very comprehensive! We should just remove it from "on ingestion" otherwise this is good to go! Thank You!

hown3d:
> Overall this is a great PR, very comprehensive! We should just remove it from "on ingestion" otherwise this is good to go! Thank You!

Rechecking the code i don't think it's neccessary to use a flag instead just defaulting to add vulnerability metadata. It introduces so many places to add this flag that it gets quite confusing. 
I would argue that it's not a problem to always store the included severity in GUAC.

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -145,13 +154,33 @@ func parseVulns(_ context.Context, s *attestation_vuln.VulnerabilityStatement) (
...
+			score, err := parseScoreBasedOnMethod(severity)
+			if err != nil {
+				severityErrors = errors.Join(fmt.Errorf("parsing severity score failed for method %s: %w", severity.Method, err))
+			}
+			vmi = append(vmi, assembler.VulnMetadataIngest{
+				Vulnerability: vuln,
+				VulnMetadata: &generated.VulnerabilityMetadataInputSpec{
+					ScoreType:  generated.VulnerabilityScoreType(severity.Method),
+					ScoreValue: score,
+					Timestamp:  now,
```
lumjjb, line 169: It seems like OSV has a last modified timestamp, would it make sense to use that?

e.g. in https://google.github.io/osv.dev/get-v1-vulns/
```
  "modified": "2022-04-13T03:04:37.331327Z",
  "published": "2020-06-24T01:51:14.570467Z",
```

lumjjb (commented):
Awesome - this is a great addition to the OSV certifier, thanks @hown3d ! I have one request, else LGTM!

pkg/ingestor/parser/vuln/vuln.go
hown3d: Seems reasonable.
The question is where to embed this information in the vulnerability attestation since this is the base for the `VulnMetadataIngest` struct.

lumjjb: I think this timestamp field works fine, it seems to match up well (to a very close approximation that the score is derived when a vuln entry is updated) with the graphQL spec description (https://github.com/guacsec/guac/blob/main/pkg/assembler/graphql/schema/vulnMetadata.graphql#L62)

lumjjb: we chatted during the community all and since this is only for the vulns attestation, so we don't have access to OSV scan info here, so now is fine for now.

lumjjb (approved):

pkg/ingestor/parser/vuln/vuln.go
```diff
@@ -124,8 +130,11 @@ func parseMetadata(s *attestation_vuln.VulnerabilityStatement) *generated.ScanMe
 
 // TODO (pxp928): Remove creation of osv node and just create the vulnerability nodes specified
 func parseVulns(_ context.Context, s *attestation_vuln.VulnerabilityStatement) ([]*generated.VulnerabilityInputSpec,
-	[]assembler.VulnEqualIngest, error) {
+	[]assembler.VulnEqualIngest, []assembler.VulnMetadataIngest, error,
+) {
+	now := time.Now()
```
pxp928, line 135: Based on the discussion on the meeting, I do see that if we set the timestamp to be now, it will always get re-created in the graph DB as it does the field does not match. It would be better to capture a timestamp of the score from the scanner and use that so that we do not keep re-creating the `VulnMetadata` node multiple time. This is similar to the comment @lumjjb made. If we can keep the timestamp static per the OSV database, unless the score changes, the node will not get re-created.

pxp928 (commented):
LGTM! I just have a concern on the timestamp that I may be misunderstanding?

pkg/ingestor/parser/vuln/vuln.go
lumjjb: Actually we had another discussion during the maintainers call on 25 nov, and we were concerned around there being too many nodes generated per timestamp, we landed on having the timestamp be null'ed out, since that information on the scan would be available at the CertifyVuln level.

lumjjb (changes requested):
see follow up too comment. minor change

pkg/ingestor/parser/vuln/vuln.go
pxp928: Once the change above is done, this is ready to merge! Thanks @hown3d!

hown3d: I've dropped the `time.Now()` here

PR #2243, opened 2024-10-30
Export getGraphqlServer and accept initialized backend

### LINKED ISSUE #2240: [feature] Support for Out-of-Tree GUAC Backends
This proposal aims to allow GUAC backends to be developed and maintained out-of-tree, similar to how Kubernetes transitioned its cloud providers. This approach could be the first step towards eventually migrating some or all backends out-of-tree, however this proposal is specifically targeting the ability to register new out-of-tree backends.

A couple of thoughts:
- By supporting out-of-tree backends, they can be updated and released independently of the core GUAC releases. This allows for faster iteration on backend-specific features and bug fixes without impacting the core. A good historical analogy might be the migration of the cloud providers for Kubernetes out-of-tree.
- Decoupling reduces the complexity of the GUAC core codebase, making it easier to maintain and scale. It also isolates issues to specific backends without affecting the main application.
- Allowing backends to be out-of-tree lowers the barrier for external contributors to develop and maintain new backends, fostering a more vibrant ecosystem around GUAC.

I think there are two possible approaches to achieve this:
1. Expose Backend Methods for External Use by making backend registration and initialization methods in `cmd/guacgql` publicly accessible. This would allow external backends to integrate by importing those methods.
2. Modify GUAC to load backends at runtime via a plugin interface or service discovery.

I'd be more than happy to drive the work and put up an initial PR. I think # 1 should be the shortest path with the least change to codebase. Any feedback or suggestions would be greatly appreciated!

- I guess the alternative would be for out-of-tree backend to maintain a fork of the guac codebase.

**Additional context**
N/A

robert-cronin:
# Description of the PR

Fixes #2240 

Some notes:
- I intended to restrict this PR to refactoring only instead of altering functionality, so if there is any change to functionality it was not intentional, and we should probably flag and revert.
- The main change here is: a refactor of `getGraphqlServer` to allow it to be used programmatically by out-of-tree backends, allows the backend to be passed in as-is without relying on parsing of the flags for config.

cc @ridhoq 

# PR Checklist

### DIFF
--- a/cmd/guacgql/cmd/server.go
+++ b/cmd/guacgql/cmd/server.go
@@ -19,7 +19,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/guacsec/guac/pkg/version"
 	"net/http"
 	"os"
 	"os/signal"
@@ -27,7 +26,8 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/99designs/gqlgen/graphql/handler"
+	"github.com/guacsec/guac/pkg/version"
+
 	"github.com/99designs/gqlgen/graphql/handler/debug"
 	"github.com/99designs/gqlgen/graphql/playground"
 	"github.com/guacsec/guac/pkg/assembler/backends"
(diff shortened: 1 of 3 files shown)


cmd/guacgql/cmd/server.go
```diff
@@ -172,20 +174,12 @@ func validateFlags() error {
...
-func getGraphqlServer(ctx context.Context) (*handler.Server, error) {
-	var topResolver resolvers.Resolver
-
-	backend, err := backends.Get(flags.backend, ctx, getOpts[flags.backend](ctx))
-	if err != nil {
-		return nil, fmt.Errorf("Error creating %v backend: %w", flags.backend, err)
-	}
-	topResolver = resolvers.Resolver{Backend: backend}
-
+func GetGraphqlServer(ctx context.Context, backend backends.Backend) *handler.Server {
```
lumjjb, line 177: If the intention is to have this be used outside the cmd itself, can we move it to a `pkg/` path? Perhaps `pkg/assembler/server`?

lumjjb (approved):
LGTM, but request a slight refactor to expose this through a package instead of from the cmd package

cmd/guacgql/cmd/server.go
robert-cronin: great idea! I'll refactor it into `pkg/assembler/server`

robert-cronin: great idea! I'll refactor it into `pkg/assembler/server`

robert-cronin: done!

mihaimaruseac (approved):

mihaimaruseac (approved):

PR #2326, opened 2024-11-29
Reduce scorecard workflow permissions scope

### LINKED ISSUE #2275: zizmor audit for scorecard workflow
```
🌈 completed scorecard.yml
warning[excessive-permissions]: overly broad workflow or job-level permissions
  --> .github/workflows/scorecard.yml:18:1
   |
18 | permissions: read-all
   | --------------------- uses read-all permissions
   |
   = note: audit confidence → High

1 findings (0 ignored): 0 unknown, 0 informational, 0 low, 1 medium, 0 high
```

robert-cronin:
# Description of the PR

Fixes #2275 

# PR Checklist

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -15,7 +15,8 @@ on:
     branches: [ "main" ]
 
 # Declare default permissions as read only.
-permissions: read-all
+permissions:
+  contents: read
 
 jobs:
   analysis:


robert-cronin:
I did try with the id-token: write permission set but it seems that the Scorecard API has some checks to ensure the integrity of the data, including default branch check: 

https://github.com/robert-cronin/guac/actions/runs/12076752606/job/33678574053
https://github.com/ossf/scorecard-action?tab=readme-ov-file#workflow-restrictions

funnelfiasco (commented):
We shouldn't declare write permissions at the workflow level as a matter of practice. Write permissions should be granted on a per-job basis as needed. The practical effect is the same here, since there's only one job, but the tooling will still complain :-)

robert-cronin:
> We shouldn't declare write permissions at the workflow level as a matter of practice. Write permissions should be granted on a per-job basis as needed. The practical effect is the same here, since there's only one job, but the tooling will still complain :-)

Okay I agree, I'll take the write permissions out. Also just to note here, I believe the `id-token: write` permissions is needed to make `publish_results: true` work properly, but I'll leave that as-is for now and this PR can just target a reduction in permission scope. Thanks Ben!

robert-cronin:
> We shouldn't declare write permissions at the workflow level as a matter of practice. Write permissions should be granted on a per-job basis as needed. The practical effect is the same here, since there's only one job, but the tooling will still complain :-)

Oh, I missed the existing id-token permission set on the job level! Thank you for catching that! I'll take out the permission I added to the workflow level

funnelfiasco (approved):
LGTM!

mihaimaruseac (approved):

pxp928 (approved):

PR #2366, opened 2024-12-12
feat: adds datadog certifier

### LINKED ISSUE #2345: [feature] Add support for DataDog's malicious software package dataset

Malicious packages exist in public repos, and sometimes people bring them in.

Parse the [DataDog Malicious Packages Dataset](https://github.com/DataDog/malicious-software-packages-dataset) to pull in known-malicious packages and flag them when they appear in the graph.

Currently, the dataset provides JSON manifests for known malicious [PyPI](https://github.com/DataDog/malicious-software-packages-dataset/blob/main/samples/pypi/manifest.json) and [NPM](https://github.com/DataDog/malicious-software-packages-dataset/blob/main/samples/npm/manifest.json).

I'm thinking this would be a certifier that runs and creates a CertifyBad node for packages.

robert-cronin:
# Description of the PR

Fixes #2345

I am not sure if there is a need for a parser or attestation since were just ingesting CertifyBad for a particular pURL, but if there is a need to represent the source information in a predicate, I'd be happy to try and figure out how to add that in.

# PR Checklist

### DIFF
--- a/pkg/certifier/datadog_malware/datadog_malware.go
+++ b/pkg/certifier/datadog_malware/datadog_malware.go
@@ -1,0 +1,303 @@
+//
+// Copyright 2024 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package datadog_malware
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/guacsec/guac/pkg/assembler"
+	"github.com/guacsec/guac/pkg/assembler/clients/generated"
+	ingestor "github.com/guacsec/guac/pkg/assembler/clients/helpers"
+	"github.com/guacsec/guac/pkg/assembler/helpers"
+	"github.com/guacsec/guac/pkg/certifier"
+	"github.com/guacsec/guac/pkg/certifier/components/root_package"
+	"github.com/guacsec/guac/pkg/clients"
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/guacsec/guac/pkg/logging"
+	"github.com/guacsec/guac/pkg/version"
+	jsoniter "github.com/json-iterator/go"
+	"golang.org/x/time/rate"
+)
+
+var (
+	json              = jsoniter.ConfigCompatibleWithStandardLibrary
+	rateLimit         = 10000
+	rateLimitInterval = 30 * time.Second
+)
+
+const (
+	NPM_MANIFEST_URL        string = "https://raw.githubusercontent.com/DataDog/malicious-software-packages-dataset/main/samples/npm/manifest.json"
+	PYPI_MANIFEST_URL       string = "https://raw.githubusercontent.com/DataDog/malicious-software-packages-dataset/main/samples/pypi/manifest.json"
+	DatadogMalwareCertifier string = "datadog_malware_certifier"
+
+	// manifestRefreshInterval is the interval at which we periodically re-fetch
+	// the Datadog malicious packages manifests.
+	manifestRefreshInterval = 4 * time.Hour
+)
+
+var ErrDatadogMalwareComponentTypeMismatch error = errors.New("rootComponent type is not []*root_package.PackageNode")
+
+type MaliciousPackages map[string][]string
+
+type assemblerFuncType func([]assembler.IngestPredicates) (*ingestor.AssemblerIngestedIDs, error)
+
+type datadogMalwareCertifier struct {
+	httpClient    *http.Client
+	assemblerFunc assemblerFuncType
+
+	dataMu   sync.RWMutex
+	npmData  MaliciousPackages
+	pypiData MaliciousPackages
+}
+
+// CertifierOption defines functional options for the certifier
+type CertifierOption func(*datadogMalwareCertifier)
+
+// WithHTTPClient allows overriding the default HTTP client
+func WithHTTPClient(client *http.Client) CertifierOption {
+	return func(d *datadogMalwareCertifier) {
+		d.httpClient = client
+	}
+}
+
+// NewDatadogMalwareCertifier initializes the Datadog Malicious Software Packages certifier.
+//
+// The Datadog malicious software packages dataset is a public repository that catalogs
+// known malicious artifacts for platforms such as npm and PyPI. By using this data,
+// we can certify suspect packages in our GUAC graph as known bad (i.e. CertifyBad).
+// https://github.com/DataDog/malicious-software-packages-dataset
+//
+//   - added predicates: each discovered malicious package is ingested with a `CertifyBad`
+//     predicate, describing the justification, origin, and collector used.
+//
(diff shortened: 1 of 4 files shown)


funnelfiasco:
As a general comment, I wonder if we want to call it something more specific than "DataDog"? "DataDog Malicious Packages DataSet" is unwieldy, but I'm concerned that there might be some future thing that pulls from DataDog proper and the name is already taken. I don't have any great ideas and this may not be a concern worth worrying about right now, but I wanted to raise it.

pkg/certifier/datadog/datadog.go
```diff
@@ -0,0 +1,219 @@
...
+		if len(maliciousVersions) == 0 {
+			// package known but no malicious versions listed?
+			continue
+		}
+
+		// certify only if the package has a specified version and that exact version is known malicious
+		if pkgInput.Version == nil {
+			logger.Debugf("Package %s has no version specified, skipping...", purl)
+			continue
+		}
```
funnelfiasco, line 177: Am I reading it correctly that we'll ignore things like "aiohtttps" (in PyPI) because it applies to all versions?

robert-cronin:
> As a general comment, I wonder if we want to call it something more specific than "DataDog"? "DataDog Malicious Packages DataSet" is unwieldy, but I'm concerned that there might be some future thing that pulls from DataDog proper and the name is already taken. I don't have any great ideas and this may not be a concern worth worrying about right now, but I wanted to raise it.

yeah, that is a solid point, if DataDog eventually spin out other datasets, I can see how that might cause some confusion. The data itself mostly comes from [GuardDog](https://github.com/datadog/guarddog) but I think not exclusively. Maybe we can go with something like `datadog-malware-dataset` or `datadog-mspd` but mspd is not a known acronym. The alternative is `datadog-malicious-software-packages-dataset`. 
The `datadog-malware-dataset` one sounds like the best compromise to me between clarity and brevity.

pkg/certifier/datadog/datadog.go
robert-cronin: I am not sure, I initially thought that an empty malicious version list meant no malicious versions found and thats what the code currently does, but it is instead the case that an empty list means all versions are known malicious? I couldn't find anything in the datadog malicious dataset repo to suggest this.

funnelfiasco: I believe the empty list means all versions are known malicious. For example "aiohtttps" is a typosquat of the "aiohttps" (notice the third 't'). I opened DataDog/malicious-software-packages-dataset#135 to request clarification, but I'm confident in my interpretation.

robert-cronin: after further inspection, given the number of misspelled versions of popular packages observed, i'm also now leaning towards your conclusion. if these packages are already in the malicious dataset, that strongly suggests their malicious intent at the package level, even if a specific version isn’t listed.

robert-cronin: i've now changed the code to reflect this interpretation.

pxp928:
Thanks @robert-cronin! Sorry for the delay. We will review this soon!

pkg/certifier/datadog/datadog.go
robert-cronin: This interpretation has been confirmed on the issue you posed, thanks @funnelfiasco !

robert-cronin:
> Thanks @robert-cronin! Sorry for the delay. We will review this soon!

No problems, thanks @pxp928!

pkg/certifier/datadog_malware/datadog_malware.go
```diff
@@ -0,0 +1,241 @@
...
+		httpClient:    defaultClient,
+		assemblerFunc: assemblerFunc,
+	}
+
+	// apply options
+	for _, opt := range opts {
+		opt(d)
+	}
+
+	if err := d.fetchManifests(); err != nil {
```
lumjjb, line 91: it looks like manifests are fetched once on initialization. Given the database will be updated regularly - it would be helpful to refresh the manifests based on some frequency. Is this feasible?

pkg/certifier/datadog_malware/datadog_malware.go
```diff
@@ -0,0 +1,241 @@
...
+
+		// determine which dataset to check based on package type
+		var maliciousVersions []string
+		switch pkgInput.Type {
+		case "npm":
+			fullName := pkgInput.Name
+			if pkgInput.Namespace != nil && *pkgInput.Namespace != "" {
+				namespace := strings.TrimPrefix(*pkgInput.Namespace, "@")
+				namespace = strings.TrimPrefix(namespace, "%40")
+				fullName = "@" + namespace + "/" + pkgInput.Name
```
lumjjb, line 149: it looks like it isn't always the case that the packages start with "@" in the dataset, could we add a check here after the trim to see if the namespace had the prefix? and add the "@" only if there was a prefix trim?

pkg/certifier/datadog_malware/datadog_malware.go
```diff
@@ -0,0 +1,241 @@
...
+
+// WithHTTPClient allows overriding the default HTTP client
+func WithHTTPClient(client *http.Client) CertifierOption {
+	return func(d *datadogMalwareCertifier) {
+		d.httpClient = client
+	}
+}
+
+// NewDatadogMalwareCertifier initializes the Datadog Malicious Software Packages certifier
+func NewDatadogMalwareCertifier(ctx context.Context, assemblerFunc assemblerFuncType, opts ...CertifierOption) (certifier.Certifier, error) {
```
lumjjb, line 76: Can you add a bit of documentation here on what the datadog malicious software packages are, for those that are not familiar.

In addition could you add some details on:
- The added predicates on the graph
- The recommended interval times (considering that the current certifier will generate a certifyBad each time indefinitely).
- Any caveats: see comment on periodic fetching of manifest.

lumjjb (commented):
This is a super cool addition. I wasn't aware of this dataset but this was a really cool implementation and its such a good example on how to add another data source easily (or at least you made it look easy! - any feedback on how to make this easier would be super great as well, or any particular frictions you had). Thanks so much for yet another great contribution! 🙌

robert-cronin:
> This is a super cool addition. I wasn't aware of this dataset but this was a really cool implementation and its such a good example on how to add another data source easily (or at least you made it look easy! - any feedback on how to make this easier would be super great as well, or any particular frictions you had). Thanks so much for yet another great contribution! 🙌

Thanks @lumjjb ! I really appreciate your encouraging words 😃 In terms of frictions, I think there are some options for improving the scalability of adding new data sources be they collectors or certifiers. Perhaps one idea is to define a common interface that any collector or certifier must implement and then have a registrar similar to how the backend works today in the spirit of dedpulication. There are also some common logic items in the certifiers/collectors like initialising nats/calling ingestion flow/ emitters etc. Not sure how much of that will be changing in v2.0 but it might be worth looking into

pkg/certifier/datadog_malware/datadog_malware.go
robert-cronin: Yes, I will add it in!

pkg/certifier/datadog_malware/datadog_malware.go
robert-cronin: Certainly, I've added something to this effect 👍

pkg/certifier/datadog_malware/datadog_malware.go
robert-cronin: Sure thing, I'll add some documentation along these lines

robert-cronin:
Hello @lumjjb! Your suggestions have been implemented and all outstanding changes addressed. If you have any other suggestions, let me know.

gaganhr94:
Looks like this PR was almost good to get merged, but got closed by the stale bot unfortunately. Reopening it again.

gaganhr94:
Hi @robert-cronin, are you still interested to work on this PR ? If yes, can you please fix the lint. Thanks

PR #2278, opened 2024-11-08
Add depsdev guac client as a stepping stone to split up depsdev functionality

lumjjb:
# Description of the PR
Add depsdev guac client as a stepping stone to split up depsdev functionality, based on #2264.

This PR doesn't introduce usage of this client library yet. This will come in a follow up PR

The client is based on a couple functions:

## ** main retrieval functions **

func (d *[DepsClient](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#DepsClient)) GetDependencies(ctx [context](http://localhost:6060/pkg/context/).[Context](http://localhost:6060/pkg/context/#Context), purls [][string](http://localhost:6060/pkg/builtin/#string)) ([]*[PackageComponent](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#PackageComponent), [error](http://localhost:6060/pkg/builtin/#error))
GetDependencies gets the information about dependencies for a PURL. This should be called after a call to RetrieveDependencies.

func (d *[DepsClient](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#DepsClient)) GetMetadata(ctx [context](http://localhost:6060/pkg/context/).[Context](http://localhost:6060/pkg/context/#Context), purls [][string](http://localhost:6060/pkg/builtin/#string)) ([]*[PackageComponent](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#PackageComponent), [error](http://localhost:6060/pkg/builtin/#error))
GetMetadata gets the information about metadata for a PURL. This should be called after a call to RetrieveVersionsAndProjects. This does not return dependency information about a purl

## **functions which help performance by having concurrent retrieval **

func (*DepsClient) [RetrieveDependencies](http://localhost:6060/src/github.com/guacsec/guac/internal/client/depsdevclient/deps_dev_client.go?s=6792:6876#L189) [¶](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#DepsClient.RetrieveDependencies)
func (d *[DepsClient](http://localhost:6060/pkg/github.com/guacsec/guac/internal/client/depsdevclient/#DepsClient)) RetrieveDependencies(ctx [context](http://localhost:6060/pkg/context/).[Context](http://localhost:6060/pkg/context/#Context), purls [][string](http://localhost:6060/pkg/builtin/#string)) [error](http://localhost:6060/pkg/builtin/#error)
[...]

PR #2385, opened 2024-12-18
add depsdev scanner implementation

lumjjb:
# Description of the PR

Implement deps.dev scanner on ingestion using depsdevclient introduced in #2278 

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/common/scanner/scanner.go
+++ b/pkg/ingestor/parser/common/scanner/scanner.go
@@ -19,8 +19,11 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"time"
 
+	deps_dev_client "github.com/guacsec/guac/internal/client/depsdevclient"
 	"github.com/guacsec/guac/pkg/assembler"
+	"github.com/guacsec/guac/pkg/assembler/clients/generated"
 	cd_certifier "github.com/guacsec/guac/pkg/certifier/clearlydefined"
 	eol_certifier "github.com/guacsec/guac/pkg/certifier/eol"
 	osv_certifier "github.com/guacsec/guac/pkg/certifier/osv"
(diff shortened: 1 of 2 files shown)


pkg/ingestor/parser/common/scanner/scanner.go
```diff
@@ -115,7 +118,64 @@ func purlsLicenseScanWithClient(
 }
 
 func PurlsDepsDevScan(ctx context.Context, purls []string) ([]assembler.CertifyScorecardIngest, []assembler.HasSourceAtIngest, error) {
```
pxp928, line 120: looks like this is set to false as its not part of the flags yet? By default it should be false but there is no flag to enable it.

pkg/ingestor/parser/common/scanner/scanner.go
```diff
@@ -115,7 +118,64 @@ func purlsLicenseScanWithClient(
...
-	return nil, nil, fmt.Errorf("Unimplemented")
+	certifyScorecards := []assembler.CertifyScorecardIngest{}
+	hasSourceAts := []assembler.HasSourceAtIngest{}
+
+	ddc, err := deps_dev_client.NewDepsClient(ctx)
```
pxp928, line 124: What is the added time to the ingestion when this is enabled? Can you post in a comment the time with this enabled and without?

pxp928 (approved):
Overall looks good but missing the CLI flag.

mihaimaruseac (approved):

pkg/ingestor/parser/common/scanner/scanner.go
lumjjb: I think this should be implemented in : https://github.com/guacsec/guac/pull/2265/files#diff-fc14c5bdd69817e3f9b1f331130f5055f448774eb64db13c0ef23a8e882cd1d4R74-R76? Does that address your comment?

pxp928: did we need to add it here: https://github.com/guacsec/guac/blob/main/pkg/cli/store.go

pkg/ingestor/parser/common/scanner/scanner.go
lumjjb: Added a comment here that each scan takes about 300ms - 500ms ranging from 1-30 PURLs.

PR #2440, opened 2025-01-15
Add otel instrumentation to http/grpc/sql libraries.

jeffmendoza:
Also add cli arg to the clis that use those libraries to enable exporting metrics and traces. Docs under pkg/metrics. Default to off and also change prometheus default to off.

# Description of the PR

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/deps_dev.go
+++ b/cmd/guaccollect/cmd/deps_dev.go
@@ -102,13 +102,16 @@ you have access to read and write to the respective blob store.`,
 			os.Exit(1)
 		}
 
-		var shutdown func(context.Context) error = func(context.Context) error { return nil }
 		if opts.enableOtel {
-			var err error
-			shutdown, err = metrics.SetupOTelSDK(ctx)
+			shutdown, err := metrics.SetupOTelSDK(ctx)
 			if err != nil {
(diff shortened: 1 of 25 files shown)


cmd/guaccollect/cmd/deps_dev.go
```diff
@@ -117,6 +131,7 @@ you have access to read and write to the respective blob store.`,
 		}
 
 		initializeNATsandCollector(ctx, opts.pubsubAddr, opts.blobAddr, opts.publishToQueue)
+		shutdown(ctx)
```
mihaimaruseac, line 134: Should this be via a `defer` as soon as we create it? Right now (at least for this function) it doesn't seem to make a difference but to future proof when we add some code path that returns early

jeffmendoza: Yeah, in a number of cases (this one) we are not using special timing on shutdown and a defer would be better.

Other cases we should call shutdown when we get the `C-c` signal but before we call the context cancel function. I'll look through and see which should be change and which are more sensitive.

I think the main thing for the shutdown is to flush any metrics that are recorded but not sent (the send happens in batches). This will be more applicable in some of the `guacone` commands that use the instrumented clients but do no poll.

mihaimaruseac (approved):

Issue #2387, opened 2024-12-19
[feature] Unifying SBOM Representation and Trust Across Multiple Equivalent Image References


robert-cronin:
In container ecosystems, a single image (identified by a digest) can appear in multiple registries under different tags or repositories. For instance, an image at `registryA.com/base:1.0` can be mirrored at `registryB.com/cache/base:1.0`. Although these references differ, the underlying image content is identical. A single, signed SBOM should logically apply to all such references without needing to be regenerated (which would break its original signature and trust).

Currently, GUAC typically associates one SBOM per subject (e.g., one image reference), making it hard to unify multiple references of the same image or multiple SBOM formats (e.g., SPDX and CycloneDX) for the same underlying content. Additionally, SBOMs aren’t first-class entities with their own trust or provenance data, limiting the ability to differentiate a trusted SBOM from a suspicious or forged one.

**Use Case Example:**

**1. Multiple Registry Locations, One Image:**

- Consider an official base image at `registryA.com/base:1.0` described by a trusted SBOM.
- Another registry provides a pull-through cache of this same digest at `registryB.com/cache/base:1.0`.
- In GUAC today, there’s no simple way to say: These two references represent the same underlying image, so we should link them to the same trusted SBOM.

**2. Different SBOM Formats, Same Content:**

- If the same image from above also has a CycloneDX SBOM and an SPDX SBOM, both describing identical content, GUAC cannot directly represent these two documents as equivalent descriptions.
- Users can’t query: “Show me all SBOM representations of this image” and see them unified.

**3. Trust and Attestation at the SBOM Level:**

- Suppose a malicious actor introduces a forged SBOM claiming to describe the same package.
- Without modeling SBOMs as distinct nodes, GUAC can’t easily attach attestations, cryptographic proofs, or trust metadata to that SBOM and differentiate it from the known-trusted one.

This issue aims to explore ways to unify image references, preserve trust in original SBOMs, and enhance the expressiveness of SBOM modeling in GUAC.

robert-cronin:
**Potential Directions (not prescriptive, just brainstorming):**

**Enhanced HasSBOM Edges:**

- Ingest the canonical SBOM once for the original image.
- For each equivalent image reference (e.g. same content but different registry/tag), create only a new HasSBOM edge linking that new image node to the existing SBOM nodes.
[...]

PR #2467, opened 2025-01-28
Feat/multiple equivalent images SBOM poc

robert-cronin:
# Description of the PR

Partial fix for #2387 

The main idea in this PR is to detect if an image being ingested has a SBOM that describes the same digest but has a different image uri and then create an IsOccurrence edge between them in a hub and spoke pattern (hub being the Artifact node and spokes being the Package nodes)

# PR Checklist


robert-cronin:
Here is a rough diagram showing the logic I am trying to model with this approach:
Untitled

robert-cronin:
The main thing here is the reference parser that converts a [reference attestation](https://github.com/in-toto/attestation/blob/main/spec/predicates/reference.md) to an `IsOccurrence` node that links the Occurrence (image pURL/Pkg) to the Artifact. I've removed the hacky part in the oci parser so that we may rely on the attestation being ingested normally as an OCI Artifact Blob.

pxp928:
@lumjjb could you take a look at this when you get a chance?

pxp928 (approved):
This looks good to me! Thanks @robert-cronin

pxp928:
@robert-cronin sorry for the delay in review. Also, a rebase will fix the tilt ci issue.

mihaimaruseac (approved):

PR #2605, opened 2025-04-14
Increase ClearlyDefined rate limit interval

### LINKED ISSUE #2603: [bug] ClearlyDefined definitions rate limit interval should be 1 minute
It looks to me like the ClearlyDefined API client that hits their `/definitions` endpoint should have a rate limit interval of 1 minute per their docs: https://docs.clearlydefined.io/docs/get-involved/using-data#production-instance-rate-limits .

Image

But I think that the code in the v0.14.0 uses 30 seconds instead: https://github.com/guacsec/guac/blob/v0.14.0/pkg/certifier/clearlydefined/clearlydefined.go#L46-L70

Probably this should be adjusted.

nchelluri:
- The interval should be 1m (not 30s) per their API docs.
- Fixes #2603

# PR Checklist

### DIFF
--- a/pkg/certifier/clearlydefined/clearlydefined.go
+++ b/pkg/certifier/clearlydefined/clearlydefined.go
@@ -44,8 +44,12 @@ import (
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
-var rateLimit = 250
-var rateLimitInterval = 30 * time.Second
+
+// Rate limits: https://docs.clearlydefined.io/docs/get-involved/using-data#production-instance-rate-limits
+var (
+	rateLimit         = 250
+	rateLimitInterval = time.Minute / time.Duration(rateLimit)
+)
 
 const (
 	PRODUCER_ID string = "guacsec/guac"


mihaimaruseac (approved):

nchelluri:
It is possible that I misunderstood how the rate limiting package interprets the interval; I will double-check tomorrow.

nchelluri:
Thanks for the reviews. I actually think the interval should be `time.Minute/250` at this point in time after trying to read the docs at https://pkg.go.dev/golang.org/x/time/rate a bit more closely and doing some testing. That way the limiter could be created like so:

```go
			// We will put 250 tokens into the bucket every minute. And we should not send more than
			// 250 definition queries in a burst.
			Limiter: rate.NewLimiter(rate.Every(time.Minute/250), 250),
```

Does that seem ok? If it does, I will update this PR (the linked issue, created from my previous understanding, is slightly wrong - but this new code is correct, I think).

pxp928 (approved):

mihaimaruseac:
Kodiak is stuck, will merge manually.

PR #2752, opened 2025-09-08
vulnerability is not getting ingested #2751

### LINKED ISSUE #2751: [ingestion/data-quality issue] OSV certifier not ingesting data when error in severity parsing
**Describe the bug**
When the OSV certifier is run with the `--add-vuln-metadata` flag, it tries to hydrate vulnerability data with more metadata and creates an ingestion request for each vulnerability. There are the following issues with the current implementation of the parser package.

1. The vulnerability metadata contains a structure like follows
```"severity":[{"type":"CVSS_V3","score":"CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H"}]``` 
The parsing logic relies on the "type" field, which is not correct (or at least incomplete), so parsing fails for almost all the scores due to it matching the wrong `case` block here https://github.com/guacsec/guac/blob/main/pkg/ingestor/parser/vuln/score.go#L24

This could be fixed by instead matching using the `strings.startswith` as in this example https://pkg.go.dev/github.com/pandatix/go-cvss@v0.6.2#section-readme

2. Another issue is, the ingestion of the entire vulnerability is dropped if the parsing of the severity score fails. Instead, it might be better to ingest at least the vulnerability_id along with whatever metadata is successfully parsed. And incrementally add support for more severity score types.

**To Reproduce**
Try to ingest an SBOM with osv certifier enabled and `--add-vuln-metadata` flag enabled. And compare the number of vulnerabilities ingested with [osv-scanner](https://google.github.io/osv-scanner/) generated vulnerability for the same SBOM

**Expected behavior**

**GUAC version**
v1.0.0

**Ingested document(s)**

I will try to attach the SBOM I was using but will have to check for any propritory information before uploading

shreyasHpandya:
when you enable --add-vuln-metadata flag for osv certifier, it is encountering issues when parsing socre due to score menthod mismatch between the supplied value from OSV and the actual score in case of CVSS socres. Also there are unsupported socre types like "ubuntu". All these results in score parsing error which results in entire vulnerability being skipped from ingestion, instead more correct thing to do might be to ingest atleast the vulnerability id and whichever scores are successfully parsed and log the warning massages for unsupported scores. fixes #2751

NOTE: this is a draft PR to get review for the approach, will refine the code further if everyone is fine with the proposed approach

# PR Checklist

### DIFF
--- a/pkg/ingestor/parser/vuln/score.go
+++ b/pkg/ingestor/parser/vuln/score.go
@@ -2,8 +2,8 @@ package vuln
 
 import (
 	"strconv"
+	"strings"
 
-	"github.com/guacsec/guac/pkg/assembler/clients/generated"
 	attestation_vuln "github.com/guacsec/guac/pkg/certifier/attestation/vuln"
 	gocvss20 "github.com/pandatix/go-cvss/20"
 	gocvss30 "github.com/pandatix/go-cvss/30"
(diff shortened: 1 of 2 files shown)


shreyasHpandya:
Will cerate new PR

PR #2815, opened 2025-11-07
feat: adds custom scorecard certifier

### LINKED ISSUE #2783: [feature] Add API-based scorecard fetcher as an alternative to existing scorecard certifier
The current GUAC scorecard certifier only supports computing of OpenSSF, which hits GitHub API rate limits when processing large databases.  Additionally, local execution requires GitHub tokens (which often hits rate limits), and significant compute resources, making it unsuitable for scalable GUAC operations.

Add an API-based scorecard fetcher using the OpenSSF Scorecard REST API to bypass GitHub rate limits. Users can choose between "query" (default) and "compute" modes of scorecard via CLI flags. The query mode of scorecard requires no GitHub tokens, fetches pre-computed results from `https://api.securityscorecards.dev`, and includes configurable timeouts/endpoints. Maintains full backward compatibility with existing compute mode of scorecard.

shreyasHpandya:
Fixes #2783

# Description of the PR

See comment https://github.com/guacsec/guac/pull/2815#issuecomment-3753670486

# PR Checklist

### DIFF
--- a/pkg/certifier/scorecard/scorecard.go
+++ b/pkg/certifier/scorecard/scorecard.go
@@ -24,6 +24,7 @@ import (
 	"github.com/guacsec/guac/pkg/certifier"
 	"github.com/guacsec/guac/pkg/certifier/components/source"
 	"github.com/guacsec/guac/pkg/events"
+	"github.com/guacsec/guac/pkg/logging"
 	"github.com/ossf/scorecard/v5/docs/checks"
 	"github.com/ossf/scorecard/v5/log"
 	sc "github.com/ossf/scorecard/v5/pkg/scorecard"
@@ -111,7 +112,10 @@ func NewScorecardCertifier(sc Scorecard) (certifier.Certifier, error) {
 	// check if the GITHUB_AUTH_TOKEN is set
 	s, ok := os.LookupEnv("GITHUB_AUTH_TOKEN")
 	if !ok || s == "" {
-		return nil, fmt.Errorf("GITHUB_AUTH_TOKEN is not set")
+		// Log warning but allow initialization without token
+		// The API path will still work, only local computation will fail
+		logger := logging.FromContext(context.Background())
+		logger.Warnf("GITHUB_AUTH_TOKEN not set - scorecard API will work, but local computation fallback will be disabled")
 	}
 
 	return &scorecard{
--- a/pkg/certifier/scorecard/scorecard.go
+++ b/pkg/certifier/scorecard/scorecard.go
@@ -102,7 +102,8 @@ func (s scorecard) CertifyComponent(_ context.Context, rootComponent interface{}
 }
 
 // NewScorecardCertifier initializes the scorecard certifier.
-// It checks if the GITHUB_AUTH_TOKEN is set in the environment. If it is not, it returns an error.w
+// It checks if the GITHUB_AUTH_TOKEN is set in the environment. If it is not,
+// a warning is logged; the scorecard API path still works without the token.
 // The token is used to access the GitHub API, https://github.com/ossf/scorecard#authentication.
 func NewScorecardCertifier(sc Scorecard) (certifier.Certifier, error) {
 	if sc == nil {
--- a/pkg/certifier/scorecard/scorecardRunner.go
+++ b/pkg/certifier/scorecard/scorecardRunner.go
@@ -18,19 +18,148 @@ package scorecard
 import (
 	"context"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
 
+	"github.com/guacsec/guac/pkg/logging"
 	"github.com/ossf/scorecard/v5/checker"
 	"github.com/ossf/scorecard/v5/checks"
 	"github.com/ossf/scorecard/v5/log"
 	sc "github.com/ossf/scorecard/v5/pkg/scorecard"
 )
 
+const githubPrefix = "github.com/"
+
 // scorecardRunner is a struct that implements the Scorecard interface.
 type scorecardRunner struct {
 	ctx context.Context
 }
 
+// normalizeRepoName ensures the repo name has the github.com/ prefix required by the Scorecard API.
+// Adds the prefix if it is missing; otherwise returns the repo name unchanged.
+func normalizeRepoName(repoName string) string {
+	if strings.HasPrefix(repoName, githubPrefix) {
+		return repoName
+	}
+	return githubPrefix + repoName
+}
+
 func (s scorecardRunner) GetScore(repoName, commitSHA, tag string) (*sc.Result, error) {
+	logger := logging.FromContext(s.ctx)
+	repoName = normalizeRepoName(repoName)
+
+	// First try API approach
+	logger.Debugf("Attempting to fetch scorecard from API for repo: %s, commit: %s", repoName, commitSHA)
+	result, err := s.getScoreFromAPI(repoName, commitSHA, tag)
+	if err == nil {
+		logger.Infof("Successfully fetched scorecard from API for repo: %s", repoName)
+		return result, nil
+	}
+
+	// Log API failure and check if we can fallback to local computation
+	logger.Warnf("API fetch failed for repo %s: %v", repoName, err)
+
+	// Check if GitHub token is available for local computation
+	if _, ok := os.LookupEnv("GITHUB_AUTH_TOKEN"); !ok {
+		logger.Errorf("Cannot fall back to local computation - GITHUB_AUTH_TOKEN not set")
+		return nil, fmt.Errorf("scorecard API failed and GITHUB_AUTH_TOKEN not available for local computation: %w", err)
+	}
+
+	logger.Infof("Falling back to local computation for repo: %s", repoName)
+	result, err = s.computeScore(repoName, commitSHA, tag)
+	if err != nil {
+		logger.Errorf("Failed to compute scorecard locally for repo %s: %v", repoName, err)
+	}
+	return result, err
+}
+
+func (s scorecardRunner) getScoreFromAPI(repoName, commitSHA, tag string) (*sc.Result, error) {
+	logger := logging.FromContext(s.ctx)
+
+	// If tag is provided without a valid commitSHA, skip API and use local computation
+	// The API cannot resolve tags, but computeScore can look up the commit for a tag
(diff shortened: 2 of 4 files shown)


gaganhr94:
## Scorecard improvements description for better understanding

When querying the Scorecard API with a specific commitSHA that returns 404,
retry the request without the commit parameter to get the latest scorecard
for the default branch.

### Design Decisions:

- API-First with Retry: If a commit-specific query fails, retry without
commitSHA rather than immediately falling back to local computation.
This maximizes API usage (faster, no auth required).

- Silent Fallback for Missing Commits: When the requested commit isn't in
the API, we return the latest scorecard instead. This is a conscious
decision because:
  - Scorecard's purpose is to assess current repository health and
maintenance practices, not point-in-time snapshots
  - A recent scorecard provides more value than no scorecard for trust
decisions
  - Many checks (branch protection, security policy) are repository-wide
and don't vary between commits
- Tag Without CommitSHA Skips API: When a tag is provided without a
commitSHA, skip the API entirely and use local computation, which can
resolve tags to commits via ListReleases().

- HEAD Treated as Empty: Upstream GUAC code sets commitSHA to "HEAD" when
no specific commit is known, so we treat "HEAD" the same as empty.

gaganhr94:
Hi @mihaimaruseac @pxp928, can you please review this PR when you get a chance. Thanks !

pkg/certifier/scorecard/scorecard.go
```diff
@@ -110,11 +111,14 @@ func NewScorecardCertifier(sc Scorecard) (certifier.Certifier, error) {
...
+		logger := logging.FromContext(context.Background())
+		logger.Warnf("GITHUB_AUTH_TOKEN not set - scorecard API will work, but local computation fallback will be disabled")
 	}
 
 	return &scorecard{
 		scorecard: sc,
 		ghToken:   s,
 	}, nil
-}
+}
```
mihaimaruseac: The file should end with a newline

pkg/certifier/scorecard/scorecard.go
```diff
@@ -39,7 +40,7 @@ type scorecard struct {
 var ErrArtifactNodeTypeMismatch = fmt.Errorf("rootComponent type is not *source.SourceNode")
 
 // CertifyComponent is a certifier that generates scorecard attestations
-func (s scorecard) CertifyComponent(_ context.Context, rootComponent interface{}, docChannel chan<- *processor.Document) error {
+func (s scorecard) CertifyComponent(ctx context.Context, rootComponent interface{}, docChannel chan<- *processor.Document) error {
```
mihaimaruseac: `ctx` is still unused... Did you try to compile the code? The compiler would have given you an error.

mihaimaruseac (changes requested):

pkg/certifier/scorecard/scorecard.go
```diff
@@ -110,11 +111,14 @@ func NewScorecardCertifier(sc Scorecard) (certifier.Certifier, error) {
 	// check if the GITHUB_AUTH_TOKEN is set
 	s, ok := os.LookupEnv("GITHUB_AUTH_TOKEN")
 	if !ok || s == "" {
-		return nil, fmt.Errorf("GITHUB_AUTH_TOKEN is not set")
+		// Log warning but allow initialization without token
+		// The API path will still work, only local computation will fail
```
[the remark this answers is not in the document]
gaganhr94: Previously, a GitHub token was required for Scorecard to start. After this change, startup no longer depends on the token, so just logs a warning.  Added these comments intentionally to make that behaviour change explicit and improve transparency for future readers. If there’s a clearer way to convey this, I’m happy to adjust the wording.

mihaimaruseac (approved):

gaganhr94:
Hi @mlieberman85 can you please review this PR when you get a chance. Thanks !

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -18,19 +18,148 @@ package scorecard
...
+	if commitSHA != "" && commitSHA != "HEAD" {
+		urlWithCommit := baseURL + "?commit=" + commitSHA
+		result, err := s.fetchFromAPI(urlWithCommit)
+		if err == nil {
+			return result, nil
+		}
+		logger.Debugf("API call with commit %s failed, retrying without commit: %v", commitSHA, err)
+	}
+
+	result, err := s.fetchFromAPI(baseURL)
```
mlieberman85, line 105: This is the "silent fallback to latest scorecard" path called out in the description, and it deserves a second look. If a caller asks for the scorecard at commit X and the API returns 404 for that commit, we silently return the latest scorecard for the default branch. That is a different answer to a different question.

The design-decision rationale ("scorecard reflects current repo health, not point-in-time snapshots") is fine as a default, but GUAC downstream consumers assume that when they pass a commitSHA, the returned scorecard corresponds to that commit. This silent substitution will flow through to attestations/assertions that get tied to the wrong provenance.

Two concrete asks:

1. **Record in the result that this was a fallback.** Either annotate the returned scorecard (e.g., set commit to empty / "latest") or wrap it so downstream code can detect the mismatch. Right now there is no way for the caller to know they got something different than they asked for.
2. **Make it opt-in, or at least opt-out-able.** A `StrictCommit` option (default false to preserve current behavior once this merges, or default true if you want to be conservative) would let operators who care about commit-pinned scorecards refuse the fallback. This matters for anyone building attestations off scorecard data.

mlieberman85 (commented):
Three things worth considering before merge, all around trust boundaries and caller expectations. Not blocking — Mihai's approval stands, I'm offering a second pass.

FYI on the E2E check: the failing `PkgQ4` diff (`sha256%3A` vs `sha256:`) is a pre-existing issue on `main`. The last 5 `main` CI runs fail at the identical PURL-encoding diff, so don't block on debugging CI on this PR — it's not caused by these changes.

pkg/certifier/scorecard/scorecard.go
```diff
@@ -111,7 +112,10 @@ func NewScorecardCertifier(sc Scorecard) (certifier.Certifier, error) {
...
+		logger := logging.FromContext(context.Background())
+		logger.Warnf("GITHUB_AUTH_TOKEN not set - scorecard API will work, but local computation fallback will be disabled")
```
mlieberman85, line 118: Changing this from a hard error to a warning is a subtle behavior change for existing callers. Today, `NewScorecardCertifier` acts as a configuration validator: if `GITHUB_AUTH_TOKEN` is unset, construction fails loudly at init. With this change, construction succeeds and the failure moves to first use, which is harder to detect in config/health checks.

Fine as a product decision (the API path really does not need the token), but:

1. **Update the docstring.** Line 105 still says *"If it is not, it returns an error."* That is no longer true.
2. **Consider leaving the hard error as the default and adding an explicit opt-in** like `WithAPIOnly()` or a functional option. Operators who rely on the current fail-fast behavior to catch misconfiguration should not silently lose that guarantee.

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -18,19 +18,148 @@ package scorecard
...
+		return nil, err
+	}
+	return result, nil
+}
+
+func (s scorecardRunner) fetchFromAPI(apiURL string) (*sc.Result, error) {
+	logger := logging.FromContext(s.ctx)
+	logger.Debugf("Making API request to: %s", apiURL)
+
+	httpClient := &http.Client{
```
mlieberman85, line 116: Two minor additions worth making before merge:

1. **Retry with backoff on 5xx and 429.** `api.securityscorecards.dev` is rate-limited; a bulk GUAC run will hit it, and right now any 429 propagates straight up as an error instead of retrying.
2. **Respect `Retry-After`** when the API returns 429. Same reason.

A small wrapper around the `http.Client` or a dependency like `hashicorp/go-retryablehttp` would cover both cleanly without more complexity.

shreyasHpandya:
> Three things worth considering before merge, all around trust boundaries and caller expectations. Not blocking — Mihai's approval stands, I'm offering a second pass.
> 
> FYI on the E2E check: the failing `PkgQ4` diff (`sha256%3A` vs `sha256:`) is a pre-existing issue on `main`. The last 5 `main` CI runs fail at the identical PURL-encoding diff, so don't block on debugging CI on this PR — it's not caused by these changes.

Thank you, @mlieberman85, for the review. These are very good points. The following are the changes I want to propose based on your inputs.

- Record in the scorecard table to indicate if this is a fallback. We will look into the existing structure if this can be accommodated. or explore other options
- Have a flag to opt out. If we are able to capture in the data model that a fallback was used, then it is okay to have fallback as the default, with the option to opt out.
- `GITHUB_AUTH_TOKEN` being optional is breaking the expectation, but I think this is a change in good direction in terms of user experience, considering users quickly trying out GUAC, and doing POCs, etc. – which is what most current user base is doing with GUAC – would have less friction trying out scorecard functionality. The new scorecard API approach would work for most POCs. I would go so far as to make the local computation fallback opt in. (which you already alluded to in your comment). We will update the docstring.
- Retry with backoff on 5xx and 429. is worth adding we will work on this one.

Also, my one suggestion would be to merge this changes (after the docstring change). And do the rest of the changes as separate enhancement which we will shortly contribute. Let me know your thoughts.

Thanks

pkg/certifier/scorecard/scorecardRunner.go
gaganhr94: We have implemented the retry with backoff for the Scorecard API. The changes are a part of this PR now.

gaganhr94:
Hi @mlieberman85, we've made the changes to add retry-with-backoff on 429 and 5xx in this PR. The other two suggestions — 1) StrictCommit option, and 2) WithAPIOnly() opt-in — we'll create separate follow-up issues for these and track them there.

pkg/certifier/scorecard/scorecardRunner.go
gaganhr94: Issue #3014 created to track this

pkg/certifier/scorecard/scorecard.go
gaganhr94: The docstring has been fixed. For the recommendation of having an opt-in like `WithAPIOnly`, have created a new issue #3013.

mlieberman85 (approved):
LGTM

PR #2850, opened 2026-02-11
fix: handle FK race condition in bulk upsert by using UpdateNewValues for PackageName

dgellman:
## Summary

- Fix FK constraint race condition in `upsertBulkPackage` where concurrent transactions cause `package_versions_package_names_versions` FK violations
- Switch PackageName upsert from `DoNothing()` to `UpdateNewValues()` so PostgreSQL uses `DO UPDATE` (which blocks on concurrent uncommitted rows) instead of `DO NOTHING` (which silently skips)
- Fix `ErrNoRows` handling in all 17 remaining bulk `DoNothing()` paths to prevent spurious failures during re-ingestion

## Root Cause

Under PostgreSQL READ COMMITTED isolation, `ON CONFLICT DO NOTHING` does not block waiting for a concurrent transaction's uncommitted row -- it silently skips the insert. When `upsertBulkPackage` inserts PackageNames (Step 1) then PackageVersions referencing them (Step 2), a concurrent transaction can skip a PackageName in Step 1 and then fail the FK check in Step 2 because the row is not yet committed by the other transaction.

`ON CONFLICT DO UPDATE` acquires a `FOR NO KEY UPDATE` row lock and blocks until the conflicting transaction commits, ensuring the parent row is visible when the child insert runs.

## Changes

**package.go** (core FK race fix):
- PackageName bulk + single inserts: `DoNothing()` → `UpdateNewValues()` (generates `DO UPDATE SET col=EXCLUDED.col`)
- PackageVersion bulk insert: added `ErrNoRows` filter (retains `DoNothing()` since it's the leaf entity)

**17 other files** (ErrNoRows fix for bulk `DoNothing()` paths):
- Added `&& err != stdsql.ErrNoRows` to bulk upsert error checks
- Added `stdsql "database/sql"` import where missing
- Files: artifact.go, builders.go, certify.go, certifyVEXStatement.go, dependency.go, hasMetadata.go, hashequal.go, license.go, occurrence.go, pkgequal.go, pointOfContact.go, scorecard.go, slsa.go, source.go, vulnEqual.go, vulnMetadata.go, vulnerability.go

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./pkg/assembler/backends/ent/backend/...` passes
- [ ] Verify concurrent `ingestPackages` calls no longer produce FK violations
- [ ] Verify idempotent bulk re-ingestion succeeds without errors
- [ ] Verify OSV certifier bulk path completes successfully under concurrent load

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mihaimaruseac:
Did you compile the project? Did you run it?

dgellman:
> Did you compile the project? Did you run it?

Sry - i meant to put this in draft mode.  Having some network connectivity issues just got connectivity back.

dgellman:
Ok - built, all unit tests are passing, running locally, and working.  Lmk if there is anything else I need to do or if this is good for review.  Happy to make any changes necessary.  

FYI, I discovered these while working through a PoV of my own, so hopefully we can work through them and get them approved.

mihaimaruseac (approved):

dgellman:
Hey guys - just wanted to check back in to see if there is anything else I need to do for approvals? :)

PR #2974, opened 2026-04-13
feat: implements cloud-agnostic blob collector

### LINKED ISSUE #1320: [feature] Azure Blob Storage collector
Similar to #1020 and the [GCS collector](https://github.com/guacsec/guac/tree/main/pkg/handler/collector/gcs), guac can collect from [Azure Blob Storage](https://learn.microsoft.com/en-us/azure/storage/blobs/storage-blobs-overview)

There can be an integration into [Azure Event Grid](https://learn.microsoft.com/en-us/azure/event-grid/overview) to notify guac of a [new blob creation event](https://learn.microsoft.com/en-us/azure/event-grid/event-schema-blob-storage?tabs=cloud-event-schema#blob-storage-events) to go trigger a collection.

gaganhr94:
# Description of the PR

This PR adds a cloud agnostic blob store collector

Fixes #1320 

# PR Checklist

- [X] All CI checks are passing (tests and formatting)
- [X] All dependent PRs have already been merged

### DIFF
--- a/pkg/handler/collector/blob/blob.go
+++ b/pkg/handler/collector/blob/blob.go
@@ -36,12 +36,22 @@ import (
 
 const CollectorBlob = "BlobCollector"
 
+// DefaultMaxObjectSize caps the per-object read in bytes when no explicit
+// limit is configured, defaults to 100 MiB.
+const DefaultMaxObjectSize int64 = 100 * 1024 * 1024
+
+// ErrObjectTooLarge is returned by getObject when the object's size
+// exceeds maxObjectSize. Callers log and skip; this is not a fatal error.
+var ErrObjectTooLarge = errors.New("object exceeds max object size")
+
 type blobCollector struct {
-	url          string
-	bucket       *blob.Bucket
-	lastDownload time.Time
-	poll         bool
-	interval     time.Duration
+	url           string
+	bucket        *blob.Bucket
+	lastDownload  time.Time
+	poll          bool
+	interval      time.Duration
+	prefix        string
+	maxObjectSize int64
 }
 
 type Opt func(*blobCollector)
@@ -65,6 +75,20 @@ func WithPolling(interval time.Duration) Opt {
 	}
 }
 
+// WithPrefix limits collection to objects whose key begins with prefix.
+func WithPrefix(prefix string) Opt {
+	return func(b *blobCollector) {
+		b.prefix = prefix
+	}
+}
+
+// WithMaxObjectSize caps the number of bytes read per object.
+func WithMaxObjectSize(n int64) Opt {
+	return func(b *blobCollector) {
+		b.maxObjectSize = n
+	}
+}
+
 // NewBlobCollector creates a cloud-agnostic collector that can collect
 // documents from any blob storage supported by gocloud.dev/blob (S3, GCS,
 // Azure Blob Storage, filesystem, etc).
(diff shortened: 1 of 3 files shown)


gaganhr94:
Unit tests failing due to flakiness. Rerun should resolve the workflow failure

mihaimaruseac (approved):

mlieberman85 (commented):
A couple of things worth addressing before merge, both around bounding what this collector will read and from where. The CI unit/E2E failures look like GH runner disk-exhaustion (No space left on device), not code issues, so just a rerun should green them up.

pkg/handler/collector/blob/blob.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func (b *blobCollector) getObject(ctx context.Context, key string) ([]byte, error) {
+	reader, err := b.bucket.NewReader(ctx, key, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create reader for %q: %w", key, err)
+	}
+	defer func() { _ = reader.Close() }()
+
+	data, err := io.ReadAll(reader)
```
mlieberman85, line 183: No size cap on the read. A single oversized or adversarial object will OOM the collector, and since we hand bucket URLs off to operators, we can't assume the bucket only contains well-formed SBOMs.

Suggest wrapping with `io.LimitReader` and plumbing a `--max-object-size` flag (default maybe 100MB, since SBOMs can legitimately be large). If the limit is hit, log and skip rather than fail the whole run, same way the existing warn-and-continue handles read errors below.

pkg/handler/collector/blob/blob.go
```diff
@@ -0,0 +1,196 @@
...
+			}
+		}
+	}
+
+	return getArtifacts()
+}
+
+func (b *blobCollector) getArtifacts(ctx context.Context, docChannel chan<- *processor.Document) error {
+	logger := logging.FromContext(ctx)
+	iter := b.bucket.List(nil)
```
mlieberman85, line 132: This pulls every object in the bucket with no way to scope it down. We need one of:

1. **Filtering support.** A `--prefix` flag (simplest: pass to `blob.ListOptions.Prefix`; better: wrap with `blob.PrefixedBucket` so the scope is enforced at the bucket level, not just at list-time).
2. **Clearly document** that the bucket must contain *only* documents GUAC can ingest. Right now the `Long` description implies any bucket URL works, which invites operators to point this at a shared bucket and get weird results (or ingest content they didn't mean to).

(1) is the real fix. If (1) is out of scope for this PR, at least (2) needs to land here.

pkg/handler/collector/blob/blob.go
gaganhr94: Valid point. I have added a --max-object-size flag with a default value

pkg/handler/collector/blob/blob.go
gaganhr94: Added a filtering support in this PR itself since I see it adding value. Also clearly mentioned in the description that the bucket must only contain documents which the GUAC can ingest

mlieberman85 (approved):
Thanks for the thorough response. Size cap, prefix flag, and docs all land cleanly, and the tests explicitly check for leaks / skipping — nice.

One forward-looking thought, non-blocking: if this collector ever grows beyond list→read to anything that could take an arbitrary key (e.g. a future single-object mode, or retry against a specific object), wrapping with `blob.PrefixedBucket` would enforce the scope at the bucket level instead of at list-time. For the current shape, `ListOptions.Prefix` is fine.

E2E red is the pre-existing `PkgQ4` issue on main, not this PR. Approving.

PR #2991, opened 2026-04-16
feat: adding support of attaching labels to documents in guaccollect

### LINKED ISSUE #2990: [feature] Support attaching labels to documents during guaccollect ingestion

When ingesting SBOMs at scale via guaccollect, there is no way to attach contextual metadata (labels) to the documents being ingested. This is a significant gap for users that need to classify, tag, or enrich artifacts at ingestion time — for example, tagging SBOMs with the product line they belong to, the environment they were built for, or the release they are part of.

When the metadata about an artifact is known at ingestion time (the pipeline knows which product, team, or environment an SBOM belongs to), forcing users to re-discover and re-associate this information after ingestion is redundant and inefficient.

Add a `--label key=value` flag to guaccollect that can be specified multiple times. Labels provided at collection time should be attached to the processor.Document so they flow through the entire ingestion pipeline (collector → blob store → event stream → ingestor).

Usage example
```
guaccollect files \
--label releaseVersion=1.24.325 \
--label releaseName=Panda \
./sboms/
```

This would ingest all SBOMs in ./sboms/ and automatically attach three HasMetadata nodes to each package/artifact that has an SBOM relationship, with the key-value pairs provided. This applies uniformly across all collector subcommands: files, oci, s3, gcs, github, deps_dev, and kubescape.

- **Post-ingestion GraphQL mutations** — This breaks the scalability model of guaccollect and introduces race conditions. It also requires maintaining a separate metadata-attachment service or script.
- **Scripted workflows using guacone** — While guacone operates synchronously, it doesn't scale the way guaccollect does. It lacks the event-stream decoupling, blob store integration, and collector plugin architecture that make guaccollect suitable for large-scale production deployments.

**Additional context**

This feature is particularly valuable for users managing large software portfolios where hundreds or thousands of SBOMs are ingested from diverse sources. Being able to tag artifacts at ingestion time with details like the product names, release versions, team ownership, environments) enables users to get insights like:
- "Which teams are affected by a certain vulnerability?"
- "Which packages in a certain product have known CVEs?"

Without ingestion-time labeling, these queries require external joins against systems outside GUAC, negating much of the value of having a unified software supply chain graph.

gaganhr94:
# Description of the PR

Fixes #2990

# PR Checklist

### DIFF
--- a/cmd/guaccollect/cmd/deps_dev.go
+++ b/cmd/guaccollect/cmd/deps_dev.go
@@ -133,7 +133,7 @@ you have access to read and write to the respective blob store.`,
 			}()
 		}
 
-		initializeNATsandCollector(ctx, opts.pubsubAddr, opts.blobAddr, opts.publishToQueue)
+		initializeNATsandCollector(ctx, opts.pubsubAddr, opts.blobAddr, opts.publishToQueue, nil)
 	},
 }
 
--- a/cmd/guaccollect/cmd/files.go
+++ b/cmd/guaccollect/cmd/files.go
@@ -46,6 +46,8 @@ type filesOptions struct {
 	poll bool
 	// enable/disable message publish to queue
 	publishToQueue bool
+	// labels to attach to collected documents as HasMetadata
+	labels map[string]string
 }
 
 var filesCmd = &cobra.Command{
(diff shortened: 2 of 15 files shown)


mihaimaruseac (approved):

mlieberman85 (approved):
LGTM. Clean implementation, good test coverage, consistent across all collectors.

Minor nit: `Labels` field in `processor.go` isn't tab-aligned with the other struct fields.

```go
Labels            map[string]string  `json:"labels,omitempty"`
```

PR #2994, opened 2026-04-17
feat(guacrest): implement GET /v0/package/{purl}

### LINKED ISSUE #2993: Implement GET /v0/package/{purl}
Currently there is no implementation for this endpoint

gaganhr94:
Replaces the 501 stub for getPackagePurls with a real handler that resolves partial purls against the graph.

# Description of the PR

Fixes #2993

# PR Checklist

### DIFF
--- a/pkg/guacrest/server/server.go
+++ b/pkg/guacrest/server/server.go
@@ -102,9 +102,33 @@ func (s *DefaultServer) AnalyzeDependencies(ctx context.Context, request gen.Ana
 }
 
 func (s *DefaultServer) GetPackagePurls(ctx context.Context, request gen.GetPackagePurlsRequestObject) (gen.GetPackagePurlsResponseObject, error) {
-	return gen.GetPackagePurls500JSONResponse{
-		InternalServerErrorJSONResponse: gen.InternalServerErrorJSONResponse{
-			Message: "GetPackagePurls not implemented",
+	unescapedPurl, err := url.PathUnescape(request.Purl)
+	if err != nil {
+		return gen.GetPackagePurls400JSONResponse{
+			BadRequestJSONResponse: gen.BadRequestJSONResponse{
+				Message: fmt.Sprintf("failed to unescape package url: %v", err),
+			},
+		}, nil
+	}
+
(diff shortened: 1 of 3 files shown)


mihaimaruseac (approved):

pkg/guacrest/server/server.go
```diff
@@ -102,9 +102,33 @@ func (s *DefaultServer) AnalyzeDependencies(ctx context.Context, request gen.Ana
...
+			BadRequestJSONResponse: gen.BadRequestJSONResponse{
+				Message: err.Error(),
+			},
+		}, nil
+	}
+
+	totalCount := len(purls)
+	return gen.GetPackagePurls200JSONResponse{
+		PurlListJSONResponse: gen.PurlListJSONResponse{
+			PaginationInfo: gen.PaginationInfo{TotalCount: &totalCount},
```
mlieberman85, line 130: The response structure acknowledges pagination (`PaginationInfo` with `NextCursor` + `TotalCount` in the OpenAPI spec), but the handler returns all matches in one response and never sets `NextCursor`. For a query like `pkg:npm/react`, that's potentially thousands of purls materialized in memory and serialized in one response.

Root cause is that `getPackagePurls` in `openapi.yaml` doesn't declare `pageSize` / `cursor` query params, so this PR inherits the gap rather than creating it. Two options:

1. Add pagination params here: update the OpenAPI op, thread `pageSize`/`cursor` through `FindMatchingPurls`, set `NextCursor` when there's more. Most correct, bigger diff.
2. Land as-is and file a follow-up issue to add pagination across all `PurlList`-returning endpoints (`/package/{purl}/dependencies` has the same gap, and `/package/{purl}/vulns` too).

Leaning (2) since it's a cross-cutting API concern, but flagging so it doesn't get lost.

mlieberman85 (approved):
Approving. One thought on pagination left inline — non-blocking, flagging for awareness.

E2E red is the pre-existing `PkgQ4` flake on main, unrelated.

pkg/guacrest/server/server.go
gaganhr94: Will create a follow up issue for this

PR #3017, opened 2026-04-23
fix: remove deprecated in-toto-golang dependency

### LINKED ISSUE #2036: [bug] in-toto-golang is being deprecated, move to in-toto attestations
**Describe the bug**
in-toto-golang is being deprecated. We need to move the guesser, processor, certifier and parsers to use the new in-toto attestation statement headers.

Some of that work was done in this PR: https://github.com/guacsec/guac/pull/1988, but the guesser and other locations have not been updated.

**GUAC version**
main

Abhishek9639:
# Description of the PR

Removes the deprecated `in-toto-golang` library from the codebase. The guesser, processor, and parser were still importing types from `in-toto-golang` this PR replaces those with local type definitions.

**Changes:**
- `type_ite6.go`: replaced `in_toto.Statement` / `attestationv1.Statement` with a local `ite6Statement` struct that handles both v0.1 and v1 in-toto statement formats
- `ite6.go`: same swapped `attestationv1.Statement` for a local struct
- `parser_slsa.go`: defined local `DigestSet`, `ProvenanceMaterial`, `ProvenancePredicateV01/V02` types to replace `scommon`, `slsa01`, `slsa02` imports
- `parser_slsa_test.go` / `sigstore_verifier_test.go`: updated to use the new local types
- `go.mod`: removed `in-toto-golang` as a direct dependency

Fixes #2036

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) they are generated using `-s` flag to `git commit`.

### DIFF
--- a/pkg/handler/processor/guesser/type_ite6.go
+++ b/pkg/handler/processor/guesser/type_ite6.go
@@ -35,19 +35,32 @@ type ite6Statement struct {
 }
 
 // getType returns the statement type from whichever format was used.
+// in-toto v1 is the current standard; v0.1 is supported for backwards
+// compatibility. A document that populates both _type and type is malformed.
 func (s *ite6Statement) getType() string {
-	if s.TypeV01 != "" {
-		return s.TypeV01
+	if s.TypeV1 != "" && s.TypeV01 != "" {
+		// Both fields set: reject the ambiguous/malformed document.
+		return ""
 	}
-	return s.TypeV1
+	if s.TypeV1 != "" {
+		return s.TypeV1
+	}
+	return s.TypeV01
 }
 
 // getPredicateType returns the predicate type from whichever format was used.
+// in-toto v1 is the current standard; v0.1 is supported for backwards
+// compatibility. A document that populates both predicateType and predicate_type
+// is malformed.
 func (s *ite6Statement) getPredicateType() string {
-	if s.PredicateTypeV01 != "" {
-		return s.PredicateTypeV01
+	if s.PredicateTypeV1 != "" && s.PredicateTypeV01 != "" {
+		// Both fields set: reject the ambiguous/malformed document.
+		return ""
 	}
-	return s.PredicateTypeV1
+	if s.PredicateTypeV1 != "" {
+		return s.PredicateTypeV1
+	}
+	return s.PredicateTypeV01
 }
 
 type ite6TypeGuesser struct{}
--- a/pkg/handler/processor/guesser/type_ite6.go
+++ b/pkg/handler/processor/guesser/type_ite6.go
@@ -21,40 +21,53 @@ import (
 	jsoniter "github.com/json-iterator/go"
 
 	"github.com/guacsec/guac/pkg/handler/processor"
-	attestationv1 "github.com/in-toto/attestation/go/v1"
-	"github.com/in-toto/in-toto-golang/in_toto"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
+// ite6Statement supports unmarshalling both v0.1 ("_type", "predicateType")
+// and v1 ("type", "predicate_type") in-toto statement formats.
+type ite6Statement struct {
+	TypeV01          string `json:"_type"`
+	PredicateTypeV01 string `json:"predicateType"`
+	TypeV1           string `json:"type"`
+	PredicateTypeV1  string `json:"predicate_type"`
+}
+
+// getType returns the statement type from whichever format was used.
+func (s *ite6Statement) getType() string {
+	if s.TypeV01 != "" {
+		return s.TypeV01
+	}
+	return s.TypeV1
+}
+
+// getPredicateType returns the predicate type from whichever format was used.
+func (s *ite6Statement) getPredicateType() string {
+	if s.PredicateTypeV01 != "" {
+		return s.PredicateTypeV01
+	}
+	return s.PredicateTypeV1
+}
+
 type ite6TypeGuesser struct{}
 
 func (_ *ite6TypeGuesser) GuessDocumentType(blob []byte, format processor.FormatType) processor.DocumentType {
-	var statement in_toto.Statement
+	var statement ite6Statement
 	if json.Unmarshal(blob, &statement) == nil && format == processor.FormatJSON {
-		if strings.HasPrefix(statement.Type, "https://in-toto.io/Statement") {
-			if strings.HasPrefix(statement.PredicateType, "https://slsa.dev/provenance") {
+		stmtType := statement.getType()
+		predicateType := statement.getPredicateType()
+		if strings.HasPrefix(stmtType, "https://in-toto.io/Statement") {
+			if strings.HasPrefix(predicateType, "https://slsa.dev/provenance") {
 				return processor.DocumentITE6SLSA
-			} else if strings.HasPrefix(statement.PredicateType, "https://crev.dev/in-toto-scheme") {
+			} else if strings.HasPrefix(predicateType, "https://crev.dev/in-toto-scheme") {
 				return processor.DocumentITE6Generic
-			} else if strings.HasPrefix(statement.PredicateType, "https://in-toto.io/attestation/certify/v0.1") {
+			} else if strings.HasPrefix(predicateType, "https://in-toto.io/attestation/certify/v0.1") {
 				return processor.DocumentITE6Generic
-			} else if strings.HasPrefix(statement.PredicateType, "https://in-toto.io/attestation/vulns/v0.1") ||
-				strings.HasPrefix(statement.PredicateType, "https://in-toto.io/attestation/vulns/v0.2") {
+			} else if strings.HasPrefix(predicateType, "https://in-toto.io/attestation/vulns/v0.1") ||
+				strings.HasPrefix(predicateType, "https://in-toto.io/attestation/vulns/v0.2") {
 				return processor.DocumentITE6Vul
-			} else if strings.HasPrefix(statement.PredicateType, "https://in-toto.io/attestation/clearlydefined/v0.1") {
-				return processor.DocumentITE6ClearlyDefined
-			}
-			return processor.DocumentITE6Generic
-		}
-	}
-	var attV1Statement attestationv1.Statement
-	if json.Unmarshal(blob, &attV1Statement) == nil && format == processor.FormatJSON {
-		if strings.HasPrefix(attV1Statement.Type, "https://in-toto.io/Statement") {
-			if strings.HasPrefix(attV1Statement.PredicateType, "https://in-toto.io/attestation/vulns/v0.1") ||
-				strings.HasPrefix(attV1Statement.PredicateType, "https://in-toto.io/attestation/vulns/v0.2") {
-				return processor.DocumentITE6Vul
-			} else if strings.HasPrefix(attV1Statement.PredicateType, "https://in-toto.io/attestation/clearlydefined/v0.1") {
+			} else if strings.HasPrefix(predicateType, "https://in-toto.io/attestation/clearlydefined/v0.1") {
 				return processor.DocumentITE6ClearlyDefined
 			}
 			return processor.DocumentITE6Generic
--- a/pkg/handler/processor/guesser/type_ite6.go
+++ b/pkg/handler/processor/guesser/type_ite6.go
@@ -20,9 +20,8 @@ import (
 
 	jsoniter "github.com/json-iterator/go"
 
+	"github.com/guacsec/guac/internal/intoto"
 	"github.com/guacsec/guac/pkg/handler/processor"
-	attestationv1 "github.com/in-toto/attestation/go/v1"
-	"github.com/in-toto/in-toto-golang/in_toto"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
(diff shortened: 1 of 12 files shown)


Abhishek9639:
Hi @funnelfiasco or @jeffmendoza,
This PR removes the remaining in-toto-golang references from the guesser, processor, and parser packages as mentioned in #2036.

The main approach was to define lightweight local structs instead of pulling in the deprecated library the guesser now handles both v0.1 and v1 in-toto statement formats through a single struct. All existing tests pass without changes to test logic.

Let me know if anything needs to be adjusted. 
Thanks

pkg/handler/processor/guesser/type_ite6.go
```diff
@@ -21,40 +21,53 @@ import (
...
 	"github.com/guacsec/guac/pkg/handler/processor"
-	attestationv1 "github.com/in-toto/attestation/go/v1"
-	"github.com/in-toto/in-toto-golang/in_toto"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
+// ite6Statement supports unmarshalling both v0.1 ("_type", "predicateType")
+// and v1 ("type", "predicate_type") in-toto statement formats.
+type ite6Statement struct {
```
mlieberman85: Same struct name `ite6Statement` is defined here (4 fields, both v0.1 and v1) and in `pkg/handler/processor/ite6/ite6.go:29` (2 fields, only v0.1). Different packages so the compiler is fine, but it's a maintenance trap. Behaviors will diverge silently if someone updates one and not the other. Share via an internal package, or at minimum document why they're intentionally different and what happens to a v1 doc in the second path.

pkg/handler/processor/guesser/type_ite6.go
```diff
@@ -21,40 +21,53 @@ import (
...
+	TypeV01          string `json:"_type"`
+	PredicateTypeV01 string `json:"predicateType"`
+	TypeV1           string `json:"type"`
+	PredicateTypeV1  string `json:"predicate_type"`
+}
+
+// getType returns the statement type from whichever format was used.
+func (s *ite6Statement) getType() string {
```
mlieberman85: Preference order is v0.1 over v1 on ambiguity. Why? In-toto attestation v1 is the current standard, v0.1 is older. A document with both `_type` and `type` populated is malformed and silently treating it as v0.1 hides that. Either reject mixed-format docs explicitly, or flip the preference to v1.

pkg/handler/processor/ite6/ite6.go
```diff
@@ -21,11 +21,16 @@ import (
 	jsoniter "github.com/json-iterator/go"
 
 	"github.com/guacsec/guac/pkg/handler/processor"
-	"github.com/in-toto/in-toto-golang/in_toto"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
+// ite6Statement is used for unmarshalling in-toto statement headers.
+type ite6Statement struct {
```
mlieberman85: Only handles v0.1 (`_type`/`predicateType`), unlike the guesser version of the same struct in `processor/guesser/type_ite6.go`. A v1 statement reaching `parseStatement` will unmarshal with both fields empty and the rest of the pipeline runs on the empty result with no error. Either match the guesser's merged shape, or add an explicit error when neither set of fields populates.

pkg/ingestor/parser/slsa/parser_slsa.go
```diff
@@ -45,7 +42,80 @@ import (
 // - a pkg or source depending on what is represented by the name/URI
 // - An IsOccurence input spec which will generate a predicate for each occurence
 
-const PredicateSLSAProvenancev1 = "https://slsa.dev/provenance/v1"
+// DigestSet is a set of digests keyed by algorithm name (e.g. "sha256").
+type DigestSet = map[string]string
```
mlieberman85: These local type definitions (`DigestSet`, `ProvenanceMaterial`, `ProvenancePredicateV01/V02`, etc.) are local copies of types you just removed from upstream. They mix with parser logic in this file. I'd put them in their own file (e.g. `types_legacy.go`) with a doc comment explaining why they exist locally, pointing at the deprecated `in-toto-golang` repo for the original definitions, and a note: "delete when SLSA v0.1/v0.2 support is dropped."

pkg/ingestor/parser/slsa/parser_slsa.go
```diff
@@ -45,7 +42,80 @@ import (
...
+type ProvenanceBuilder struct {
+	ID string `json:"id"`
+}
+
+const (
+	PredicateSLSAProvenanceV01 = "https://slsa.dev/provenance/v0.1"
+	PredicateSLSAProvenanceV02 = "https://slsa.dev/provenance/v0.2"
+
+	// PredicateSLSAProvenancev1 is the predicate type for SLSAv1.0 provenance.
+	PredicateSLSAProvenancev1 = "https://slsa.dev/provenance/v1"
```
mlieberman85: Naming inconsistency: new constants use uppercase `V` (`V01`, `V02`), the pre-existing `PredicateSLSAProvenancev1` uses lowercase `v`. Pick one and apply it across all three.

pkg/ingestor/parser/slsa/parser_slsa.go
```diff
@@ -45,7 +42,80 @@ import (
...
+	Arguments         interface{} `json:"arguments,omitempty"`
+	Environment       interface{} `json:"environment,omitempty"`
+}
+
+// ProvenanceMetadataV01 holds build metadata for SLSA v0.1 provenance.
+type ProvenanceMetadataV01 struct {
+	BuildInvocationID string     `json:"buildInvocationId,omitempty"`
+	BuildStartedOn    *time.Time `json:"buildStartedOn,omitempty"`
+	BuildFinishedOn   *time.Time `json:"buildFinishedOn,omitempty"`
+	Completeness      struct {
```
mlieberman85: Anonymous inline struct for `Completeness` in both `ProvenanceMetadataV01` and `ProvenanceMetadataV02`. Can't reference the type elsewhere, hard to inspect, and the upstream `slsa01.ProvenanceMetadata` had a named type for this. Promote to named types: `CompletenessV01` and `CompletenessV02` (the field set differs between v0.1 and v0.2).

mlieberman85 (commented):
Real supply chain win to drop the deprecated dependency, thanks for picking this up. Tests still pass, go.mod cleanup is clean. A few design concerns I'd like to see addressed before merge, none of them blockers individually but together they suggest the cleanup pass would land in a better place.

**Structural:**

1. `ite6Statement` is now defined in two files with different shapes: 4 fields in `processor/guesser/type_ite6.go` (handles both v0.1 and v1), 2 fields in `processor/ite6/ite6.go` (only v0.1). Different packages so no compile error, but a maintenance trap. If a v1 statement reaches `processor/ite6/ite6.go::parseStatement`, both fields unmarshal empty and the rest of the pipeline runs on garbage. Either share the struct via an internal package, or document why they intentionally diverge and what a v1 doc does in the second path.

2. The merged v0.1+v1 struct in the guesser loses the explicit two-pass clarity of the original. The old code had two separate `Unmarshal` calls into properly-typed statements with explicit conditionals. The new code unmarshals everything into one struct and `getType()`/`getPredicateType()` prefer v0.1 over v1 on ambiguity. Why does the older spec win? A doc populating both `_type` and `type` is malformed, the silent preference is a footgun. Either reject mixed formats, or pick a defensible rule (in-toto v1 is the current standard, defaulting to v0.1 feels backwards).

3. The local type definitions (`ProvenancePredicateV01/V02`, `ProvenanceMaterial`, etc.) are dropped into `parser_slsa.go` mixed with parsing logic. They're really local copies of types you just removed from upstream. Belongs in its own file (e.g. `types_legacy.go`) with a doc comment explaining why they exist locally and a marker for when SLSA v0.1/v0.2 support could be dropped entirely.

**Smaller:**

4. `Completeness` is an inline anonymous struct in both `ProvenanceMetadataV01` and `ProvenanceMetadataV02`. Works, but you can't reference these types elsewhere and they're hard to inspect. Named types would be cleaner.

5. New constants `PredicateSLSAProvenanceV01`/`V02` use uppercase V, the pre-existing `PredicateSLSAProvenancev1` uses lowercase v. Pick one convention.

6. Test coverage: I see updates for `parser_slsa_test.go` and `sigstore_verifier_test.go`, none for `type_ite6.go`. The merged-struct guesser logic is the most behavior-changing part of this PR and should have explicit tests for both v0.1 and v1 inputs, plus the ambiguous-fields case.

7. `sigstore_verifier_test.go` now constructs the statement with hand-rolled local types instead of canonical types. Test still passes, but it's now testing self-defined shapes rather than the format the verifier actually consumes in production. Worth keeping the test types as close to upstream as possible.

None of this is hard to address. Happy to look again.

pkg/handler/processor/guesser/type_ite6.go
Abhishek9639: Done!

pkg/handler/processor/guesser/type_ite6.go
Abhishek9639: Done!

pkg/handler/processor/ite6/ite6.go
Abhishek9639: Done!

pkg/ingestor/parser/slsa/parser_slsa.go
Abhishek9639: Done!

pkg/ingestor/parser/slsa/parser_slsa.go
Abhishek9639: Done!

pkg/ingestor/parser/slsa/parser_slsa.go
Abhishek9639: Done!

Abhishek9639:
Hi @mlieberman85, 
Addressed all the feedback unified the ite6Statement struct across both files, flipped the v1 preference in getType()/getPredicateType() and added explicit rejection for ambiguous docs, moved the legacy SLSA types into a separate types_legacy.go with a delete marker, promoted Completeness to named types, fixed the V1 naming inconsistency, and added v1 + ambiguous-fields test cases. 
Let me know if anything else needs tweaking.
Thanks

Abhishek9639:
Hi @mlieberman85, 
Pushed a fix for the CI failures there was a missing "time" import in parser_slsa.go causing a compile error, plus two copy-paste bugs where FinishedOn was using StartedOn's values in both fillSLSA1 and fillSLSA02. All fixed now, CI should pass. 
Thanks

mihaimaruseac (approved):

pkg/ingestor/parser/slsa/types_legacy.go
```diff
@@ -0,0 +1,101 @@
+//
+// Copyright 2022 The GUAC Authors.
```
gaganhr94: Please change this to 2026

pkg/handler/processor/guesser/type_ite6.go
```diff
@@ -21,40 +21,66 @@ import (
...
+		return s.TypeV1
+	}
+	return s.TypeV01
+}
+
+// getPredicateType returns the predicate type from whichever format was used.
+// in-toto v1 is the current standard; v0.1 is supported for backwards
+// compatibility. A document that populates both predicateType and predicate_type
+// is malformed.
+func (s *ite6Statement) getPredicateType() string {
```
gaganhr94: Ambiguity handling here is asymmetric with `getType()`.

When a document sets both `predicateType` and `predicate_type`, this returns `""` to signal "malformed, reject". But the caller only treats an empty *statement* type as fatal: if `_type`/`type` is unambiguous and valid, `strings.HasPrefix(stmtType, "https://in-toto.io/Statement")` still passes, every `predicateType` prefix check fails against `""`, and the function falls through to the final `return processor.DocumentITE6Generic`.

So an ambiguous SLSA or vulns attestation is silently misclassified as generic rather than rejected as `DocumentUnknown` — the opposite of what the ambiguous-`getType()` case (which the new test asserts on) does.

Non-blocking: this only triggers on a malformed document that populates both spellings, and the failure mode is misrouting to the generic parser, not corruption. But it's worth making the two paths consistent, e.g. distinguishing "absent" from "ambiguous" (a second return value or a sentinel) and returning `DocumentUnknown` on ambiguity.

pkg/handler/processor/ite6/ite6.go
```diff
@@ -21,11 +21,21 @@ import (
...
-	"github.com/in-toto/in-toto-golang/in_toto"
 )
 
 var json = jsoniter.ConfigCompatibleWithStandardLibrary
 
+// ite6Statement is used for unmarshalling in-toto statement headers.
+// It handles both v0.1 (_type/predicateType) and v1 (type/predicate_type)
+// in-toto statement formats. Exactly one set of fields should be populated;
+// if neither is set after unmarshalling, the document is rejected with an error.
+type ite6Statement struct {
```
gaganhr94: This `ite6Statement` is an exact duplicate of the struct added in `pkg/handler/processor/guesser/type_ite6.go` — same four fields, same JSON tags — but without the `getType`/`getPredicateType` accessors.

Two independent copies of the same wire-format header type will drift: a future change to how v0.1/v1 fields are reconciled has to be made in both places, and the version that lacks the accessors is easy to miss.

Non-blocking (no behavior impact), but worth defining the shared header type and its accessors once and importing it in both packages.

gaganhr94 (commented):
Two non-blocking notes on the new in-toto header handling. Neither affects the correctness of well-formed documents — the `BuildStartedOn`/`FinishedOn` fixes in this diff look right.

gaganhr94 (changes requested):

pkg/ingestor/parser/slsa/types_legacy.go
Abhishek9639: Done!

pkg/handler/processor/guesser/type_ite6.go
Abhishek9639: Done!

pkg/handler/processor/ite6/ite6.go
Abhishek9639: Done!

gaganhr94:
@Abhishek9639 can you rebase your branch on main and resolve all the conflicts

Abhishek9639:
Hi @gaganhr94, 
I’ve addressed all the requested changes and carefully rechecked the implementation. Could you please review the PR again when you have a chance? 
Thank you!

gaganhr94:
@Abhishek9639 can you please fix the lint and also look into why the static analysis pipeline is failing

Abhishek9639:
Hi @gaganhr94, 
I've fixed the lint errors and resolved the merge conflicts. Could you please take a look and confirm if everything looks good now?
Thanks!

gaganhr94 (approved):

Issue #1, opened 2022-08-03
Interested in Dev/Contributing to GUAC?


lumjjb:
Welcome! This thread is on expressing interest in contributing to GUAC! We are glad to welcome our fellow open source contributors! As the project is starting up, we will be creating issues that folks can pick up and work on. In the meantime, as the code base is forming up, we'd like to engage directly with our contributors!

BTW we now have a slack channel: https://openssf.slack.com/archives/C03U677QD46

If you are interested in contributing, it would be very helpful to provide the following details (copy and paste into your comment):
```
1. I am interested in contributing to:
- [ ] Development
- [ ] Documentation
- [ ] Issue triage and community
- [ ] Technical advisory (review [governance document](https://github.com/artifact-ff/artifact-ff/blob/main/GOVERNANCE.md#technical-advisory-members))

2. I am here because:
- [ ] Personal interest
- [ ] My company/orgs i work with are interested in this

3. What is your associated company/org if you're contributing in their capacity? _________

4. Depending on how things go, I may be interested in becoming a maintainer of the project
- [ ] Yes

5. (optional) I have expertise in:
- [ ] Neo4j
- [ ] Cypher
- [ ] GraphQL
- [ ] Intoto
- [ ] SPDX
- [ ] CycloneDX
- [ ] Others (fill in):
```

cpendery:
1. I am interested in contributing to:
- [x] Development
- [ ] Documentation
- [ ] Issue triage and community
- [ ] Technical advisory (review [governance document](/GOVERNANCE.md#technical-advisory-members))

2. I am here because:
- [x] Personal interest
- [ ] My company/orgs i work with are interested in this

3. What is your associated company/org if you're contributing in their capacity? 

4. Depending on how things go, I may be interested in becoming a maintainer of the project
- [x] Yes

5. (optional) I have expertise in:
- [ ] Neo4j
- [ ] Cypher
- [ ] GraphQL
- [ ] Intoto
- [ ] SPDX
- [x] CycloneDX
- [x] Others (fill in): Grype, Syft, Trivy, OSV data formats, Golang

Note: my company may be interested in the project and me contributing in their capacity, so I'll update this note if they approve that work

shafeeshafee:
1. I am interested in contributing to:
- [x] Development
- [x] Documentation
- [x] Issue triage and community
[...]

PR #3018, opened 2026-04-23
feat: add VEX Hub certifier

### LINKED ISSUE #2296: [feature] Add VEX Hub support

This was mentioned by Lukas Hoehl in [Slack](https://openssf.slack.com/archives/C03U677QD46/p1731679983846979). VEX statements aren't always easy to find. [VEX Hub](https://github.com/aquasecurity/vexhub) automatically collects VEX statements and makes them easily discoverable.

A certifier that queries VEX Hub for VEX statements affecting pURLs found in GUAC data.

Directly incorporating [vexhub-crawler](https://github.com/aquasecurity/vexhub-crawler) or similar into GUAC to fetch the VEX statements directly. 

**Additional context**
This was the subject of a conversation at KubeCon today. It should be relatively simple to create a certifier for this, it "just" needs someone with the time to write the code.

Abhishek9639:
# Description of the PR
This PR adds a new certifier to support VEX Hub integration, as discussed in the linked issue. 

The `vexhub` certifier queries repositories conforming to the VEX Repo Spec. It fetches the manifest (`vex-repository.json`), downloads the tar.gz archive, and uses `index.json` to map package PURLs to their corresponding OpenVEX documents. The manifest URL is configurable via the `--vexhub-manifest-url` flag (defaults to the Aqua Security VEX Hub). 

Both `guacone` and `guaccollect` CLI commands have been added.

Fixes #2296

# PR Checklist

### DIFF
--- a/pkg/certifier/vexhub/vexhub.go
+++ b/pkg/certifier/vexhub/vexhub.go
@@ -1,0 +1,359 @@
+//
+// Copyright 2024 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package vexhub
+
+import (
+	"archive/tar"
+	"compress/gzip"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/guacsec/guac/pkg/certifier"
+	"github.com/guacsec/guac/pkg/certifier/components/root_package"
+	"github.com/guacsec/guac/pkg/clients"
+	"github.com/guacsec/guac/pkg/events"
+	"github.com/guacsec/guac/pkg/handler/processor"
+	"github.com/guacsec/guac/pkg/logging"
+	"github.com/guacsec/guac/pkg/version"
+
+	jsoniter "github.com/json-iterator/go"
+	"golang.org/x/time/rate"
+)
+
+var json = jsoniter.ConfigCompatibleWithStandardLibrary
+
+const (
+	DefaultManifestURL = "https://raw.githubusercontent.com/aquasecurity/vexhub/main/vex-repository.json"
+	VEXHubCollector    = "vexhub_certifier"
+	rateLimit          = 100
+)
+
+var rateLimitInterval = time.Second
+
+var ErrComponentTypeMismatch = errors.New("rootComponent type is not []*root_package.PackageNode")
+
+// Manifest represents a vex-repository.json file.
+type Manifest struct {
+	Name     string            `json:"name"`
+	Versions []ManifestVersion `json:"versions"`
+}
+
+// ManifestVersion holds the spec version and locations of a VEX repo.
+type ManifestVersion struct {
+	SpecVersion    string             `json:"spec_version"`
+	Locations      []ManifestLocation `json:"locations"`
+	UpdateInterval string             `json:"update_interval"`
+}
+
+// ManifestLocation holds the URL for a VEX repo archive.
+type ManifestLocation struct {
+	URL string `json:"url"`
+}
+
+// Index represents the index.json file inside the VEX repo archive.
+type Index struct {
+	UpdatedAt string         `json:"updated_at"`
+	Packages  []IndexPackage `json:"packages"`
+}
+
+// IndexPackage maps a PURL to its VEX document location.
+type IndexPackage struct {
+	ID       string `json:"id"`
+	Location string `json:"location"`
+	Format   string `json:"format,omitempty"`
+}
+
+// vexHubCertifier queries VEX repositories for VEX statements.
+type vexHubCertifier struct {
+	httpClient  *http.Client
+	manifestURL string
+}
+
+// NewVEXHubCertifier creates a new VEX Hub certifier.
+func NewVEXHubCertifier(manifestURL string) certifier.Certifier {
+	limiter := rate.NewLimiter(rate.Every(rateLimitInterval), rateLimit)
+	transport := clients.NewRateLimitedTransport(version.UATransport, limiter)
+	client := &http.Client{Transport: transport}
+	if manifestURL == "" {
+		manifestURL = DefaultManifestURL
+	}
+	return &vexHubCertifier{
+		httpClient:  client,
+		manifestURL: manifestURL,
+	}
+}
+
+// CertifyComponent fetches VEX documents from the VEX Hub for the given packages.
+func (v *vexHubCertifier) CertifyComponent(ctx context.Context, rootComponent interface{}, docChannel chan<- *processor.Document) error {
+	logger := logging.FromContext(ctx)
+
+	packageNodes, ok := rootComponent.([]*root_package.PackageNode)
+	if !ok {
+		return ErrComponentTypeMismatch
+	}
+
+	var purls []string
+	for _, node := range packageNodes {
+		purls = append(purls, node.Purl)
+	}
+
+	if len(purls) == 0 {
+		return nil
+	}
+
+	// Fetch the manifest to get the archive URL.
+	manifest, err := fetchManifest(ctx, v.httpClient, v.manifestURL)
+	if err != nil {
+		return fmt.Errorf("failed to fetch VEX Hub manifest: %w", err)
+	}
+
+	archiveURL, subdir := getArchiveURL(manifest)
+	if archiveURL == "" {
+		logger.Infof("no archive location found in VEX Hub manifest")
+		return nil
+	}
+
+	// Download and extract the archive, building a PURL→VEX doc map.
+	vexDocs, err := downloadAndIndex(ctx, v.httpClient, archiveURL, subdir)
+	if err != nil {
+		return fmt.Errorf("failed to download VEX Hub archive: %w", err)
+	}
+
+	logger.Infof("VEX Hub: indexed %d packages from archive", len(vexDocs))
+
+	// Look up each PURL and emit matching VEX documents.
+	if _, err := emitVEXDocuments(purls, vexDocs, docChannel); err != nil {
+		return fmt.Errorf("failed to emit VEX documents: %w", err)
+	}
+
+	return nil
+}
+
+// fetchManifest downloads and parses the vex-repository.json manifest.
+func fetchManifest(ctx context.Context, client *http.Client, url string) (*Manifest, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating request: %w", err)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching manifest: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("manifest fetch returned status %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("reading manifest body: %w", err)
+	}
+
+	var manifest Manifest
+	if err := json.Unmarshal(body, &manifest); err != nil {
+		return nil, fmt.Errorf("parsing manifest: %w", err)
+	}
+	return &manifest, nil
+}
+
+// getArchiveURL extracts the archive URL and optional subdirectory from the manifest.
+// The subdirectory is specified after "//" in the URL (per VEX Repo Spec).
+// Example: "https://example.com/archive.tar.gz//vexhub-main" → URL + subdir "vexhub-main"
+func getArchiveURL(manifest *Manifest) (archiveURL, subdir string) {
+	if len(manifest.Versions) == 0 {
+		return "", ""
+	}
+	// Use the first version with a location.
+	for _, v := range manifest.Versions {
+		if len(v.Locations) > 0 {
+			rawURL := v.Locations[0].URL
+			// Look for "//" that is NOT part of the scheme (e.g., "https://").
+			// The subdirectory separator always appears after the host/path portion.
+			schemeEnd := strings.Index(rawURL, "://")
+			searchStart := 0
+			if schemeEnd >= 0 {
+				searchStart = schemeEnd + 3
+			}
+			rest := rawURL[searchStart:]
+			if idx := strings.Index(rest, "//"); idx >= 0 {
+				archiveURL = rawURL[:searchStart+idx]
+				subdir = rest[idx+2:]
+			} else {
+				archiveURL = rawURL
+			}
+			return archiveURL, subdir
+		}
+	}
+	return "", ""
+}
+
+// downloadAndIndex downloads a tar.gz archive, extracts it, parses index.json,
+// and returns a map of PURL→VEX document bytes.
+func downloadAndIndex(ctx context.Context, client *http.Client, archiveURL, subdir string) (map[string][]byte, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, archiveURL, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating archive request: %w", err)
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("downloading archive: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("archive download returned status %d", resp.StatusCode)
+	}
+
+	gz, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("creating gzip reader: %w", err)
+	}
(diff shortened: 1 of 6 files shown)


Abhishek9639:
Hi @funnelfiasco or @jeffmendoza,
I've opened the PR for the VEX Hub certifier implementation. The CI checks are passing locally. 
Could you please take a look when you have some time? Let me know if any changes are needed.
Thanks

pkg/certifier/vexhub/vexhub.go
```diff
@@ -0,0 +1,359 @@
...
+type vexHubCertifier struct {
+	httpClient  *http.Client
+	manifestURL string
+}
+
+// NewVEXHubCertifier creates a new VEX Hub certifier.
+func NewVEXHubCertifier(manifestURL string) certifier.Certifier {
+	limiter := rate.NewLimiter(rate.Every(rateLimitInterval), rateLimit)
+	transport := clients.NewRateLimitedTransport(version.UATransport, limiter)
+	client := &http.Client{Transport: transport}
```
mlieberman85: No `Timeout` on the client. If the archive endpoint stalls, this certifier goroutine pins indefinitely. Set `client.Timeout` or thread a context with deadline through `fetchManifest` and `downloadAndIndex`.

pkg/certifier/vexhub/vexhub.go
```diff
@@ -0,0 +1,359 @@
...
+	gz, err := gzip.NewReader(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("creating gzip reader: %w", err)
+	}
+	defer func() { _ = gz.Close() }()
+
+	tr := tar.NewReader(gz)
+
+	// First pass: extract all files into memory.
+	files := make(map[string][]byte)
```
mlieberman85: Unbounded `io.ReadAll` per entry, no total-size cap, full archive buffered in memory before anything is parsed. This OOMs as soon as the hub gets non-trivial. Two passes: parse `index.json` first, then only read entries referenced in `index.Packages`. Wrap each read in `io.LimitReader` with a per-entry cap and track cumulative bytes.

pkg/certifier/vexhub/vexhub.go
```diff
@@ -0,0 +1,359 @@
...
+
+// getArchiveURL extracts the archive URL and optional subdirectory from the manifest.
+// The subdirectory is specified after "//" in the URL (per VEX Repo Spec).
+// Example: "https://example.com/archive.tar.gz//vexhub-main" → URL + subdir "vexhub-main"
+func getArchiveURL(manifest *Manifest) (archiveURL, subdir string) {
+	if len(manifest.Versions) == 0 {
+		return "", ""
+	}
+	// Use the first version with a location.
+	for _, v := range manifest.Versions {
```
mlieberman85, line 266: First version with a location wins, `spec_version` is ignored. Forward-incompatible: a spec bump published by Aqua first will be silently downloaded with the wrong parser. Match against a known-compatible spec version, log+skip otherwise.

pkg/certifier/vexhub/vexhub.go
```diff
@@ -0,0 +1,359 @@
...
+		}
+		emitted = append(emitted, doc)
+	}
+
+	return emitted, nil
+}
+
+// stripPurlVersion removes the version, qualifiers, and subpath from a PURL.
+// e.g. "pkg:npm/lodash@4.17.21" → "pkg:npm/lodash"
+func stripPurlVersion(purl string) string {
```
mlieberman85: `stripPurlVersion` is a hand-rolled PURL parser. PURLs are not just `name@version`, qualifier and subpath ordering and namespace casing are all in scope for canonicalization. Two valid PURL strings for the same package will miss each other in the map lookup above.

I ran into this recently while doing PURL conformance testing across SBOM generators: SBOMs describing the same package consistently produced PURL strings that did not compare equal as raw strings. The map lookup pattern here will hit the same problem. Use `packageurl-go`'s `FromString` then re-stringify both sides before keying.

mlieberman85 (changes requested):
Good scope and the tests aren't trivial, the end-to-end with httptest is the right shape. A few things need to land before this is mergeable.

**Blockers (P0):**

1. `downloadAndIndex` reads the whole archive into memory with no size cap. Tar bomb territory and an OOM as soon as the hub grows past a small set. Two-pass it: parse `index.json` first, then only buffer files referenced by the index. Or set explicit per-entry and total-size limits with `io.LimitReader`.
2. `NewVEXHubCertifier` builds an `http.Client` with no `Timeout`. If the archive endpoint hangs, the certifier goroutine pins forever. Combined with #1 this is hang-then-OOM.

**Should fix before merge (P1):**

3. PURL matching in `emitVEXDocuments` is direct string lookup, fallback to a hand-rolled `stripPurlVersion`. Different valid spellings of the same PURL (qualifier order, namespace casing, default vs explicit type, percent-encoding) will silently miss matches. I ran into this recently while working on PURL conformance testing across SBOM generators, where two SBOMs describing the same package routinely produced PURL strings that did not compare equal. Use `packageurl-go` to canonicalize both sides before lookup.
4. `getArchiveURL` walks `manifest.Versions` and grabs the first one with a location, ignoring `spec_version`. If Aqua publishes a spec bump and lists it first, this will happily download a manifest format it can't parse. Match against known-compatible spec versions.

**Nits:**

5. Copyright headers on the five new files say 2024.
6. `stripPurlVersion` should not exist. `packageurl-go` already parses these.
7. Test coverage skips: malformed manifest, missing `index.json`, oversized archive, network errors, multi-version manifest, stripped-version-still-misses.

What I like: configurable manifest URL, rate limiter wired up, `pkg:guac/*` skip, dedupe via `seen`, the `//` subdir hack from the VEX Repo Spec parsed correctly in `getArchiveURL`.

Happy to look again once 1 and 2 are addressed.

pkg/certifier/vexhub/vexhub.go
Abhishek9639: Done!

pkg/certifier/vexhub/vexhub.go
Abhishek9639: Done!

pkg/certifier/vexhub/vexhub.go
Abhishek9639: Done!

pkg/certifier/vexhub/vexhub.go
Abhishek9639: Done!

Abhishek9639:
Hi @mlieberman85, 
I've addressed all the feedback from your review. All P0, P1, and nits have been fixed including the two-pass archive parsing with io.LimitReader, HTTP client timeout, PURL canonicalization via packageurl-go, spec_version filtering, copyright year updates, and the additional test cases. 
Would love your re-review when you get a chance. 
Thanks!

gaganhr94:
Thanks for picking this up, and sorry about the stale-bot closure — the shape here is good and the tests are not trivial, so it's worth finishing.

I went back through the code at the head of the branch (`dca21a8`). The earlier P0/P1 items do look addressed. Below is what I think still needs a fix before this is mergeable. Line references are against `pkg/certifier/vexhub/vexhub.go` at that commit.

### Blocking

**1. The whole archive is re-fetched on every batch.**
`certify.Certify` calls `CertifyComponent` once per batch of packages (`vexHubQuerySize = 1000`), and each call does `fetchManifest` → download the full `.tar.gz` → gunzip → two passes over it. For a graph with 100k packages that's 100 full downloads of the hub per poll cycle, which will not be kind to the hub or to us.

The manifest already parses `update_interval` into `ManifestVersion.UpdateInterval` (L80) and never reads it — that field is exactly the TTL the spec intends. Suggest caching the parsed manifest and the built `vexDocs` map on the `vexHubCertifier` struct, keyed by manifest URL, with that interval as the TTL. Adding `If-Modified-Since`/ETag handling on top would be a nice extra but the cache is the part that matters.

**2. `seen` is allocated per call (L407), so dedup is per-batch only.**
The same VEX document gets re-emitted once for every batch it matches in, which multiplies ingestion work downstream. Hoisting it alongside the cache from (1) handles both.

**3. `emitVEXDocuments` can block forever on shutdown.**
The function takes no `context.Context` (L405) and the send at L453 is a bare `docChannel <- doc`. On SIGINT the ingestion goroutine returns via `ctx.Done()` while this one stays blocked on the send, so `wg.Wait()` in the guacone command never returns. Please pass `ctx` through and `select` on `ctx.Done()` around the send.

While you're in there: the `error` return is unconditionally `nil` and the `[]*processor.Document` return is only used by tests. Worth simplifying the signature.

**4. `IndexPackage.Format` is parsed but never used.**
Every blob is emitted as `processor.DocumentOpenVEX` regardless of what the index declares. The VEX Repository Spec allows other formats, so a CSAF entry would be handed to the OpenVEX parser and fail with a confusing error. Suggest skipping entries whose `format` is set and is not openvex (treating an empty value as openvex for compatibility).

### Non-blocking, but I'd like to see them

**5. Only the first location is tried (L219, `v.Locations[0].URL`).**
`locations` is a list so that mirrors can be used as fallbacks. Right now a 404 on the first URL fails the whole certifier with no attempt at the rest. Iterating until one succeeds is a small change.

**6. `normalizeName` (L290–304) is correct only by coincidence.**
The two branches are mutually exclusive: when `subdir != ""` the tarball's top-level directory is never stripped, and when `subdir == ""` the leading component is stripped blindly. This works against the default Aqua manifest purely because its subdir (`vexhub-main`) happens to be the same string as the archive's top-level directory. Any repo where those differ will silently fail to find `index.json`.

Suggest stripping the leading path component first, then applying the subdir prefix, and adding a test case where a subdir is set *and* the archive has an unrelated top-level directory.

**7. Memory during extraction.**
`io.ReadAll` at L280 buffers up to `maxArchiveBytes` (512 MiB) of compressed data and holds it for both passes. Buffering to a temp file via `os.CreateTemp` and `Seek(0)` between passes keeps the two-pass design with a fixed memory footprint.

### Nits

- `getArchiveURL` takes an anonymous `interface{ Infof(string, ...interface{}) }` (L207). Everywhere else in the certifier packages we take a `context.Context` and use `logging.FromContext`. Please match that.
- The `(P0 fix #2)`, `(P1 fix #1)`, `(P0 fix #1)` markers in the comments (L112, L206, L241, L255, L404) are review-thread bookkeeping and should come out — the comments read fine without them.
- The other certifiers have a page under `docs/`. Adding one for `vexhub`, covering the `--vexhub-manifest-url` flag and what a compatible VEX repository looks like, would help people find this.

Happy to re-review once these are addressed.

gaganhr94 (changes requested):
@Abhishek9639 left a few coments, can you fix them and rebase this PR. Thanks

Abhishek9639:
@gaganhr94 pushed a fix for all the points from your review.

- Manifest and vexDocs are now cached on the certifier struct (keyed by manifest URL), using update_interval as the TTL, with a mutex around it.
- seen moved up to the same struct-level cache, so dedup holds across batches instead of resetting per call.
- emitVEXDocuments now takes a context and the channel send is wrapped in a select against ctx.Done(), so it won't hang on shutdown anymore.
- Entries with a non-openvex format get skipped now (empty format still defaults to openvex).
- getArchiveURL loops through all locations instead of stopping at the first one, and now takes context.Context + logging.FromContext to match the rest of the package.
- normalizeName strips the leading path component first and then applies the subdir prefix, so it doesn't just work by coincidence anymore.
- Extraction now streams to a temp file between the two passes instead of holding the whole archive in memory.
- Added test cases for the malformed manifest, missing index.json, oversized archive, network error, multi-version manifest, and subdir-mismatch scenarios.
- Cleaned up the leftover P0/P1 comments and updated the docs.

Rebased on main. DCO's green now too. Let me know if there's anything else you want changed.

gaganhr94:
Thanks for the quick turnaround @Abhishek9639 — re-reviewed at `c8941de`. The changes are all there, but two of them don't take effect at runtime, so the underlying problems from the last round are still present. Details below, line refs against `pkg/certifier/vexhub/vexhub.go`.

I've also dropped a couple of items from my earlier review that I decided on re-reading were not worth your time — noted at the bottom so you're not left guessing.

### Blocking

**1. The caching and dedup state never survives a batch.**

`certify.generateDocuments` builds a fresh certifier for every component batch:

```go
// pkg/certifier/certify/certify.go:159-163
for _, certifier := range documentCertifier {
    c := certifier()          // new instance, every batch
    go func() { errChan <- c.CertifyComponent(ctx, collectedComponent, docChan) }()
}
```

and the registered factory returns a new struct each call:

```go
vexHubCertifierFunc := func() certifier.Certifier {
    return vexhub.NewVEXHubCertifier(opts.manifestURL)
}
```

`generateDocuments` is invoked once per batch off `compChan` (`certify.go:88`), and `root_package.go:98-112` flushes a batch every `serviceBatchSize` nodes *and* on a ticker — so a real deployment produces many batches. Net effect: `cache`, `seen`, and both mutexes are re-allocated per batch, the cache never hits, and the archive is still re-downloaded exactly as often as before.

Worth noting the framework contract here: every other certifier registers a bare constructor (`eol.NewEOLCertifier`, `clearlydefined.NewClearlyDefinedCertifier`, …), i.e. certifiers are expected to be stateless per invocation. Caching has to live outside the factory — construct one `*vexHubCertifier` and have the closure return that same pointer. That also makes the two mutexes meaningful; right now they guard nothing.

**2. `seen` is never reset across a cache refresh, so updated VEX documents are permanently suppressed.**

This one only bites *after* #1 is fixed, which is why it's easy to miss. The cache TTL is the manifest's `update_interval`; on expiry the archive is re-fetched and re-indexed (`:152-189`), but `seen` still holds every PURL emitted so far (`:503-506`). A revised VEX statement for an already-seen package would never reach the pipeline again for the lifetime of the process — which defeats the point of polling on `update_interval`.

Suggest keying dedup on the document digest rather than the PURL — `events.GetDocRef(docData)` is already computed at `:515` — so changed content re-emits and unchanged content still doesn't. Clearing `seen` whenever a new archive is indexed also works.

These two need to land together: today #1 masks #2 (fresh instance per batch resets `seen`, so docs do re-emit, just with duplicates). Fixing #1 alone converts the duplicate-emission behaviour into silent suppression.

### Should fix

**3. No usable archive location returns success, not an error.** `:162-165` — when `getArchiveURL` finds nothing, this logs at Info and returns `nil`. A completely broken or unreachable hub is then indistinguishable from "nothing to do". Worth a Warn at minimum, and I'd argue an error. Related: the HEAD probe at `:272-282` skips any location not answering 200, and a fair number of object stores and CDNs answer 405/403 to HEAD while serving GET fine — so this path is more reachable than it looks.

**4. `maxEntryBytes` truncates silently.** `:440` — `io.ReadAll(io.LimitReader(tr2, maxEntryBytes))` yields a truncated document rather than an error, which surfaces downstream as a confusing JSON parse failure rather than "this file was too big". The `maxArchiveBytes` check at `:340-346` already gets this right with the read-one-extra-byte-and-compare pattern; same treatment here.

### Nits

- `manifestCacheEntry.manifest` is stored and never read.
- `logger.Infof("VEX Hub: indexed %d packages …")` at `:192` fires once per batch.
- `emitVEXDocuments` holds `seenMu` across the blocking channel send (`:519-525`). Harmless today, less so once the certifier instance is shared.

### Withdrawn from my earlier review

For completeness, two things I raised in an earlier draft and then talked myself out of — no action needed:

- **Archive-URL SSRF.** Following URLs out of a manifest fetched from an operator-configured endpoint is the VEX Repo Spec's design, and GUAC's collectors already fetch operator-pointed remote content throughout. No boundary is crossed that the feature doesn't cross by existing.
- **Decompression cost of unreferenced tar entries.** Requires a hostile hub, which is already game over for a certifier whose job is ingesting that hub. Memory stays bounded; only CPU is at risk.

The `normalizeName` helper at `:350-354` unconditionally strips the leading path component, which would misbehave for an archive with no top-level wrapper directory — but I couldn't point to a hub that ships one, so treat that as an observation rather than a request.

Shape of the feature still looks good, and the two-pass extraction reads well. Happy to re-review once 1 and 2 are sorted.

gaganhr94:
Please fix the lint as well @Abhishek9639

gaganhr94:
@Abhishek9639 can you please resolve the comments and fix the lint in this PR

Abhishek9639:
Hi @gaganhr94,
Rebased on latest `main` the `certifier.go` conflict is resolved (kept both `CertifierDatadogMalware` and `CertifierVEXHub`). Both blocking items are fixed:

1. **Caching/dedup not surviving a batch** - you were right that the fix had to be outside the factory: the package changes alone did nothing. Both cobra commands now build one certifier and the closure returns that same pointer. Also collapsed the cache map into a single entry since it was keyed on a URL that's already a struct field that removes the unused `manifest` field you spotted.

2. **seen suppressing updated docs** - dedup now keys on `events.GetDocRef(docData)` as you suggested. Added a test for it, and confirmed it actually fails against the old PURL-keyed behaviour rather than just passing by luck.

Also done: `getArchiveURL` returns an error instead of `nil` when nothing is usable, and the HEAD probe no longer requires a 200 (good catch on 405/403); `maxEntryBytes` now errors instead of truncating; the three nits. I fixed `normalizeName` too the prefix is discovered from where `index.json` actually lands, so wrapper-less archives work.

On the lint: it wasn't golangci-lint (that was clean) `make format` was reformatting `docs/vexhub_certifier.md` and `generated_up_to_date` then failed on the dirty tree. Ran prettier; it's a no-op now.

The earlier E2E failure was ClearlyDefined returning 502s during ingestion, so the expected-results wait timed out — unrelated to this certifier.

Could you approve the workflow run when you get a chance? Both are sitting in "awaiting approval". 
Thanks for the detailed reviews.

gaganhr94:
@Abhishek9639 can you please fix the pipeline failures (lint)

Abhishek9639:
Hi @gaganhr94,
Please review it

gaganhr94 (approved):

gaganhr94:
LGTM ! Thanks for the contribution @Abhishek9639

mihaimaruseac (approved):

PR #2202, opened 2024-10-18
Fixed Issue: #2195 500 error when sending empty JSON payload to Graph...

000xs:
Fixed Issue: #2195 500 error when sending empty JSON payload to Graph…
# Description of the PR
This pull request addresses an issue where the GraphQL server returned a 500 internal server error when an empty or invalid JSON payload was sent to the /query endpoint.

### Changes:

- Error handling: Updated the server to return a 400 bad request instead of 500 for invalid JSON input.
- Error message: Improved the error message to provide more context about the issue.

# PR Checklist

- [x] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [x] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If GraphQL schema is changed, GraphQL client updates/additions have been made
- [ ] If OpenAPI spec is changed, `make generate` has been run
- [ ] If ent schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [x] All CI checks are passing (tests and formatting)
- [x] All dependent PRs have already been merged

PR #3067, opened 2026-06-05
fix(metrics): return 400 not 500 for empty or malformed GraphQL request body

### LINKED ISSUE #2195: [bug] GraphQL server returns 500 when sending empty json payload
**Describe the bug**

When empty or broken JSON is send to the /query endpoint, a 500 internal error is returned, which is mis-leading and should be a 400 bad request instead.

**To Reproduce**

1. Start the graphql server
2. fire a `post` request to the `/query` endpoint, wih a payload of 0 bytes

**Actual behaviour**

Input (raw) sent:
```text
POST /query HTTP/1.1
user-agent: guac-visualizer
accept: application/json
accept-encoding: gzip, deflate, br
Host: localhost:8080
Connection: keep-alive
Content-Length: 0
```

The server returns status 500 `unexpected end of JSON input`
Raw reponse from the server:
```text
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Sun, 13 Oct 2024 18:06:52 GMT
Content-Length: 29

unexpected end of JSON input
```

**Expected behaviour**
Since this is related to user input data, a status code 400 "bad request" should be used.
Also the error message could be a bit more proactive: e.g. `fmt.Sprintf("provided input json couldn't be parsed. likely it was empty or was wrongly structured; error message: %v", err.String()`).
When I developed a reverse proxy for the visualizer, I was completely puzzled and mis-lead because of the 500 error and thought of a bigger issue inside GraphQL, looked in to the error logs, but couldn't find. spend too much time thinking, until realizing my error was "just" in the empty json payload.

**Screenshots**
n/a

**GUAC version**
build from repository, as of Oct. 14th, 2024

**Additional context**
n/a

arpitjain099:
## Summary

Fixes #2195.

Posting an empty or malformed JSON body to the `/query` GraphQL endpoint returned `500 Internal Server Error` ("unexpected end of JSON input") instead of `400 Bad Request`. As the reporter noted, a 500 is misleading for what is purely a client input problem and sends people hunting for a server-side fault that does not exist.

## Root cause

The 500 did not come from the GraphQL server itself. The current gqlgen POST transport already answers a bad request body with 400. The 500 came from the Prometheus metrics middleware `MeasureGraphQLResponseDuration` in `pkg/metrics/prometheus.go`, which wraps the GraphQL handler (wired in `cmd/guacgql/cmd/server.go`). That middleware read the body, ran its own `json.Unmarshal` to pull out `operationName` for a metric label, and on any unmarshal error short-circuited the request with `http.StatusInternalServerError` before the GraphQL handler ever ran. An empty body is not valid JSON, so it tripped this path.

This matches the review feedback on the earlier (now stale-closed) attempt in #2202: a maintainer noted the change there was in the wrong place and that the fix belongs at the GraphQL server side, not bolted onto the metrics path with a second status decision.

## Fix

The metrics middleware should observe, not adjudicate. It now best-effort parses the operation name and, regardless of whether the body is valid JSON, forwards the request to the downstream GraphQL handler. The handler then returns the correct status: 400 for an empty or malformed body, 200 for a valid query. The unmarshal error no longer produces a 500. When the body does not parse, the `operationName` metric label is simply empty, which is the natural outcome for a request that has no parseable operation.

This also removes a redundant body copy that the old code made only to feed that unmarshal.

## Behavior before / after

Reproduced with a test that posts to the handler.

Before:
- empty body: 500 ("unexpected end of JSON input")
- malformed JSON: 500
- valid query: 200

After:
- empty body: 400
- malformed JSON: 400
- valid query: 200

## Tests

- `pkg/assembler/server/server_badrequest_test.go`: end-to-end test that runs the real GraphQL server behind the metrics middleware and asserts 400 for empty and malformed bodies and 200 for a valid query. This is the direct regression guard for #2195.
- `pkg/metrics/measure_graphql_test.go`: unit test asserting the middleware no longer short-circuits a bad body and forwards the original bytes downstream unchanged.

`go build ./...`, `go vet`, and `go test ./pkg/metrics/... ./pkg/assembler/server/...` pass.

## Note on the prior PR

#2202 targeted this same issue but was auto-closed by the stale bot after 90 days of inactivity (it was never rejected on the merits; it had an unrelated merge-conflict artifact and a maintainer question about placement). This PR takes the placement the reviewer pointed to and adds the regression tests that were missing there.

### DIFF
--- a/pkg/assembler/server/server_badrequest_test.go
+++ b/pkg/assembler/server/server_badrequest_test.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2024 The GUAC Authors.
+// Copyright 2026 The GUAC Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/assembler/server/server_badrequest_test.go
+++ b/pkg/assembler/server/server_badrequest_test.go
@@ -1,0 +1,85 @@
+//
+// Copyright 2024 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package server
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	_ "github.com/guacsec/guac/pkg/assembler/backends/keyvalue"
+
+	"github.com/guacsec/guac/internal/testing/stablememmap"
+	"github.com/guacsec/guac/pkg/assembler/backends"
+	"github.com/guacsec/guac/pkg/metrics"
+)
+
+// TestQueryEndpointBadRequest covers issue #2195: posting an empty or malformed
+// JSON body to the /query endpoint must return 400 Bad Request, not 500
+// Internal Server Error. The handler is exercised together with the Prometheus
+// metrics middleware because that middleware (not the GraphQL server) was the
+// source of the original 500 response.
+func TestQueryEndpointBadRequest(t *testing.T) {
+	ctx := metrics.WithMetrics(context.Background(), "guac_query_test")
+
+	store := stablememmap.GetStore()
+	backend, err := backends.Get("keyvalue", ctx, store)
+	if err != nil {
+		t.Fatalf("Error getting backend: %v", err)
+	}
+
+	srv := GetGraphqlServer(ctx, backend)
+	collector := metrics.FromContext(ctx, "guac_query_test")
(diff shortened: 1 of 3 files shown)


pkg/assembler/server/server_badrequest_test.go
```diff
@@ -0,0 +1,85 @@
+//
+// Copyright 2024 The GUAC Authors.
```
gaganhr94: Please change this to 2026

pkg/metrics/measure_graphql_test.go
```diff
@@ -0,0 +1,77 @@
+//
+// Copyright 2024 The GUAC Authors.
```
gaganhr94: Change to 2026  here as well

pkg/assembler/server/server_badrequest_test.go
arpitjain099: Done, changed to 2026.

pkg/metrics/measure_graphql_test.go
arpitjain099: Done here too, both new files now say 2026.

mihaimaruseac (approved):

gaganhr94 (approved):

gaganhr94:
Thanks for the contribution @arpitjain099

PR #3076, opened 2026-06-20
fix(processor): bound decompressed document size to prevent decompression bombs

### LINKED ISSUE #3075: [bug] Document decompression has no size limit (decompression bomb)
**Describe the bug**

Document decompression has no size limit. `decompressDocument` in `pkg/handler/processor/process` reads the bzip2/zstd stream with `io.ReadAll` and no cap, so a small compressed document can expand to an arbitrary size and exhaust the processor's memory. The encoding is auto-detected from the blob's magic bytes by the guesser, so any ingested artifact reaches this path regardless of filename.

**To Reproduce**

1. Create a small zstd blob that decompresses to a large payload (e.g. 500 MiB of zeros compresses to ~56 KB).
2. Ingest it, or call `decodeDocument` directly with `Encoding: processor.EncodingZstd`.
3. Observe that the full uncompressed payload is buffered into `doc.Blob` with no limit.

As a unit test in the `process` package (commit `362e6da`):

```go
func TestDecompressionBomb(t *testing.T) {
    var buf bytes.Buffer
    enc, _ := zstd.NewWriter(&buf)
    zeros := make([]byte, 1<<20)
    for i := 0; i < 500; i++ { // 500 MiB of zeros
        enc.Write(zeros)
    }
    enc.Close()
    doc := &processor.Document{Blob: buf.Bytes(), Encoding: processor.EncodingZstd}
    if err := decodeDocument(context.Background(), doc); err != nil {
        t.Fatal(err)
    }
    t.Logf("input %d bytes -> output %d bytes", buf.Len(), len(doc.Blob))
}
```

Output:

```text
input 56003 bytes -> output 524288000 bytes
```

A 56 KB input produced 500 MiB in memory with no limit. The output size is attacker-controlled, so a larger target OOM-kills the process.

**Expected behavior**

Decompression is bounded. A document that decompresses past a configured limit is rejected instead of being fully buffered, the same way the blob collector already caps object size with `io.LimitReader`.

**Screenshots**

N/A.

**GUAC version**

Built from commit `362e6dacedaa22af63c157b2c9d3e39a51da437f`.

**Additional context**

Availability impact only (memory exhaustion of the processor). Any ingestion source feeding an attacker-influenced artifact reaches it. I have a fix that bounds the decompressed size with `io.LimitReader`, matching the blob collector, with a regression test, and will open a PR.

kanywst:
# Description of the PR

`decompressDocument` read the bzip2/zstd stream with `io.ReadAll` and no size limit, so a small compressed document could expand to an arbitrary size and exhaust the processor's memory. The encoding is auto-detected from the blob's magic bytes, so any ingested artifact can reach this path.

This bounds the decompressed size with `io.LimitReader` and rejects documents that exceed the limit, matching the object-size handling already used by the blob collector (`pkg/handler/collector/blob/blob.go`). A 56 KB zstd input previously materialized 500 MiB in memory; it is now rejected once it passes the limit.

Fixes #3075

# PR Checklist

### DIFF
--- a/pkg/handler/processor/process/process.go
+++ b/pkg/handler/processor/process/process.go
@@ -308,10 +308,17 @@ func decodeDocument(ctx context.Context, i *processor.Document) error {
 	return nil
 }
 
+// maxDecompressedSize caps how many bytes a document may decompress to, so a
+// small blob cannot expand without bound (a decompression bomb).
+var maxDecompressedSize int64 = 1 << 30 // 1 GiB
+
 func decompressDocument(i *processor.Document, reader io.Reader) error {
-	uncompressed, err := io.ReadAll(reader)
+	uncompressed, err := io.ReadAll(io.LimitReader(reader, maxDecompressedSize+1))
 	if err != nil {
 		return fmt.Errorf("unable to decompress document: %w", err)
+	}
+	if int64(len(uncompressed)) > maxDecompressedSize {
+		return fmt.Errorf("decompressed document exceeds the %d byte limit", maxDecompressedSize)
 	}
 	i.Blob = uncompressed
 	return nil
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):

gaganhr94 (approved):
LGTM !

gaganhr94:
Thanks for the contribution @kanywst :)

PR #3153, opened 2026-08-14
fix(ent): avoid lock-order-inversion deadlocks on package_names upserts

### LINKED ISSUE #3149: [ingestion/data-quality issue]: concurrent IngestPackages deadlocks  on shared package_names rows
**Describe the bug**
`IngestPackages` (bulk) and `IngestPackage` (single) can deadlock in Postgres and fail with SQLSTATE `40P01`, causing the entire SBOM ingestion to be discarded.

Root cause has two parts:

1. In `upsertBulkPackage`, `pkgNameCreates` is built in the SBOM's own package order (unsorted) before being passed to `tx.PackageName.CreateBulk(...).OnConflict(...).UpdateNewValues()`. When two concurrent ingestions share overlapping package names (very common, since most images share dependencies), Postgres takes row-exclusive locks on `package_names` rows in whatever order each batch was built. If two transactions lock the same shared rows in opposite order, this is a classic lock-order-inversion deadlock.

2. `UpdateNewValues()` is used on conflict, which forces a row-exclusive lock on every existing row even though the write is a no-op — the row's data (name) and its ID (deterministic hash of the name) never actually change on conflict. This unnecessarily strengthens the lock taken on the most frequently-written rows in the table, making the deadlock far more likely to occur.
Since the batch runs inside the document's graph-assembly transaction, a deadlock abort discards the whole SBOM ingestion, not just the offending batch.

**To Reproduce**
1. Ingest two or more SBOMs concurrently (e.g. via multiple ingestor replicas/pods) that share common package names (e.g. shared base-image dependencies).
2. Repeat ingestion under load until two concurrent `IngestPackages`/`IngestPackage` calls attempt to upsert overlapping `package_names` rows in different orders.
3. Observe ingestion failure.

**Expected behavior**
Concurrent ingestion of SBOMs with overlapping packages should not deadlock, and a transient deadlock (if it occurs) should be retried automatically rather than discarding the whole document.

**Screenshots**
N/A 

**GUAC version**
v1.1.0

**Ingested document(s)**
Any two or more SBOMs that share common package names (e.g., SBOMs for container images built from a common base image) are sufficient to reproduce this when ingested concurrently.

**Additional context**
Suggested fix, in `internal/db/ent/backend/package.go` (or wherever `upsertBulkPackage`/`upsertPackage` live):

1. **Change the conflict strategy from `UpdateNewValues()` to `DoNothing()`** on the `PackageName` upsert. The upsert never actually changes existing row data (name/ID are derived deterministically), so there is nothing to update — using `DoNothing()` avoids taking an unnecessary row-exclusive lock.

2. **Sort `pkgNameCreates` by a deterministic key (e.g. the derived name ID) before the bulk insert**, so concurrent transactions always attempt to lock shared rows in the same order, eliminating lock-order-inversion.

3. **Retry the outer transaction on `40P01`** (with backoff) in `IngestPackages`/`IngestPackage`, since a deadlocked transaction is aborted and cannot be retried from inside itself. This write path is idempotent (deterministic UUIDs), so retrying the whole transaction is safe.

4. **Add a configurable sleep/jitter delay before each ingestion run** (e.g. a `--ingest-delay` flag or random startup stagger per ingestor instance). This doesn't fix the underlying lock-ordering issue, but reduces the window in which two ingestors process overlapping SBOMs at the same instant, lowering deadlock frequency as a stopgap for anyone hitting this before the code fix lands.Happy to open a PR with these changes if useful.

pujitha24:
# Description of the PR

Fixes the two root causes of concurrent `IngestPackages`/`IngestPackage` deadlocks (Postgres SQLSTATE `40P01`) on shared `package_names` rows, as diagnosed in the linked issue:

1. `PackageName` bulk creates in `upsertBulkPackage` were built in unsorted (SBOM-input) order before being passed to `CreateBulk(...).OnConflict(...)`. When two concurrent transactions upsert overlapping package names, Postgres takes row-exclusive locks in whatever order each batch happened to be built in — if two transactions lock the same shared rows in opposite order, that's a lock-order-inversion deadlock. Fixed by sorting each batch's `PackageName` creates by their deterministic ID before the bulk insert, so concurrent transactions always attempt to lock shared rows in the same relative order.
2. `UpdateNewValues()` was used on the `PackageName` conflict clause (both the bulk and single-package upsert paths), forcing an unnecessary row-exclusive lock on every existing row even though nothing ever changes on conflict: `PackageName`'s only fields are `type`/`namespace`/`name` (the conflict/unique-index columns themselves), and its `id` is a deterministic hash of those same fields. Changed to `DoNothing()`, which is behaviorally equivalent here and takes a weaker lock, matching the pattern already used for `PackageVersion`'s conflict clause in this same file.

This intentionally does NOT include the issue's other two suggestions (retrying the outer transaction on `40P01`, and a configurable ingest-delay/jitter flag) — those are more invasive changes (transaction-wrapping code shared by many callers, and new CLI/config plumbing) better suited to a separate, focused PR.

Fixes #3149

# PR Checklist

## Validation

Ran locally (no live Postgres/docker available in this environment):
- `go build ./...` — passes
- `go vet ./...` — clean
- `go test -race ./pkg/assembler/backends/ent/backend/...` — passes
- `gofmt -l` on changed files — clean
- `golangci-lint run ./pkg/assembler/backends/ent/backend/...` — only 2 pre-existing staticcheck issues in `search.go`, a file untouched by this change

The new unit test (`package_sort_test.go`) proves the sort helper's determinism — that PackageName creates converge to the same relative order regardless of starting order, and that IDs stay correctly paired with their creates through the swap — which is the property lock-order-inversion prevention actually depends on. It does not reproduce the live Postgres `40P01` deadlock itself, since no Postgres instance is available in this environment; that reproduction would require the multi-replica concurrent-ingestion setup described in the issue. The `DoNothing()` change is proven correct by inspection of the `PackageName` ent schema (`pkg/assembler/backends/ent/schema/packagename.go`), which has no fields beyond the immutable ID and the three conflict-key columns, so there is no case where an existing row's data could need updating on conflict.

Note: this repo's CI is gated by `.github/workflows/ci-integration.yaml` for integration-level testing; I was unable to run that locally in this environment. Also note: `ci-integration` (jobs `E2E` and `CI for integration tests`) is currently failing on `main` itself (e.g. run for #3150), unrelated to this change.

### DIFF
--- a/pkg/assembler/backends/ent/backend/package.go
+++ b/pkg/assembler/backends/ent/backend/package.go
@@ -270,8 +270,11 @@ func upsertBulkPackage(ctx context.Context, tx *ent.Tx, pkgInputs []*model.IDorP
 			// the conflict columns (name, namespace, type), so an existing row
 			// never actually changes on conflict. DoNothing avoids taking an
 			// unnecessary row-exclusive lock on the most frequently-written rows.
+			// DoNothing means Postgres returns no row via RETURNING on an actual
+			// conflict, which ent surfaces as stdsql.ErrNoRows; that's expected
+			// here, matching the PackageVersion upsert below.
 			DoNothing().
-			Exec(ctx); err != nil {
+			Exec(ctx); err != nil && err != stdsql.ErrNoRows {
 
 			return nil, errors.Wrap(err, "bulk upsert pkgName node")
 		}
@@ -315,9 +318,11 @@ func upsertPackage(ctx context.Context, tx *ent.Tx, pkg model.IDorPkgInput) (*mo
 		OnConflict(sql.ConflictColumns(packagename.FieldName, packagename.FieldNamespace, packagename.FieldType)).
 		// See upsertBulkPackage: PackageName's ID and fields are deterministically
 		// derived from the conflict columns, so DoNothing avoids an unnecessary
-		// row-exclusive lock on conflict.
+		// row-exclusive lock on conflict. As with the PackageVersion upsert
+		// below, a real conflict makes Postgres return no row via RETURNING,
+		// which ent surfaces as stdsql.ErrNoRows; that's expected, not an error.
 		DoNothing().
-		Exec(ctx); err != nil {
+		Exec(ctx); err != nil && err != stdsql.ErrNoRows {
 
 		return nil, errors.Wrap(err, "upsert package name")
 	}
--- a/pkg/assembler/backends/ent/backend/package.go
+++ b/pkg/assembler/backends/ent/backend/package.go
@@ -206,6 +206,20 @@ func generatePackageVersionCreate(tx *ent.Tx, pkgVersionID *uuid.UUID, pkgNameID
 		SetHash(versionHashFromInputSpec(*pkgInput.PackageInput))
 }
 
+// sortablePackageNameCreates sorts a batch of PackageName creates by their
+// deterministic ID, keeping ids and creates in lockstep.
+type sortablePackageNameCreates struct {
+	ids     []string
+	creates []*ent.PackageNameCreate
+}
+
+func (s sortablePackageNameCreates) Len() int           { return len(s.creates) }
+func (s sortablePackageNameCreates) Less(i, j int) bool { return s.ids[i] < s.ids[j] }
+func (s sortablePackageNameCreates) Swap(i, j int) {
+	s.ids[i], s.ids[j] = s.ids[j], s.ids[i]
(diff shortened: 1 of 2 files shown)


mihaimaruseac (approved):

gaganhr94 (approved):
LGTM ! Thanks for the contribution @pujitha24

PR #3157, opened 2026-08-16
feat: add --compute flag for local-only scorecard computation

### LINKED ISSUE #3013: [feature] Scorecard Certifier: add opt-in for usage of Scorecard API

Follow-up to PR #2815 ([review comment](https://github.com/guacsec/guac/pull/2815#discussion_r3107797798)).

Before PR #2815, `NewScorecardCertifier` acted as a configuration validator: if
`GITHUB_AUTH_TOKEN` was unset, construction failed loudly at init. PR #2815
changed this to a warning so the API-only path (which doesn't need the token)
can be used without forcing callers to set a token. This is a subtle behavior change for existing callers.

The fail-fast (the older method of scorecard calculation) which would need a `GITHUB_AUTH_TOKEN` can be a opt-in functionality. The default behaviour of the scorecard-collector will not have any changes (i.e, first hit the scorecard api and then compute using the scorecard package if the api call fails). Introduce a flag `--compute` into scorecard which will only compute the score using scorecard package and will not do scorecard API calls for the scorecard data.

Nikhil-cp1905:
# Description of the PR

Fixes #3013

Adds a `--compute` flag to the scorecard certifier that skips the Scorecard API and computes the score locally with the scorecard package. Opting in also restores the fail-fast `GITHUB_AUTH_TOKEN` check at startup. **Default behaviour is unchanged.**

| Invocation | Score source | Missing `GITHUB_AUTH_TOKEN` |
| --- | --- | --- |
| _(no flag)_ — default | API first, local computation as fallback | Warning only |
| `--compute` | Local computation only | Hard error at startup |

## 

Before #2815, a missing `GITHUB_AUTH_TOKEN` failed loudly at init. #2815 added the API path, which needs no token, so that failure was relaxed to a warning — as noted in [this review comment](https://github.com/guacsec/guac/pull/2815#discussion_r3107797798), operators who relied on the startup check lost it, and the failure now surfaces mid-run instead.

This makes it a choice rather than a default: compute-only is useful for air-gapped networks, or repos the public API has never scanned, and it can't work without a token — so opting in brings the startup check back with it.

## Changes

- **`scorecardRunner.go`** — adds a `computeOnly` field and a `WithComputeOnly(bool)` option. `NewScorecardRunner` is now variadic so existing callers compile unchanged, and errors when compute-only is set without a token. `GetScore` skips the API leg in that mode; the default branch is the existing code, unchanged.
- **`store.go`** — registers `--compute` (default `false`).
- **Both CLIs** — wires the flag through `validateScorecardFlags`, following the existing pattern.
- **`scorecardRunner_compute_test.go`** — new tests, in their own file because `scorecardRunner_test.go` is behind `//go:build integrationMerge` and never runs in normal CI.

## Testing

New unit tests cover the token matrix (default without token OK, compute-only without token errors, compute-only with token OK) and that compute-only reaches local computation instead of the API. Both run without network.

Verified manually: the flag shows in both `--help` outputs; `--compute` with no token exits at startup; without the flag and no token the certifier starts and runs as before. `go build ./...`, `go vet`, and `gofmt` are clean.

# PR Checklist

- [x] If GraphQL schema is changed, GraphQL client updates/additions have been made
- [x] If ent schema is changed, `make generate` has been run


mihaimaruseac (approved):

pkg/certifier/scorecard/scorecardRunner_compute_test.go
```diff
@@ -0,0 +1,88 @@
+//
+// Copyright 2025 The GUAC Authors.
```
gaganhr94: Change the year to 2026 please

gaganhr94 (commented):
Changes look good. Just one nit - can you add these tests in scorecardRunner_test.go. Compute mode tests will not require a new file

pkg/cli/store.go
```diff
@@ -95,6 +95,9 @@ func init() {
 	// add artificial latency to throttle the certifier
 	set.String("certifier-latency", "", "sets artificial latency on the certifier. Defaults to empty string (not enabled) but can set m, h, s...etc")
 
+	// scorecard certifier
+	set.Bool("compute", false, "only compute the scorecard locally with the scorecard package, skipping the scorecard API. Requires GITHUB_AUTH_TOKEN")
```
gaganhr94: Remove the word "locally"

gaganhr94 (approved):

PR #3159, opened 2026-08-17
fix(scorecard): skip checks unsupported at non-HEAD commits

### LINKED ISSUE #3158: Fix the post merge workflow
Currently the post merge workflows are failing. Here is a reference of it:
https://github.com/guacsec/guac/actions/runs/31961008767/job/95198882531

Investigate and come up with a fix for this

Nikhil-cp1905:
# Description of the PR

The post-merge workflow has been red on every run since 2026-06-30. The failure is
`Test_scorecardRunner_GetScore/actual_test#01`:

```
getting enabled checks: internal error: Unsupported RequestType [1] by check: Contributors
```

**Root cause.** When a tag is supplied, `computeScore` resolves it by overwriting `commitSHA`
with the release's `TargetCommitish`. That field is the *branch* a release was cut from, so for
`ossf/scorecard v4.10.4` it is `"main"` — not `"HEAD"`. Scorecard treats any commit other than
`HEAD` as a `CommitBased` request, and when it is handed an explicit check list it rejects the
entire run if even one check does not support that request type. `Contributors` is registered
with no supported request types, so the run aborted before scoring anything.

This is not specific to tags: a genuine pinned commit SHA fails identically, so local
computation at a pinned commit has never worked.

**Fix.** Filter the requested checks by supported request type before passing them to scorecard,
using the same predicate scorecard applies internally (`policy.isSupportedCheck`). At `HEAD` the
list is unchanged; at a pinned commit the 9 checks that can run are kept and the rest are logged
as skipped:

```
Binary-Artifacts, Vulnerabilities, Pinned-Dependencies, CI-Tests, License,
Code-Review, Dangerous-Workflow, Security-Policy, Token-Permissions
```

The 10 dropped checks are HEAD-only in scorecard, so there is no way to evaluate them at a
pinned commit.

Fixes #3158
# PR Checklist


mihaimaruseac (approved):
Thank you

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -300,6 +330,8 @@ func (s scorecardRunner) computeScore(repoName, commitSHA, tag string) (*sc.Resu
 		}
 	}
 
+	checkNames = s.supportedChecks(checkNames, commitSHA)
```
gaganhr94: **Fix the tag resolution in place (the actual defect in #3158).**

A few lines above, tag resolution does `commitSHA = release.TargetCommitish`. `TargetCommitish` is the *branch* a release was cut from, so today we score `main` (moving HEAD of a branch) and attach the result to a tag — the attestation points at the wrong revision. That is the real bug, and it survives this PR.

Replace it with a real tag→commit lookup against GitHub:

1. `GET /repos/{owner}/{repo}/git/ref/tags/{tag}`
2. if `object.type == "tag"` (annotated tag), deref via `GET /repos/{owner}/{repo}/git/tags/{sha}` and take `object.sha`

That yields the commit the tag actually points at, which is what the attestation should reference.

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -250,6 +251,35 @@ func parseRetryAfter(value string, now time.Time) (time.Duration, bool) {
 	return 0, false
 }
 
+// supportedChecks drops the checks scorecard cannot evaluate at a non-HEAD
+// commit. Scorecard treats any commit other than HEAD as a CommitBased request
+// and rejects the whole run if an explicitly requested check does not support
+// that request type, so an unfiltered list fails with "Unsupported RequestType
+// [1] by check: Contributors" rather than scoring the checks that do work.
+func (s scorecardRunner) supportedChecks(checkNames []string, commitSHA string) []string {
```
gaganhr94, line 291: **Keep this — the filtering is genuinely required.** Given the constraint that tagged requests always go to local compute, even a *correct* commit SHA is `CommitBased`, so scorecard still rejects the explicit check list. So this helper is not made redundant by fixing the tag resolution.

One preference, not a blocker: consider dropping `sc.WithChecks(...)` entirely instead of adding this helper — scorecard's default path applies the same predicate internally, and GUAC's 19-name list is just the full set restated. The helper works, so either is fine.

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -250,6 +251,35 @@ func parseRetryAfter(value string, now time.Time) (time.Duration, bool) {
...
+	for _, name := range checkNames {
+		if len(checker.ListUnsupported(required, allChecks[name].SupportedRequestTypes)) == 0 {
+			supported = append(supported, name)
+			continue
+		}
+		skipped = append(skipped, name)
+	}
+
+	if len(skipped) > 0 {
+		logging.FromContext(s.ctx).Infof("Skipping scorecard checks unsupported at commit %s: %s",
```
gaganhr94, line 310: **Surface the reduced coverage more loudly.** Dropping 10 of 19 checks silently degrades the score's meaning, so this should be `Warnf`, not `Infof`. Ideally also record the effective check count in the emitted document, so a consumer can tell a 9-check score from an 18-check one without digging through logs.

pkg/certifier/scorecard/scorecardRunner.go
```diff
@@ -300,6 +330,8 @@ func (s scorecardRunner) computeScore(repoName, commitSHA, tag string) (*sc.Resu
 		}
 	}
 
+	checkNames = s.supportedChecks(checkNames, commitSHA)
+
 	opts := []sc.Option{
 		sc.WithCommitSHA(commitSHA),
 		sc.WithChecks(checkNames),
```
gaganhr94, line 390: If the `sc.WithChecks` call were dropped (see the comment on `supportedChecks`), scorecard would filter by the same predicate itself and this line plus the new helper both go away. Optional.

gaganhr94 (commented):
The check filtering here is correct and worth keeping, but it isn't the defect in #3158 — the tag→commit resolution above it is still wrong, and that part is untouched by this PR. Details inline.

Stating the tradeoff explicitly, since this PR accepts it: with the API flow untouched, every tagged request permanently takes the local-compute fallback and therefore always scores 9 of 19 checks — even though the scorecard API has a full 18-check result for that exact commit sitting behind the guard. Tag-based scores stay non-comparable with HEAD-based ones. That's a legitimate scope decision, but worth being on record so nobody later reads a tag score as equivalent to a HEAD score.

gaganhr94 (changes requested):

claude-progress.md
gaganhr94, line 1: remove this file

gaganhr94 (commented):
let's remove all the comments in the PR. the code itself is self explanatory in this case

gaganhr94:
@Nikhil-cp1905 can you please fix the pipeline failures

gaganhr94 (approved):
LGTM ! Thanks @Nikhil-cp1905 for the fix

gaganhr94:
Screenshot 2026-08-23 at 9 18 09 PM

Issue #208, opened 2022-11-10
[collector] PyPI packages


lumjjb:
Create a PyPI package collector:

The collector should be able to:
- Be configured with a list of packages to collect from, and a URL to the target server
- Either on a poll or watch basis, collect information from packages
- Create SLSA, SBOM or SLSA attestations based on the information collected

PR #3165, opened 2026-08-18
feat: add PyPI package metadata collector

ChrisJr404:
Adds a PyPI collector so GUAC can pull package metadata from the PyPI JSON API and feed it into the supply-chain graph, closing the gap for pypi purls that the deps.dev collector only covers indirectly.

It reads pypi purls from a collect data source and fetches each package's metadata from `https://pypi.org/pypi/<name>/json`, emitting one `processor.Document` per package. I modeled it on the deps.dev collector: `RetrieveArtifacts` runs once or loops on an interval when `poll` is set, packages already fetched are deduped across poll cycles, and non-pypi sources are skipped with a warning. Bare package names work too, not just purls. A new `DocumentPyPI` document type tags the emitted docs.

Scope here is the collector only. The processor/parser side (registering a document parser + guesser for `DocumentPyPI`) is a natural follow-up and I'm happy to send that separately.

Tests spin up an `httptest` server standing in for the PyPI API and serve fixtures from `testdata/`, so they run without network. Cases cover a single purl, a bare name, dedupe of a repeated package, a non-pypi purl getting skipped, and a missing package emitting nothing. `go build ./...`, `go vet`, and `go test ./pkg/handler/collector/pypi/...` all pass.

Refs #208

### DIFF
--- a/pkg/handler/collector/pypi/pypi.go
+++ b/pkg/handler/collector/pypi/pypi.go
@@ -1,0 +1,171 @@
+//
+// Copyright 2024 The GUAC Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


mihaimaruseac (approved):

gaganhr94:
Hi @ChrisJr404, closing this PR since this is not a change planned in the project roadmap.

PR #3166, opened 2026-08-18
feat: add Go module proxy collector

### LINKED ISSUE #215: [collector] Golang module proxy collector
Golang module proxy keeps track of a ton of information for public golang as well as when used in the context of an organization. This could be useful as a data source to provide high quality information for an organization's golang software catalog.

The collector should be able to:
- Be configured with a list of packages to collect from, and a URL to the target server
- Either on a poll or watch basis, collect information from packages
- Create SLSA, SBOM or SLSA attestations based on the information collected

ChrisJr404:
Adds a collector that ingests Go modules straight from the module proxy (proxy.golang.org) so golang purls can be pulled without cloning anything.

For each golang purl it figures out the module path. If the purl carries a version it fetches just that one; if not, it hits `@v/list` and expands to every version the proxy knows about. Per version it grabs the `.info` (version metadata, JSON) and the `.mod` (the go.mod file) and emits a document for each. Bare module paths work too, which matters because a purl lowercases the namespace and the proxy needs the real casing for modules like `github.com/BurntSushi/toml` — that's why the case-encoding (`!` before each uppercase letter) lives in `escapePath` and gets exercised through the bare-path route. Shape follows the deps.dev and pypi collectors: `version.UATransport` client, poll/once modes, dedupe across poll cycles, purl parsing via `packageurl-go`.

I added a `GO_MOD` document type constant alongside the existing ones. This PR is scoped to collection like the issue asks; wiring a processor/parser for the new type is a sensible follow-up.

Tests spin up an `httptest` server backed by a `testdata` tree laid out like the proxy's escaped paths, then drive everything through `collector.Collect`. The `.info` docs go through `dochelper.DocTreeEqual`; go.mod is plain text so those are compared on raw bytes since the tree helper normalizes blobs as JSON. Cases cover a pinned version, version expansion via `@v/list`, the uppercase case-encoding, bare paths, dedupe, non-golang purls getting skipped, and a missing module emitting nothing. `escapePath` also has its own unit test.

`go build ./...`, `go vet`, `gofmt`, and `go test ./pkg/handler/collector/gomod/...` all pass locally.

Fixes #215


mihaimaruseac (approved):

gaganhr94:
Hi @ChrisJr404, closing this PR since this is not a change planned in the project roadmap.

PR #2217, opened 2024-10-25
Included REST Spec Changes

nathannaveen:
# Description of the PR

* Second Part of https://github.com/guacsec/guac/pull/2089
* All the OpenAPI Spec changes have been included
* The `/query/dependencies` is no longer going to be used, instead, the `/v0/package/{purl}/dependencies` and `/v0/artifact/{digest}/dependencies` will be replacing it.
* The code in retrieveDependencies.go has been updated to work for the new endpoints.
* The error handling code has also been updated.

# PR Checklist

- [ ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [ ] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If GraphQL schema is changed, GraphQL client updates/additions have been made
- [ ] If OpenAPI spec is changed, `make generate` has been run
- [ ] If ent schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #3172, opened 2026-08-22
fix(guacrest): restore pagination on purl and deps REST endpoints

### LINKED ISSUE #3170: [bug] guacrest: pagination is declared but not applied on any REST endpoint
**Describe the bug**

`pkg/guacrest/pagination` is implemented and tested but has no non-test callers. None of the six list endpoints paginate — every one returns the full result set and `PaginationInfo.NextCursor` is never set.

Two causes:

**1. Regression on the purl/artifact endpoints.** Pagination was wired up in #1720 and removed in #2217, which dropped the `pagination.Paginate` call from `retrieveDependencies.go` and the `PaginationSpec` param from `openapi.yaml`. #3001 and #3002 later set `TotalCount` inline (`server/server.go:130`, `:195`, `:254`) instead of restoring the call, so responses carry a count but no cursor.

Affects `getPackagePurls`, `getPackageDeps`, `getArtifactDeps`.

**2. Endpoints returning a bare array have nowhere to put a cursor.** `analyzeDependencies`, `getPackageVulns` and `getArtifactVulns` respond with `PackageNameList` (`openapi.yaml:292`) and `VulnerabilityList` (`openapi.yaml:300`), both plain `type: array` with no `PaginationInfo` envelope — unlike `PurlList`. `analyzeDependencies` is the only path still declaring `PaginationSpec` (`openapi.yaml:36`) and it ignores it (`server/server.go:70` reads only `Sort`). The vulns handlers return the slice straight through (`server.go:164`, `:223`); `getPackageVulns?includeDependencies=true` is unbounded.

**To Reproduce**

```bash
curl 'localhost:8081/v0/package/pkg%3Aguac%2Ffoo' | jq .PaginationInfo
# {"TotalCount": 3}  -- NextCursor absent

curl 'localhost:8081/analysis/dependencies?sort=frequency&PageSize=1' | jq length
# full result set -- paginationSpec ignored
```

**Expected behavior**

`paginationSpec` is honoured on all six list endpoints and `NextCursor` is returned when more results remain.

**Proposed fix**

- Non-breaking: restore the `pagination.Paginate` call in the three purl/artifact handlers and re-add `PaginationSpec` to those paths in `openapi.yaml`.
- Breaking: give `PackageNameList` and `VulnerabilityList` the same envelope shape as `PurlList` so `analyzeDependencies`, `getPackageVulns` and `getArtifactVulns` can paginate consistently with the rest. For `analyzeDependencies`, `GetDependenciesBySortedDependentCnt` (`pkg/dependencies/dependents.go:49`) builds and sorts the whole set up front, so the limit is best pushed down rather than applied to the finished array.

pujitha24:
# Description of the PR

`pkg/guacrest/pagination` is implemented and tested but was never invoked by three
of guacrest's REST list endpoints. `getPackagePurls`, `getPackageDeps`, and
`getArtifactDeps` used to call `pagination.Paginate` (added in #1720), but a later
refactor (#2217) dropped that call and the `PaginationSpec` openapi parameter for
these three paths. Follow-up PRs (#3001/#3002) patched in a manually-computed
`TotalCount` but never restored real pagination, so these endpoints always return
the full result set and `PaginationInfo.NextCursor` is never set.

This PR restores the non-breaking part of the fix described in the issue: it
re-adds the `PaginationSpec` query parameter to the three affected paths in
`openapi.yaml` and wires `request.Params.PaginationSpec` through to
`pagination.Paginate` in the three handlers, matching how `analyzeDependencies`
already declares (but doesn't yet honor) `PaginationSpec`.

It deliberately does NOT touch `analyzeDependencies`, `getPackageVulns`, or
`getArtifactVulns` — those return bare arrays with no `PaginationInfo` envelope,
so paginating them requires a breaking schema change. That's a separate, larger
proposal per the issue and is left out of scope here.

**Approach**

- `openapi.yaml`: added `$ref: "#/components/parameters/PaginationSpec"` to the
  `getPackagePurls`, `getPackageDeps`, and `getArtifactDeps` paths, then ran
  `go generate ./...` to regenerate `generated/*.go` and `client/*.go`.
- `server.go`: replaced the manual `TotalCount := len(...)` construction in the
  three handlers with a real `pagination.Paginate(ctx, ..., request.Params.PaginationSpec)`
  call, returning the existing 400 response shape if `Paginate` errors (e.g.
  invalid cursor). `GetPackageDeps`/`GetArtifactDeps` get their purls back as a
  `map[string]string`; since Go map iteration order is randomized per call, the
  flattened slice is explicitly sorted (`sort.Strings`) before pagination so that
  a `NextCursor` from one request reliably lines up with the next request's slice
  ordering. `GetPackagePurls` doesn't need this since its source
  (`FindMatchingPurls`) already returns a slice.

**Validation**

- `go build ./...` passes.
- `go test ./pkg/guacrest/...` passes, including new subtests that: (a) set a
  small `PageSize` and assert `PurlList` is truncated and `NextCursor` is set,
  and (b) for the two map-backed endpoints, follow the returned `NextCursor` into
  a second request and assert the second page doesn't repeat the first page's
  item — this specifically catches the map-ordering hazard described above. I
  confirmed this test actually catches that regression: with `sort.Strings`
  removed, 10 consecutive runs of `go test ./pkg/guacrest/server/... -run
  Test_DepsEndpoints_RespectPaginationSpec -count=1` failed roughly half the
  time; with it restored, 10/10 runs passed.
- `go generate ./pkg/guacrest/...` is idempotent (rerunning produces no further
  diff).
- `golangci-lint run ./pkg/guacrest/...` reports 0 issues in the changed files
  (pre-existing staticcheck findings in `pkg/guacrest/helpers/errors.go` and
  `pkg/guacrest/pagination/*.go` are untouched by this diff).
- `gofmt -l` / `goimports -l` report no issues on the hand-written files.
- Note: this repo's `Lint` GitHub Actions job (the `golangci-lint` step) is
  currently failing on `main` at the base commit this branch is built from,
  for reasons unrelated to this change.

Fixes #3170

# PR Checklist

- [x] All CI checks are passing (tests and formatting) — except the pre-existing `Lint` failure on `main` noted above, unrelated to this change


gaganhr94:
Functionally this looks right — the three handlers now run their result sets through `pagination.Paginate` and return a real cursor.

One nit before merge: the comments in the hand-written files can be tighter, and a few are repeated verbatim.

- `pkg/guacrest/server/retrieveDependencies_test.go` — the three-line "Follow NextCursor and confirm the second page is disjoint from the first, i.e. the underlying ordering used for pagination is stable across requests" block appears twice, once per subtest. Same for `retrievePackagePurls_test.go`. Since both subtests are doing the identical second-page assertion, say it once at the top of the test (or in the shared helper) and drop the per-subtest repeats — or shorten to a single line, e.g. `// second page must be disjoint from the first (stable ordering)`.
- The test's leading doc comment already says what the test covers; the inline comments then restate it. One or the other is enough.
- `pkg/guacrest/server/server.go` — `GetPackageDeps` and `GetArtifactDeps` now have byte-identical copy/sort/paginate blocks. Worth pulling into a small helper; that removes the duplicated code and the need to comment it twice.

Nothing blocking, just cleanup.

pkg/guacrest/server/retrieveDependencies_test.go
gaganhr94, line 1: nit: please try to keep the comments concise here, and avoid redundant comments

gaganhr94 (approved):
LGTM ! Thanks for the contribution @pujitha24

mihaimaruseac (approved):
Thank you!

PR #3173, opened 2026-08-22
fix(guacrest): require <algorithm>:<digest> for artifact REST endpoints

### LINKED ISSUE #3171: [bug] guacrest: artifact endpoints should require <algorithm>:<digest>
**Describe the bug**

The artifact endpoints take a bare digest with no algorithm. `FindArtifactWithDigest` (`pkg/guacrest/helpers/artifact.go:32`) sets only `Digest` on the spec and leaves `Algorithm` nil:

```go
filter := gql.ArtifactSpec{Digest: &digest}
```

Since ent stores algorithm and digest in separate columns, `sha256:abc...` never matches — only `abc...` works. Dropping the algorithm also means the same hex ingested under two algorithms matches both and returns a 500 (`helpers/artifact.go:41`).

Affects `getArtifactVulns` and `getArtifactDeps`.

**Expected behavior**

Require `<algorithm>:<digest>`. Parse the prefix, set both `Algorithm` and `Digest` on `ArtifactSpec`, and return 400 when the prefix is missing or unrecognised.

Breaking for callers passing a bare digest, which is what the endpoints accept today.

pujitha24:
# Description of the PR

The REST artifact endpoints (`GET /v0/artifact/{digest}/vulns` and
`GET /v0/artifact/{digest}/dependencies`) documented their path parameter as
`<algorithm>:<digest>` in the OpenAPI spec, but `FindArtifactWithDigest`
only ever set the `Digest` field on the `ArtifactSpec` filter, never
`Algorithm`. Since ent stores algorithm and digest in separate columns, a
correctly-formed `sha256:abc...` identifier never matched anything (only the
bare hex `abc...` worked), and if the same hex digest existed under two
different algorithms, the filter matched both rows and returned a 500.

This fixes `FindArtifactWithDigest` to parse the `<algorithm>:<digest>`
prefix, set both `Algorithm` and `Digest` on the filter, and return a 400
when the prefix is missing or malformed (empty algorithm or digest half).

This is a breaking change for any caller currently passing a bare digest,
which is what the endpoints accepted before this fix (and what the
documented API contract never actually supported).

# PR Checklist

Fixes #3171

### DIFF
--- a/pkg/guacrest/helpers/artifact.go
+++ b/pkg/guacrest/helpers/artifact.go
@@ -18,18 +18,25 @@ package helpers
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/Khan/genqlient/graphql"
 	gql "github.com/guacsec/guac/pkg/assembler/clients/generated"
 	"github.com/guacsec/guac/pkg/logging"
 )
 
-// Queries for an artifact by digest and returns its id. The digest must uniquely
-// identify an artifact, otherwise an error is returned.
+// Queries for an artifact by its <algorithm>:<digest> identifier and returns its
+// id. The identifier must uniquely identify an artifact, otherwise an error is
+// returned.
 func FindArtifactWithDigest(ctx context.Context, gqlClient graphql.Client,
-	digest string) (gql.AllArtifactTree, error) {
+	algorithmDigest string) (gql.AllArtifactTree, error) {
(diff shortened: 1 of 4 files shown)


gaganhr94 (approved):
LGTM !

mihaimaruseac (approved):
Thank you!

PR #3188, opened 2026-08-26
fix(guacrest): resolve artifact vulns through the SBOM graph

### LINKED ISSUE #3174: [bug] guacrest: getArtifactVulns returns empty for SBOM subject artifacts
**Describe the bug**

`getArtifactVulns` returns an empty array for an artifact that is the subject of an ingested SBOM, even when the SBOM's packages have `CertifyVuln` records.

`packageIDsForArtifact` (`pkg/guacrest/server/retrieveVulns.go:74`) only traverses `IsOccurrence`:

```go
gql.Neighbors(ctx, gqlClient, artID, []gql.Edge{gql.EdgeArtifactIsOccurrence})
gql.Neighbors(ctx, gqlClient, occ.Id, []gql.Edge{gql.EdgeIsOccurrencePackage})
```

An SBOM subject artifact reaches its packages via `EdgeArtifactHasSbom` -> `EdgeHasSbomIncludedSoftware`, which is never queried. The first call returns no neighbours, `pkgIDs` stays empty, and the handler returns the initialised-empty slice (`retrieveVulns.go:114`) with a 200.

`/v0/artifact/{digest}/dependencies` on the same digest works, because `byDigest.getDirectDependencies` (`retrieveDependencies.go:145`) does traverse `HasSBOM`. The two artifact endpoints use different traversals.

**To Reproduce**

1. Ingest an SBOM whose subject is an artifact, with vulnerabilities certified against its packages
2. `curl 'localhost:8081/v0/artifact/<digest>/dependencies'` -> packages returned
3. `curl 'localhost:8081/v0/artifact/<digest>/vulns'` -> `[]`

**Expected behavior**

Vulnerabilities certified against the packages reachable from the artifact, consistent with what `/dependencies` reports for the same digest.

**Proposed fix**

Have `GetVulnsForArtifact` use the `byDigest` traversal already used by `GetDepsForArtifact` (`retrieveDependencies.go:378`) to build the package ID set, then pass it to the existing `vulnsForPackageIDs`. `GetVulnsForPackage` already consumes `GetDepsForPackage` this way (`retrieveVulns.go:55`). This also picks up `HasSLSA`, `HashEqual` and transitive packages.

Existing tests only cover the `Artifact -> IsOccurrence -> package` shape (`retrieveVulns_test.go:56`); a `HasSBOM` case is needed.

Nikhil-cp1905:
<html><head></head><body><p>Fixes #3174</p>
<h2>Problem</h2>
<p><code>GET /v0/artifact/{digest}/vulns</code> returned <code>[]</code> with <code>200 OK</code> for any artifact
that is the subject of an ingested SBOM. That's indistinguishable from a
genuinely clean artifact, so a release gate reading this endpoint would pass a
vulnerable one. <code>/dependencies</code> answered correctly for the same digest.</p>
<h2>Root Cause</h2>
<p><code>packageIDsForArtifact</code> only walked <code>artifact -&gt; IsOccurrence -&gt; package</code>.</p>
<h2>Fix</h2>
<p><code>packageIDsForArtifact</code> now runs the same <code>byDigest</code> traversal <code>/dependencies</code>
uses, via a new <code>getReachableNodes</code> split out of <code>getTransitiveDependencies</code>.
The split is needed because the dependency walk drops nodes equivalent to the
start node, and a CVE on the artifact's own package is the artifact's CVE.</p>
<p>Also filters <code>novuln</code> records ("scanned, found nothing"), which would
otherwise be reported as one finding per clean package in an SBOM.</p>
<h2>Defects Found Along the Way</h2>
<p>Routing vulns through the shared traversal surfaced three existing defects.
All are fixed at the choke point, so <code>/dependencies</code> benefits too:</p>

Defect | Detail
-- | --
SLSA edge over-expansion | EdgeArtifactHasSlsa resolves both ways, so walking it from a build material reached every sibling material of an unrelated build. Attestations are now expanded only when the current node is the subject.
Nil-node panic | neighborToNode returns a nil node for a package name node, which the traversal could dequeue and panic on. neighbors now drops them.
Wrong error status | handleErr compared sentinels by identity while callers wrap with %w, so a GraphQL outage returned 400 instead of 502.

<p>Only the keyvalue backend was tested locally; relying on CI for <code>ent</code> and <code>arango</code>.</p>
<p><code>generated/spec.go</code> is the regenerated <code>openapi.yaml</code> — only the encoded blob
changed, and it re-chunks in full because the document is one compressed
stream.</p>
<h2>PR Checklist</h2>
<ul>
<li>[] All commits have a <a href="https://wiki.linuxfoundation.org/dco">Developer Certificate of Origin (DCO)</a> (<code>-s</code> flag to <code>git commit</code>)</li>
<li>[x] All new changes are covered by tests</li>
<li>[ ] If GraphQL schema is changed, <code>make generate</code> has been run — n/a</li>
<li>[ ] If GraphQL schema is changed, GraphQL client updates/additions have been made — n/a</li>
<li>[x] If OpenAPI spec is changed, <code>make generate</code> has been run</li>
<li>[ ] If ent schema is changed, <code>make generate</code> has been run — n/a</li>
<li>[ ] If <code>collectsub</code> protobuf has been changed, <code>make proto</code> has been run — n/a</li>
<li>[x] All CI checks are passing (tests and formatting)</li>
<li>[x] All dependent PRs have already been merged</li>
</ul></body></html>Fixes #3174


pkg/guacrest/server/retrieveDependencies_test.go
```diff
@@ -208,6 +208,38 @@ func Test_RetrieveDependencies_ByPurl(t *testing.T) {
 			purl:           "pkg:guac/foo",
 			expectedByName: []string{"pkg:github/package-url/purl-spec"},
 		},
+		{
+			name: "package -> occurrence -> artifact -> SLSA -> material -> occurrence -> package",
```
gaganhr94: can this be a sentence instead

pkg/guacrest/server/traversal_internal_test.go
gaganhr94, line 1: see if this test file can be removed entirely and cover these cases in one of the existing test files

gaganhr94:
there are a few more testcases with the x -> y -> z pattern, see if you can make all of them into readable sentences

gaganhr94 (approved):

mihaimaruseac (approved):

PR #3190, opened 2026-08-28
fix(guacrest): implemented scorecard sort

### LINKED ISSUE #3175: [feature] guacrest: implement sort=scorecard for /analysis/dependencies

`/analysis/dependencies` supports `sort=frequency` but not `sort=scorecard`, so there is no way to ask which dependencies are the weakest by OpenSSF score. The value is already in the OpenAPI enum (`openapi.yaml:48`) and passes validation, but the handler returns a raw Go error (`server/server.go:98`), which surfaces as a bare 500.

Implement `sort=scorecard`: return packages ordered by ascending OpenSSF aggregate score, as the parameter description states.

Notes for implementation:

- `Scorecards` is already available in the generated client (`clients/generated/operations.go:39334`); `AllCertifyScorecard` exposes `aggregateScore` and per-check scores.
- `CertifyScorecard` attaches to a **source**, not a package, so packages must be mapped to sources (`HasSourceAt`) before sorting. Packages with no linked source need a defined position in the ordering.
- The `PackageName` response schema (`openapi.yaml:219`) only carries `Name` and `DependentCount` — there is nowhere to return the score. Adding it is additive but changes the response shape.
- Overlaps #3170: this endpoint's response is a bare array and does not paginate.

Until the sort is implemented, the handler should return a typed 400 or 501 rather than a bare 500 for a value the spec declares valid.

Nikhil-cp1905:
Fixes #3175

Before this change the value was accepted by spec validation but the handler
returned a bare Go error, so a documented enum value produced a 500.

## How packages are scored

A Scorecard certifies a source, not a package, so packages are scored through
their `HasSourceAt` links: package -> source -> that source's most recent
`CertifyScorecard`.

Two details are worth a reviewer's attention because the issue does not specify
them:

- **Sources are matched on type, namespace and name, with tag and commit
  dropped.** The Scorecard certifier records the commit it scanned; the
  `HasSourceAt` links produced by the deps.dev collector carry neither tag nor
  commit. Keying on them would match nothing. The consequence is that scans
  across all refs of a repository merge into one key and the most recent wins,
  so a repo scanned more recently on a side branch reports that branch's score.
- **A package linked to several sources takes the lowest score,** on the basis
  that it is only as well maintained as the worst repository it is built from.

Ordering rules: ascending by score; packages with no score sort last, since an
absent score is not a bad score and `0` is a real Scorecard value; ties
name so repeated requests return the same order.

## Other changes

- `PackageName` gains an optional `ScorecardScore` field. Additive, and absent
  from `sort=frequency` responses.
- An unrecognised sort now returns a typed 400 instead of a bare 500, which is
  the alternative the issue asks for.
- `internal/testing/graphqlClients`: `GuacData` gains `HasSourceAts` and
  `CertifyScorecards`. This also fixes `ingestSource`, which never set
  every source in a test collapsed onto a single empty-named `SourceName` node.
  No existing test relied on that behaviour.

## Testing

Unit tests cover ascending order, zero vs absent score, most-recent-sca
equal-timestamp resolution, multi-source lowest, a linked source with no
scorecard, and name ordering. Handler tests cover both sorts and the 40

Verified end to end against `guacgql` (keyvalue) and `guacrest` with da
ingested over GraphQL: every rule above held, eight consecutive requests
returned identical ordering, and no errors appeared in either server lo

# PR Checklist


gaganhr94 (approved):
LGTM !

mihaimaruseac (approved):

PR #2033, opened 2024-07-17
Implemented a Delete functionality for KeyValue

nathannaveen:
# Description of the PR

* Fixes https://github.com/guacsec/guac/issues/1983

# PR Checklist

- [ ] All commits have [a Developer Certificate of Origin (DCO)](https://wiki.linuxfoundation.org/dco) -- they are generated using `-s` flag to `git commit`.
- [ ] All new changes are covered by tests
- [ ] If GraphQL schema is changed, `make generate` has been run
- [ ] If GraphQL schema is changed, GraphQL client updates/additions have been made
- [ ] If OpenAPI spec is changed, `make generate` has been run
- [ ] If ent schema is changed, `make generate` has been run
- [ ] If `collectsub` protobuf has been changed, `make proto` has been run
- [ ] All CI checks are passing (tests and formatting)
- [ ] All dependent PRs have already been merged

PR #3191, opened 2026-08-28
feat: KeyValue add delete node functionality 

### LINKED ISSUE #1983: [feature] KeyValue add delete node functionality

Based on the new PR: https://github.com/guacsec/guac/pull/1982

ENT backend has been updated but the same changes need to be made for the keyvalue backend. Implementation of delete node (and relationships) for `CertifyVuln`, `hasSBOM`, and `hasSLSA`.

Implementation of delete node (and relationships) for `CertifyVuln`, `hasSBOM`, and `hasSLSA` for the keyvalue backend

Nikhil-cp1905:
# Add Delete Support to the KV Backend

Fixes #1983

## Summary

Adds a `Delete` capability to the key-value (KV) backend. This builds on the earlier approach in #2033 (by @nathannaveen), rebased onto `main` with several correctness issues resolved.

### 1. New `Remove` method on `kv.Store`
*(`pkg/assembler/kv/kv.go`)*

Implemented across all backends:

| Store | Implementation |
|---|---|
| memmap | Deletes from the collection map; returns `NotFoundError` if absent |
| redis | Uses `HDel`; returns `NotFoundError` if no fields were removed |
| tikv | Checks existence, then deletes the collection/key pair |
| stablememmap (test store) | Delegates to the wrapped memmap |

> ⚠️ **Breaking change**: any out-of-tree `kv.Store` implementation must add this method.

### 2. Shared delete helpers
*(`keyvalue/backend.go`)*

- **`delkv`** – removes a node and its ID from the index, so it stops resolving via `Node()`
- **`removeLink`** – removes one ID from a node's back-edge list (mirrors the existing `setXLinks` helpers)

### 3. Delete implementations

Each cleans up back-edges before removing the node:

- **`deleteCertifyVuln`** – package version + vulnerability edges
- **`deleteHasSLSA`** – subject artifact, deduped `builtFrom` materials, builder
- **`deleteHasSBOM`** – subject + included `isDependency`/`isOccurrence` nodes (via new `deleteIsDependency` / `deleteIsOccurrence` helpers)

### 4. Dispatch logic
*(`keyvalue/path.go`)*

`Delete` takes the write lock and dispatches by node type. Unknown types return `(false, nil)` instead of erroring, matching `matchingEntBackend.Delete` behavior.

## Notable Fix: Shared SBOM Includes

Since included nodes are deduped by key, two `HasSBOM` records can reference the same included node. Previously, deleting one SBOM left a dangling ID that broke **all** `HasSBOM` queries.

**Fix**: `convHasSBOM` now skips included IDs that return `NotFoundError` instead of failing the whole query.

**Trade-off**: real store corruption will now show up as a silently shorter `includes` list rather than an error. Refcounting included nodes would remove this trade-off — flagged as a follow-up.

PR Checklist

- [x] All commits have a Developer Certificate of Origin (DCO) -- they are generated using -s flag to git commit.
- [ ] If GraphQL schema is changed, make generate has been run
- [ ] If OpenAPI spec is changed, make generate has been run
- [ ] If ent schema is changed, make generate has been run
- [ ] If collectsub protobuf has been changed, make proto has been run

### DIFF
--- a/internal/testing/backend/main_test.go
+++ b/internal/testing/backend/main_test.go
@@ -97,11 +97,11 @@ var skipMatrix = map[string]map[string]bool{
 	// redis order issues, tikv: not a valid VexStatus
 	"TestVEXBulkIngest": {arango: true, redis: true, tikv: true},
 	// redis order issues
-	"TestFindSoftware":  {redis: true, arango: true},
+	"TestFindSoftware": {redis: true, arango: true},
 	// remove these once its implemented for the other backends
-	"TestDeleteCertifyVuln":              {arango: true, memmap: true, redis: true, tikv: true},
-	"TestDeleteHasSBOM":                  {arango: true, memmap: true, redis: true, tikv: true},
-	"TestDeleteHasSLSAs":                 {arango: true, memmap: true, redis: true, tikv: true},
+	"TestDeleteCertifyVuln":              {arango: true},
+	"TestDeleteHasSBOM":                  {arango: true},
+	"TestDeleteHasSLSAs":                 {arango: true},
 	"TestQueryPackagesListForScan":       {arango: true, redis: true, tikv: true},
 	"TestBatchQueryPkgIDCertifyVuln":     {arango: true, redis: true, tikv: true},
 	"TestBatchQueryPkgIDCertifyLegal":    {arango: true, redis: true, tikv: true},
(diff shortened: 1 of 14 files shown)


mihaimaruseac (approved):

gaganhr94 (approved):
LGTM ! Thanks Nikhil :)

PR #3196, opened 2026-08-29
fix(guaccollect): warn when --enable-prometheus is set on unsupported subcommands

### LINKED ISSUE #3194: guaccollect accepts --enable-prometheus on every subcommand but only deps-dev honours it
`--enable-prometheus` is registered as a persistent flag on the `guaccollect`
root command (`cmd/guaccollect/cmd/root.go:40`), so every subcommand accepts it.
Only one subcommand reads it.

**Code** (at `1e98dab2`): the sole consumer is
`cmd/guaccollect/cmd/deps_dev.go:92` (`viper.GetBool("enable-prometheus")`).
The other twelve subcommands — `blob`, `datadog_malware`, `eol`, `files`, `gcs`,
`github`, `kubescape`, `license`, `oci`, `osv`, `s3`, `scorecard` — never read
the value.

**Impact:** `guaccollect osv --enable-prometheus` (and the same for any
subcommand but `deps-dev`) is accepted, exits 0, starts no metrics server, and
reports nothing. The user has no signal that the flag did nothing. Same applies
to `GUAC_ENABLE_PROMETHEUS`, since the root binds viper with the `GUAC` prefix.

**Related inconsistency:** `--prometheus-port` is *not* a root persistent flag —
it is registered locally on `deps-dev` only
(`cmd/guaccollect/cmd/deps_dev.go:211`). So on every other subcommand you can
ask for Prometheus but cannot set its port, which is a good indication the
persistent registration of `--enable-prometheus` was not intended to be
binary-wide.

**Possible fixes**, in rough order of preference:

1. Wire the metrics server up in the shared collector path so the flag works
   everywhere it is accepted (and promote `--prometheus-port` to match).
2. Move both flags off the root and register them locally on `deps-dev`, so the
   other subcommands reject them with `unknown flag`.
3. At minimum, log a warning when the flag is set on a subcommand that ignores
   it.

Option 1 is the one that matches user expectation — the flag is currently
documented in guac-docs as a general `guaccollect` flag.

pujitha24:
# Description of the PR

`--enable-prometheus` is registered as a persistent flag on the `guaccollect`
root command, so all 13 subcommands accept it, but only `deps_dev` actually
reads it and starts a metrics server. The other 12 subcommands (`blob`,
`datadog_malware`, `eol`, `files`, `gcs`, `github`, `kubescape`, `license`,
`image`, `registry`, `osv`, `s3`, `scorecard`) silently accept the flag,
exit 0, start no metrics server, and give the user no signal that it did
nothing.

This PR implements the minimal fix suggested in the issue: a warning is now
printed to stderr when `--enable-prometheus` (or `GUAC_ENABLE_PROMETHEUS`) is
set on any subcommand other than `deps_dev`. This does not change runtime
behavior for any subcommand — the flag is still silently a no-op on the other
12 subcommands — it only makes that no-op visible to the user instead of
silent.

# PR Checklist

Fixes #3194

### DIFF
--- a/cmd/guaccollect/cmd/osv.go
+++ b/cmd/guaccollect/cmd/osv.go
@@ -71,6 +71,10 @@ type osvOptions struct {
 	addVulnMetadata bool
 	// enable otel
 	enableOtel bool
+	// enable prometheus server
+	enablePrometheus bool
+	// prometheus address
+	prometheusPort int
 }
 
 var osvCmd = &cobra.Command{
@@ -105,6 +109,8 @@ you have access to read and write to the respective blob store.`,
 			viper.GetInt("last-scan"),
 			viper.GetBool("add-vuln-metadata"),
 			viper.GetBool("enable-otel"),
+			viper.GetBool("enable-prometheus"),
+			viper.GetInt("prometheus-port"),
 		)
 		if err != nil {
 			fmt.Printf("unable to validate flags: %v\n", err)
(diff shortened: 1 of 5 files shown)


mihaimaruseac (approved):

gaganhr94:
I think we can go with option 1 proposed in the issue. We can wire up the prometheus for all the other commands as well, that would be a better fix.

pujitha24:
Fair point, option 1 is the more complete fix and matches what users would actually expect from a persistent flag. It's a bigger change than what's here though — it means wiring up the metrics server (currently only started in `deps_dev.go`) across the other 12 subcommands, and promoting `--prometheus-port` off its local-only registration to match. Given this PR already has an approval on the narrower warning-only fix, do you want me to expand this PR to do the full wiring, or would you rather land this as-is and track option 1 as a follow-up?

gaganhr94:
It would be better to go with the full wiring

pujitha24:
Looked into what full wiring actually involves before touching anything: `deps_dev` is the only collector that has any Prometheus instrumentation today (a `Metrics MetricCollector` field wired up in `pkg/handler/collector/deps_dev/deps_dev.go`). None of the other 12 collectors/certifiers collect any metrics at all, so this isn't just "start an HTTP server in each command" — it means adding metrics instrumentation to each of those implementations first, then wiring the server per subcommand. That's a much bigger, per-collector design task than I can confidently do correctly in one pass here.

I'd rather not guess at what metrics each collector/certifier should expose. Given the scope, would it be OK to keep this warning-only fix as the immediate mitigation and open a separate tracked issue for instrumenting the other collectors one at a time? Or if you'd prefer, I can take a first pass at just one other subcommand (e.g. `osv`) as a template for review before doing the rest.

gaganhr94:
Yeah, can you try implementing it for one of the subcommands, and based on that, we can take a call on the rest

pujitha24:
Pushed the osv template you asked for and CI is green — happy to extend it to the rest of the subcommands once you get a chance to look, or adjust the approach if this isn't quite the shape you had in mind.

gaganhr94:
Changes look good for osv. Can you go ahead and extend it to the rest of the subcommands as well. Also please make sure to keep the comments minimal.

gaganhr94:
The changes look good. Can you please extend it to the other subcommands as well.

gaganhr94:
@pujitha24 wire it up for all the subcommands mentioned in the PR description

PR #3199, opened 2026-08-30
fix(ent): avoid lock-order-inversion deadlocks on occurrence upserts

### LINKED ISSUE #3198: [ingestion/data-quality issue] concurrent SBOM ingestion drops documents with Postgres deadlock (40P01) in the occurrence bulk upsert
**Describe the bug**

Ingesting SBOMs concurrently against the Postgres/ent backend drops documents with a Postgres deadlock in the `isOccurrence` bulk upsert:

```
unable to ingest document "file:///<file_name>.json" : error assembling graphs for "file:///sbom-data/<file_name>.json" : isOccurrencesPkg failed with error: input: ingestOccurrences IngestOccurrences :: bulk upsert Occurrence node: insert nodes to table "occurrences": pq: deadlock detected (40P01)
```

The document is lost — the error surfaces at `cmd/guacingest/cmd/ingest.go:137` and the batch is not retried.

Two things line up with this:

1. `upsertBulkOccurrences` builds each `CreateBulk` batch in SBOM order and inserts it as-is (`pkg/assembler/backends/ent/backend/occurrence.go:225-233`). Two transactions ingesting SBOMs that share packages/artifacts therefore acquire the same row locks in different orders, which is exactly the deadlock shape already fixed for package names by sorting creates by deterministic ID (`pkg/assembler/backends/ent/backend/package.go:257-263`). The occurrence path never got the equivalent treatment.

2. The existing retry helper only matches SQLSTATE `23503` (foreign-key violation) and deliberately excludes everything else (`pkg/assembler/backends/ent/backend/retry.go:27-43`). A `40P01` deadlock is the textbook retryable error — Postgres aborts one transaction precisely so it can be replayed — but here it propagates straight to the caller.

**To Reproduce**

1. Run GUAC with the ent/Postgres backend and more than one concurrent ingestion worker.
2. Ingest a set of SBOMs that share a meaningful number of packages and artifacts (same base image or shared dependency tree is enough).
3. Watch the ingestor logs.
4. Some documents fail with `bulk upsert Occurrence node: ... pq: deadlock detected (40P01)`; those SBOMs are absent from the graph.

**Expected behavior**

Concurrent ingestion of overlapping SBOMs completes without deadlocks, and a deadlock that does occur is retried rather than dropping the document. Concretely: order the occurrence `CreateBulk` batch deterministically the way the package-name upsert does, and extend the retry helper to treat `40P01` (and arguably `40001` serialization failure) as retryable at the transaction level.

**GUAC version**

Built from `main`, ent/Postgres backend.

**Ingested document(s)**

Not specific to one document — reproduces with any set of concurrently-ingested SBOMs with overlapping packages and artifacts. Failures are non-deterministic and shift between documents run to run, which is itself consistent with lock-ordering rather than bad input.

**Additional context**

- Retrying a `40P01` correctly means retrying the whole transaction, not just the failing `CreateBulk` — the transaction is already aborted by the time the error is returned, so the fix is not a drop-in use of `retryOnFKViolation` at the batch level.
- Worth auditing the other bulk upsert paths for the same unsorted-batch pattern rather than fixing only `occurrences`.

pujitha24:
# Description of the PR

Concurrent SBOM ingestion against the Postgres/ent backend can deadlock (SQLSTATE 40P01) in the `isOccurrence` bulk upsert: `upsertBulkOccurrences` built each `CreateBulk` batch in SBOM (unsorted) order, so two concurrent transactions ingesting SBOMs that share packages/artifacts could acquire the same row-exclusive locks in different orders. This is the exact lock-order-inversion shape already fixed for `PackageName` upserts in `upsertBulkPackage` (see PR #3153) — the `isOccurrence` path never got the equivalent treatment.

This PR sorts each Occurrence batch by its deterministic occurrence ID (a hash of subject + artifact + justification/origin/collector/documentRef — the same tuple that identifies the row for conflict purposes) immediately before `CreateBulk`, so concurrent transactions always attempt to lock shared occurrence rows in the same relative order. The function's returned `ids` (used by the caller to map back to GraphQL global IDs) are left in their original, unsorted order — only the local slice passed to `CreateBulk` is reordered.

Deliberately out of scope, matching how PR #3153 scoped the PackageName fix: retrying the outer transaction on 40P01/40001 (this would mean changing the generic `WithinTX` wrapper used by nearly every `Ingest*` function in the backend — a much larger, riskier change), and auditing other bulk-upsert paths for the same pattern. Both are better suited to a separate PR.

# PR Checklist

Fixes #3198

### DIFF
--- a/pkg/assembler/backends/ent/backend/occurrence.go
+++ b/pkg/assembler/backends/ent/backend/occurrence.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	stdsql "database/sql"
 	"fmt"
+	"sort"
 
 	"entgo.io/contrib/entgql"
 	"entgo.io/ent/dialect/sql"
@@ -167,6 +168,20 @@ func occurrenceConflictColumns() []string {
 	}
 }
 
+// sortableOccurrenceCreates sorts a batch of Occurrence creates by their
+// deterministic ID, keeping ids and creates in lockstep.
+type sortableOccurrenceCreates struct {
+	ids     []string
+	creates []*ent.OccurrenceCreate
+}
+
+func (s sortableOccurrenceCreates) Len() int           { return len(s.creates) }
+func (s sortableOccurrenceCreates) Less(i, j int) bool { return s.ids[i] < s.ids[j] }
+func (s sortableOccurrenceCreates) Swap(i, j int) {
+	s.ids[i], s.ids[j] = s.ids[j], s.ids[i]
+	s.creates[i], s.creates[j] = s.creates[j], s.creates[i]
+}
+
 func upsertBulkOccurrences(ctx context.Context, tx *ent.Tx, subjects model.PackageOrSourceInputs, artifacts []*model.IDorArtifactInput, occurrences []*model.IsOccurrenceInputSpec) (*[]string, error) {
 	ids := make([]string, 0)
 
@@ -196,6 +211,7 @@ func upsertBulkOccurrences(ctx context.Context, tx *ent.Tx, subjects model.Packa
 	index := 0
 	for _, occurs := range batches {
 		creates := make([]*ent.OccurrenceCreate, len(occurs))
+		createIDs := make([]string, len(occurs))
 		for i, occur := range occurs {
 			occur := occur
 			switch {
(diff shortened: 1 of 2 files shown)


gaganhr94 (commented):
Hi @pujitha24, thanks for the contribution. The changes look good. One nit, can we try to keep the comments minimal/concise. The test files don't require comments and in the occurrence.go, the comments if needed can be limited to one line.

pujitha24:
Good call, trimmed them down — the two comments in occurrence.go are now one line each, and I removed the comments from occurrence_sort_test.go entirely (kept the license header, obviously).

gaganhr94 (approved):

mihaimaruseac (approved):

